PR #3, opened 2024-04-30
PL1 PoC impl

asatklichov:
PL1 PoC implementation


KUGDev:
Hello @asatklichov 
Thank you for providing the PR
For the code changes, please, do the following before the PR approval:
1. Make DCO check to pass
2. Move the pl-one folder's content to the root of the repo
3. For the PL/I examples, we have the code_samples folder in the root of the repo, so please, move them from the examples folder to there
4. For the examples theirselves, please, provide, if possible, the following:
- PL/I language version of the source code applicable to
- Compiler used to build the source code / supported by compilers
- Operating system, where the code is possible to run + the version of the OS
- Brief inputs and outputs of the program
- Expected instructions to be highlighted / supported by LSP
Thank you!

KUGDev (changes requested):
Please, see the comment above

asatklichov:
> Please, see the comment above

Thanks, just returned today, will look at it.  This is just a simple PoC works only with hello-world. For full scale we will come with another shot, so better to keep this PoC as separate folder.

PR #8, opened 2024-09-04
Initial prototype implementation

msujew:
This is an initial implementation of the PL/I language server using Langium. It supports most (as far as we can tell) of the base language. Missing prominent features:

* Macro language (preprocessor)
* Compiler flags


packages/extension/src/extension/builtin-files.ts
```diff
@@ -0,0 +1,61 @@
+import * as vscode from 'vscode';
```
KUGDev, line 1: Missing license header

packages/language/src/documentation.ts/pli-documentation-provider.ts
```diff
@@ -0,0 +1,52 @@
+import { AstNode, JSDocDocumentationProvider } from 'langium';
```
KUGDev, line 1: Missing license header

packages/language/src/parser/pli-grammar.ts
```diff
@@ -0,0 +1,31 @@
+// import { AstUtils, GrammarAST, GrammarUtils, stream, Stream } from "langium";
```
KUGDev: Missing license header

scripts/merge-tmlanguage.mjs
```diff
@@ -0,0 +1,73 @@
+import deepmerge from 'deepmerge';
```
KUGDev, line 1: Missing license header

packages/language/src/parser/pli-lexer.ts
```diff
@@ -0,0 +1,58 @@
...
+                i++;
+            }
+            lines.push(text.substring(start, i + 1));
+        }
+        return lines;
+    }
+
+    private adjustLine(line: string): string {
+        let eol = '';
+        if (line.endsWith('\r\n')) {
```
asatklichov, line 39: Can't we just use Node.js `os.EOL` constant?

vitest.config.ts
```diff
@@ -0,0 +1,25 @@
...
+ * https://vitest.dev/config/
+ */
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+    test: {
+        deps: {
+            interopDefault: true
+        },
+        include: ['packages/**/test/**/*.test.ts']
```
asatklichov, line 23: Can we set some basic coverageThreshold in vitest config? It automatically fails if coverage drops down.  
like in Jest e.g.: 
thresholds: {
        lines: 80,
        functions: 80,
        branches: 80,
        statements: 80
      }

asatklichov (commented):
Code is nice 👍, now will check tests and do some testing

packages/language/src/parser/pli-lexer.ts
msujew: No, because `os.EOL` is OS dependant. Meanwhile, a Linux user could have a Windows encoded file on their system and the reverse could also be true. Using `os.EOL` would disregard that edge case and lead to incorrect offsets in the lexer.

KUGDev:
LGTM, works the same way in IntelliJ IDEA

KUGDev (approved):

vitest.config.ts
VitGottwald: I would strongly advise against that because it won't help with the quality and will create a number of undesired side effects.

asatklichov: This is common practice. Side effect only happens in code change not in test, this is just a test coverage measurement. Same measures tuned in sonar or Jenkins upon project.

VitGottwald:
Please, add `.nvmrc` file with expected node version (e.g. `v20.17.0`).

packages/language/src/pli.langium
```diff
@@ -0,0 +1,662 @@
...
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+grammar Pl1
+
+entry PliProgram:
+    (statements+=TopLevelStatement)*;
+
+TopLevelStatement: Package | ProcedureStatement | Directives;
```
asatklichov, line 17: let's discuss these grammar in separate meeting, i have many findings/inputs here

packages/extension/.vscodeignore
```diff
@@ -0,0 +1,3 @@
+.vscode/**
+.vscode-test/**
+src/**
```
VitGottwald, line 3: How about explicitly only including what is needed? `package.json` and `README.md` are included automatically.
```suggestion
# Ignore everything
*
**/*
../**/*     # when `vsce` run with `--no-dependencies` this is not needed 
../../**/*  # when `vsce` run with `--no-dependencies` this is not needed 

# Add what we actually want
!out/extension/main.cjs
!out/extension/main.cjs.map
!language-configuration.json
!syntaxes/pli.merged.json
```

.vscode/launch.json
```diff
@@ -0,0 +1,72 @@
...
+        },
+        {
+            "name": "Vitest: Run Selected File",
+            "type": "node",
+            "request": "launch",
+            "autoAttachChildProcesses": true,
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "runtimeExecutable": "npx",
```
VitGottwald, line 62: `pnpm exec` ?

VitGottwald:
Can we replace `pl-one` everywhere in the code and package.json(s) with `pli` to have single word language name like we do for the other extensions (hlasm, cobol, rexx, jcl, ...)?

code_samples/MACROS.pli
VAranouski, line 1: 5655-PL6  IBM(R) Enterprise PL/I for z/OS  6.1      (Built:20230320)
                     Options Specified
  Install:
  Command: SOURCE
  Line.File Process Statements
     1.0    *PROCESS X(S),A(S),LINECOUNT(60);
  Install:
Options can be specified this way inside the PL/I source.

I suggest to redo the code modification.

code_samples/PLI0002.pli
VAranouski, line 1: https://www.ibm.com/docs/en/db2-for-zos/12?topic=statements-equivalent-sql-pli-data-types
for instance, SQL TYPE IS CLOB_FILE is one of the Db2 data type for host vars. Those data types should be recognizable.

IBM3500I E      41.0    DSNH5034I DSNHSQLA LINE 41 COL 0  DECLARE VARIABLE NOT VALID FOR "1"
IBM1482I E      40.0    RULES(NOLAXDCL) violation: the variable BUF1_CLOB is declared without any data attributes.
So, this change has broken the source:
DCL BUF1_CLOB; // SQL TYPE IS CLOB_FILE;

I suggest to redo the code modification.

packages/extension/package.json
```diff
@@ -0,0 +1,47 @@
+{
+    "name": "pl-one-extension",
+    "description": "The extension specific package",
```
VitGottwald, line 3:
```suggestion
    "name": "pli-language-support",
    "description": "Autocomplete, highlighting and diagnostics for PL/I",
    "author": "Zowe",
    "license": "EPL-2.0",
    "preview": false,
    "publisher": "Zowe",
    "repository": {
        "type": "git",
        "url": "https://github.com/zowe/zowe-pli-language-support"
    },
```

asatklichov:
- adding CI build, vsix gen, and browser exec
- dummy validation and semantic class

code_samples/CHART.pli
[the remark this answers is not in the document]
msujew: I fixed this with my most recent commit. Thank you!

code_samples/MACROS.pli
msujew: We don't support preprocessor statements right now. This will be enabled once we do.

code_samples/PLI0002.pli
msujew: Supporting the SQL types is a separate feature. We will revert the change once we support them.

KUGDev:
From the repo perspective, it would be better to separate the actual solution between different folders. The implementation is intended to be supported both in VS Code and IntelliJ IDEA. Can I ask the team to bring all the VS Code related things into the "vscode-extension" folder? So that it would be a clear separation between the VS Code and IntelliJ IDEA parts in the future

azatsatklichov:
> From the repo perspective, it would be better to separate the actual solution between different folders. The implementation is intended to be supported both in VS Code and IntelliJ IDEA. Can I ask the team to bring all the VS Code related things into the "vscode-extension" folder? So that it would be a clear separation between the VS Code and IntelliJ IDEA parts in the future

It is a good idea, I agree.

msujew:
> Can I ask the team to bring all the VS Code related things into the "vscode-extension" folder? So that it would be a clear separation between the VS Code and IntelliJ IDEA parts in the future

@KUGDev I agree, that's why all the VS Code related parts already are part of the `packages/extension` directory. Are we missing something?

KUGDev:
@msujew ok I see. Then the IntelliJ IDEA related stuff will go into the "packages" folder as well. Will the "extension" folder remain with the same name? Cause if we add the "plugin" folder, it would mean that we are going to provide IntelliJ IDEA plugin only. I guess, the "intellij-plugin" naming would be more appropriate, and to stay consistent, it would mean that for the VSCode extension the folder name "vscode-extension" would be more appropriate as well

.vscode/launch.json
```diff
@@ -0,0 +1,88 @@
+// A launch configuration that launches the extension inside a new window
```
KUGDev, line 1: paths are changed, please reflect

.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
+name: Build
```
KUGDev, line 1: please, provide more specific name for the action, like "Build VSIX" or "Build VS Code Extension"

azatsatklichov (approved):
💯

azatsatklichov (approved):

azatsatklichov (approved):

.vscode/launch.json
msujew: Yes, thanks for catching that 👍

package.json
```diff
@@ -0,0 +1,31 @@
...
+        "eslint": "~8.57.0",
+        "deepmerge": "^1.5.0",
+        "langium": "~3.2.0",
+        "shx": "~0.3.4",
+        "typescript": "~5.4.5",
+        "vitest": "^1.6.0"
+    },
+    "volta": {
+        "node": "18.20.3",
+        "npm": "10.7.0"
```
VitGottwald, line 29: What about pnpm, can Volta manage it as well?

Also how about having an `.nvmrc` file? I think it is more common than Volta. In addition to `nvm` it is also supported by [`fnm`](https://github.com/Schniz/fnm) which I have been successfully using and very happy with for years.

msujew: I believe our yeoman generator automatically adds the `volta` entry. I can add a `.nvmrc` file instead 👍

asatklichov (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+    branches:
+      - main
+
+jobs:
+  build:
+    name: CI (${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, ubuntu-latest]
```
VitGottwald, line 17: macOS missing

.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Install pnpm
+      uses: pnpm/action-setup@v4
+      with:
+        version: 9
+    - name: Use Node.js
+      uses: actions/setup-node@v3
+      with:
+        node-version: '18'
```
VitGottwald, line 30: does not match `.nvmrc` value

msujew: True, it now uses the `.nvmrc` file to use the expected node version.

packages/language/src/pli.langium
```diff
@@ -0,0 +1,666 @@
...
+
+StringLiteral: value=STRING_TERM;
+NumberLiteral: value=NUMBER;
+
+FQN returns string: ID ('.' ID)*;
+FeatureID returns string: ID;
+
+hidden terminal WS: /\s+/;
+terminal ExecFragment: /(?<=EXEC\s*)[a-zA-Z]+\s[^;]*/i;
+terminal ID: /[$@#_a-zA-Z][\w_$@#]*/;
```
VitGottwald, line 652: Extralingual characters are allowed in identifiers via the usage of [NAMES](https://www.ibm.com/docs/en/SSY2V3_6.1/pdf/pg.pdf#page=113) compiler option. The default is `@#$` but other characters are allowed as well.

VitGottwald: Created an issue for this https://github.com/zowe/zowe-pli-language-support/issues/30

.vscode/launch.json
```diff
@@ -0,0 +1,88 @@
...
+            "type": "extensionHost",
+            "request": "launch",
+            "args": [
+                "--extensionDevelopmentPath=${workspaceFolder}/packages/vscode-extension",
+                "${workspaceFolder}/code_samples"
+            ],
+            "sourceMaps": true,
+            "outFiles": [
+                "${workspaceFolder}/packages/language/out/**/*.js",
+                "${workspaceFolder}/packages/vscode-extension/out/**/*.js"
```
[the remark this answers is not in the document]
msujew: It makes sense, but I'm nevertheless able to debug both the language server and the VS Code extension. Have you had trouble debugging the language server? Make sure to run the `Attach to Language Server` configuration after the language server has started up (i.e. after opening a `.pli` file).

packages/language/src/documentation.ts/pli-documentation-provider.ts
```diff
@@ -0,0 +1,69 @@
+/**
```
VitGottwald, line 1: Is the `.ts` in the folder name `documentation.ts` on purpose?

msujew: No, just an oversight, let me fix that :D

VitGottwald (approved):
LGTM. Great start for this project. Thank you @msujew !

PR #39, opened 2025-02-01
Add linking tests

### LINKED ISSUE #15: [validation] Setup baseline linking tests
Baseline linking tests need to be setup in place of the current stub. This issue doesn't require us to have a complete set of linking tests, that will come with time, just a nice bed to base our subsequent work on. Once we have a baseline that we're happy with, we can close this out, and open additional issues targeting specific linking cases as we identify them.

didrikmunther:
Closes #15


montymxb (commented):
Nice, thanks for putting this up! We can take this out of draft, update the tests with .fails (or .skips) and bring it into `development` over main.

PR #42, opened 2025-02-06
Adds Extracted PLI Compiler Messages/Codes

montymxb:
(Pulled from our fork)

Adds in what was extracted from the 6.1 compiler messages & codes document. Much in the same fashion as tests were extracted, there's over 1k codes, along with detailed messages & explanations. This pulls it all together into a TS file that we can reference with full information intact, ideally making it easier as we start to work these into validations.

Heads-up, the synthesized codes file is over 18k lines long. It's a bit large up front, but we're going to, eventually, need all of these for full support down the road.

One validation is updated to show how this works in practice, the rest can be trivially updated as well. I was also able to auto-detect when params were likely needing to be substituted, and modified those cases to be parametric pli code objects. In that way, usage requires you to provide the params to complete the message as intended.

Suggestions Markus made on the original issue:
- Add code description into the object itself for printing later on (deferred to later)
- "Accepting" the message code could be a one-liner using a helper method. That can be done in a follow-up PR


montymxb:
Pulling this into dev to support resolving #46 .

Issue #9, opened 2025-01-15
[feature] Compiler Options


msujew:
The PL/I language server should be able to parse compiler options from different sources (`*PROCESS` directive, `pgm_conf.json` file) and apply those compiler options during the runtime to correctly parse and validate the input files.

msujew:
FYI this currently depends on https://github.com/zowe/zowe-pli-language-support/issues/9, as some lexer features aren't available otherwise.

msujew:
I'll remove the lexer features from the branch for now and create a PR for that.

Issue #11, opened 2025-01-15
[feature] Preprocessor Support


msujew:
The PL/I language server should be able to correctly deal with preprocessor statements in the input file:

1. Preprocessor statements should be evaluated in the language server to provide information for the "normal" non-preprocessor code
2. The language server should help with writing preprocessor code by providing all the expected language server features (code completion, go to definition, go to references, etc.).

msujew:
Supported with #92.

PR #44, opened 2025-02-11
Support compiler options

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/9

Provides an initial implementation of the compiler options feature of PL/I. This includes:

1. Parsing the compiler options from the `*PROCESS` statement of files
3. Providing error markers for incorrectly specified compiler options

What is still left to do:

1. A lot of the compiler options haven't been completely implemented yet. These will need to be implemented as necessary, as some simply aren't relevant for the language server
2. Some values of `OR` and `NOT` do not work yet, as they require the preprocessor PR to land first (similar with the `NAMES` option)
3. The `pgm_conf.json` does not align to the schema of the other zowe language servers yet. We will address this in a later PR.


asatklichov:
Regarding last compiler options discussion here is the  [ findings ](https://github.com/zowe/zowe-pli-language-support/issues/45)

packages/language/src/pli.langium
```diff
@@ -647,6 +642,17 @@ NumberLiteral: value=NUMBER;
...
+terminal CONCAT_TOKEN: '||';
+terminal OR_EQUAL: '|=';
+terminal OR_TOKEN: '|';
+
+// TODO: Remove all ^ tokens once the lexer is working correctly
+terminal NOT_SMALLER: '¬<' | '^<';
+terminal NOT_EQUAL: '¬=' | '^=';
+terminal NOT_LARGER: '¬>' | '^>';
+terminal NOT_TOKEN: '¬' | '^';
+
```
asatklichov, line 655: Do we consider adding composite symbol? Like: 
` (:` Start of type function parameter list  
`:)` End of type function parameter list

msujew: I'm not quite sure I fully understand what you mean. Can you give an example?

asatklichov: In Language Ref. In Composite symbols Table

image

msujew: I see. I don't think we support those at all right now (none of the examples have used that construct so far. However, this is unrelated to the compiler options, as the reason why we extract the `NOT` and `OR` tokens into dedicated terminals is because they can be configured. Meanwhile, the `:)` and (:` token types cannot be configured via the compiler options, so we don't need special handling for them.

msujew:
@asatklichov I believe I've resolved most issues. Two are left:

1. Numbers vs IDs: This is a consequence of the lexing and will change with the preprocessor. I will keep it for now, and we will revisit it later
2. The `NOT` and `OR` change is deactivated for now. We will enable it once https://github.com/zowe/zowe-pli-language-support/issues/11 is integrated.

packages/language/test/compiler/options-parser.test.ts
```diff
@@ -0,0 +1,142 @@
+import { describe, test, expect } from "vitest";
```
KUGDev, line 1: Please add the copyright header

KUGDev (approved):

msujew:
Blocked by https://github.com/zowe/zowe-pli-language-support/pull/68 right now :)

PR #47, opened 2025-02-19
Better Diagnostics for Bad Label Ref

### LINKED ISSUE #46: Label Issue
As shown in below program, error is expected on  last line. 
Even wrong label is used, compiler does not show an error message. 

```
 MYPROC: PROCEDURE OPTIONS (MAIN);
    DCL TRUE BIT(1) INIT(1);
    DCL FALSE BIT(1) INIT(0);
    DCL OR_VALUE; 
    OR_VALUE = TRUE | FALSE;
    DCL NOT_VALUE;  

 END MYPOG;
```

montymxb:
Closes #46 by adding additional validations where a closing label fails to link correctly. Notably this retains the cross-reference aspect in the grammar, but supplements with 2 more diagnostics that match those from the compiler. There was one that wasn't picked up from the docs in a previous batch of added codes, so that was added in as well.

Beyond that I set up the validation tests file for checking these diagnostics as well, and will likely continue down that path with subsequent validation additions.

Also, this bumps the outer dev dependency from langium 3.2 to 3.3. This looks to have fallen out of version, and resulted in two different langium projects showing up in the node_modules. This caused some errors when resolving types in the parser tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "devDependencies": {
     "@types/node": "^18.19.70",
     "deepmerge": "^1.5.2",
-    "langium": "~3.2.1",
+    "langium": "~3.3.1",
     "prettier": "^3.4.2",
     "shx": "~0.3.4",
     "typescript": "~5.4.5",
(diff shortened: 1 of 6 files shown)


msujew (approved):

msujew:
Please rebase on the latest `development` before we can merge this.

PR #49, opened 2025-02-19
Run CI on Development branches

msujew:
Currently, only PRs against the `main` branch run their CI. This ensures that we also run the CI on PRs targeting the `development` branch.


montymxb (approved):
:octocat: approved, thanks for catching that!

PR #50, opened 2025-02-20
Reinstates Validations + Supports Calling Procs w/ Secondary Entry Point

### LINKED ISSUE #33: [improvement] Support Calling Procedures with Secondary Entry Point
Currently we support (and link) calls to internally declared procedures. However sometimes procedures can be external, and can declared via `dcl` with an ['entry' statement as a secondary entry point](https://www.ibm.com/docs/en/SSY2V3_6.1/pdf/lrm.pdf#page=148). These ones aren't currently handled, ex:
```pli
dcl a ext('a') entry(...);

call a( ... );
```

To resolve this we can loosen up our grammar definition for PL/I to handle calls to `NamedElement`s more generally (encompassing literal procedures & declarations). This then pushes the check up to the TypeSystem at the validation phase, which should be able to screen calls by whether the name refers to an actual procedure, or a secondary entry point of a procedure via a `dcl` with an `entry` attribute.

montymxb:
Resolves #33 by expanding the grammar to accept `NamedElements` when considering callable things. The checking for declarations is setup in a new validation rule + associated validation tests.

Additionally, reinstates validations for the new parser architecture to support the PR as well.

Also took a moment to refactor a bunch of incorrectly generated compiler message fields in the pli-codes file. Many of these were supposed to be parametric, but the extraction & conversion scripts I wrote up didn't take into account some funky spacing. Those cases have been corrected to accept args accordingly for better message output.

### DIFF
--- a/packages/language/src/validation/validator.ts
+++ b/packages/language/src/validation/validator.ts
@@ -14,32 +14,48 @@ import {
   IRecognitionException,
   MismatchedTokenException,
 } from "chevrotain";
-import { Diagnostic, DiagnosticInfo, Range, Severity } from "../language-server/types";
+import {
+  Diagnostic,
+  DiagnosticInfo,
+  Range,
+  Severity,
+} from "../language-server/types";
 import { ReferencesCache } from "../linking/resolver";
 import { isValidToken } from "../linking/tokens";
 import { PliProgram, SyntaxKind, SyntaxNode } from "../syntax-tree/ast";
 import { forEachNode } from "../syntax-tree/ast-iterator";
-import { PliValidationChecks, PliValidationFunction, registerValidationChecks } from "./pli-validator";
+import {
+  PliValidationChecks,
+  PliValidationFunction,
+  registerValidationChecks,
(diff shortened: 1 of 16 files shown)


packages/language/src/validation/validator.ts
```diff
@@ -14,23 +14,88 @@ import {
...
+  severity: Severity,
+  message: string,
+  info: DiagnosticInfo,
+) => void;
+
+/**
+ * Generates validation diagnostics (semantic checks) from the given AST node.
+ */
+export function generateValidationDiagnostics(root: PliProgram): Diagnostic[] {
+  // TODO @montymxb Mar. 27th, 2025: Checks are generated on each invocation, not ideal, needs a rework still
```
msujew, line 46: FYI: The `SourceFileDiagnostics` interface contains a `validation` list that should be used to store these validation diagnostics :)

msujew (approved):
Thanks, that's very nice! You can merge as it is or incorporate the comment below (if applicable).

KUGDev (approved):

PR #52, opened 2025-02-21
Additional Validation & Codes Support

montymxb:
Collection of a variety of validation changes + code improvements:

- adds missing pli codes & messages from before (previously generated off a sub-set of the document), improves existing ones to properly generate as their parametric variants, when that makes sense
- adds validations for empty programs
- fixes parsing of get statements, namely `GET LIST (N) FILE(SYSIN);`, which previously broke apart into several sub-statements
- validate various mutually exclusive returns attributes, like ALIGNED & UNALIGNED (previously attrs were being discarded)
- break up previous constant test and add a case for each discovered constant type
- added support for previously unhandled mixed-char constants
- ensures packages generate label prefixes so they can be generally referenced
- adds a pair of alias types to the built-ins (`__SIGNED_INT` & `__UNSIGNED_INT`). These are [implicitly added to any package](https://www.ibm.com/docs/en/SSY2V3_6.1/pdf/lrm.pdf#page=190), but will also be added to after the *PROCESS statements to any program w/out a package declaration as well, so I've added them into the built-ins for now
    - these are parametric based on the `LP(32)` and `LP(64)` compiler flags. Default appears to be `LP(32)` which is how they're coded for now


msujew (approved):
LGTM 👍

PR #55, opened 2025-02-25
Improve Ordinal Declaration Handling

montymxb:
Reworks ordinals a bit to allow:
- multiple redundant `define ordinal xyz (...) signed signed` or `define ordinal xyz (...) unsigned unsigned`
- mixed signed & unsigned attributes in any number, which pass in the 6.1 compiler w/out any issues
- adding to that, this does add a warning here to the case when mixed attributes are detected
- this also adds a warning when multiple precisions are detected, for as similar reason
- test cases to enforce expected behavior


KUGDev (approved):

packages/language/src/parser/parser.ts
```diff
@@ -2238,88 +2237,77 @@ export class PliParser extends AbstractParser {
...
+
+    // zero or more attributes following
+    this.MANY(() => {
+      // either signed/unsigned OR a precision attribute
       this.OR2([
         {
+          // signed | unsigned
           ALT: () => {
-            this.CONSUME_ASSIGN1(tokens.SIGNED, (token) => {
+            this.OR3([
```
msujew, line 2248: Suggestion: Please don't nest `OR` calls, unless absolutely necessary. This alternative needs to be evaluated multiple times in order to actually be parsed.

montymxb: Good point, I can flatten this out under a single alternative.

msujew (approved):
Looks good to me 👍

PR #59, opened 2025-02-27
Adds Development Previews

montymxb:
Modifies the existing playground workflow to also publish separately from the dev branch on push, while retaining the same behavior for main. Could also do `github.ref_name` for the branch checking as well, but opted for the full ref.

### DIFF
--- a/.github/workflows/playground.yml
+++ b/.github/workflows/playground.yml
@@ -9,6 +9,7 @@ on:
   push:
     branches:
       - main
+      - development
 
 jobs:
   docs:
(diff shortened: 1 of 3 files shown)


.github/workflows/playground.yml
```diff
@@ -34,10 +35,24 @@ jobs:
...
           force: false # Ensure that we don't override the pr previews
+      - name: Deploy Dev 🚀
+        if: github.ref == 'refs/heads/development'
+        uses: JamesIves/github-pages-deploy-action@v4.7.2
+        with:
+          folder: packages/playground/out
+          branch: previews
+          clean-exclude: | # Ensure that we don't delete the pr & dev previews
+            dev/
+            pr/
```
msujew, line 56: AFAIK this will result in the `main` deployment being deleted, as the `clean` deletes all files that aren't listed in `clean-exlude`. We should push the changes from `main` to its own `main` folder:
```suggestion
          clean-exclude: | # Ensure that we don't delete the pr & main previews
            main/
            pr/
```

.github/workflows/playground.yml
```diff
@@ -34,10 +35,24 @@ jobs:
           pnpm install --frozen-lockfile
           pnpm build
           pnpm playground
-      - name: Deploy 🚀
+      - name: Deploy Main 🚀
+        if: github.ref == 'refs/heads/main'
```
msujew, line 39: We need a `main` directory here, see above.

.github/workflows/playground.yml
montymxb: Yep good point, I'll put that in.

.github/workflows/playground.yml
montymxb: I'll also update the README to point to the new location then as well.

README.md
```diff
@@ -2,7 +2,7 @@
 
 PL/I Language Support by Zowe Community
 
-See it in an interactive [playground](https://zowe.github.io/zowe-pli-language-support/)!
+See it in an interactive [playground](https://zowe.github.io/zowe-pli-language-support/main/)!
```
msujew, line 5: Suggestion: Perhaps also add a link to the `development` playground here?

msujew (approved):
Yep, that should probably work 👍

PR #60, opened 2025-02-27
Link Fetch Entries

montymxb:
Small change to allow fetches to link to their entries + a single checking test.

As a small point, o validations are present at this point to assert whether the entry we're linking to is valid. Will put up a small issue for that.

### DIFF
--- a/packages/language/test/linking.test.ts
+++ b/packages/language/test/linking.test.ts
@@ -9,18 +9,8 @@
  *
  */
 
-import { describe, expect, test } from "vitest";
-import {
-  FetchStatement,
-  ProcedureStatement,
-  Statement,
-  SyntaxKind,
-} from "../src/syntax-tree/ast";
-import { collectDiagnostics } from "../src/workspace/compilation-unit";
-import {
-  expectGotoDefinition as expectGotoDefinitionRoot,
-  parse,
-} from "./utils";
+import { describe, test } from "vitest";
+import { expectGotoDefinition as expectGotoDefinitionRoot } from "./utils";
 
 /**
  * Scoping report: https://github.com/zowe/zowe-pli-language-support/issues/94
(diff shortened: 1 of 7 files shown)


KUGDev (approved):

packages/language/test/linking.test.ts
```diff
@@ -44,223 +50,253 @@ test("TODO", () => {});
...
+    const subStmt = procedureStatement.statements[1] as Statement;
+    expect(
+      subStmt &&
+        subStmt.value &&
+        subStmt.value.kind === SyntaxKind.FetchStatement,
+    ).toBeTruthy();
+
+    // look up the symbol
+    const fetchStmt = subStmt.value as FetchStatement;
+    const fetchStmtNameRef = fetchStmt.entries[0].name as string;
```
msujew, line 298: Suggestion: Doesn't it make more sense that the `name` property is a `ref: Reference<NamedElement>` or `call: MemberCall` instead? That way, we also get LSP support for that.

msujew (changes requested):

packages/language/test/linking.test.ts
montymxb: I hadn't set that up yet after the initial rebase to the new state, but I was just coming back to this one thinking something similar as well. I can just add that on top here

packages/language/test/linking.test.ts
```diff
@@ -318,20 +328,49 @@ describe("Linking tests", async () => {
...
+  test("fetch linking", async () => {
+    const doc = parse(
+      `
+        MAINPR: PROCEDURE OPTIONS(MAIN);
+        dcl A entry;
+        fetch A;
+        end MAINPR;
+    `,
+      { validate: true },
+    );
```
msujew, line 357: Question: Should we rewrite this into a `expectGotoDefinition` test? I'm ambivalent to it.

msujew (approved):

packages/language/test/linking.test.ts
montymxb: On second thought, just going to do this as part of the re-base here.

PR #63, opened 2025-03-03
Fix `GET` statement parser ambiguity

msujew:
Fixes a minor issue introduced in https://github.com/zowe/zowe-pli-language-support/pull/52. After adding the `LIST` option, we got into a parser conflict with the `LIST` option of the `DataSpecificationOptions` parser rule.

This fix resolves the issue by adhering to the additional information added to the `GET` statement in the documentation:

> **The keywords can appear in any order**. The data specification must appear unless the SKIP option is
specified.


montymxb (approved):
:octocat: Looks good, thanks! Plus relaxing to handle arbitrary keyword ordering is a nice benefit too.

I'll make a note to check later if we need to add a validation for screening out repeat declarations of the same option. I have a feeling it'll still pass in the compiler, but I'll keep it in mind.

PR #64, opened 2025-03-03
Fix remaining grammar ambiguities

msujew:
Our current grammar contains some ambiguities which prevents us from using a static (i.e. known $k$) lookahead. This PR fixes these issues in the least intrusive way possible. That way, I brought all decisions down to `k=2`. This behavior can be confirmed by running `pnpm test` and ensuring that no `AMBIGUOUS ALTERNATIVE` warning is printed to the console.

Note that I was nevertheless unable to enter the `maxLookahead` property in the Langium config due to our `ID` hack that we're using to parse keywords as IDs of procedures and variables.

### DIFF
--- a/packages/language/src/pli.langium
+++ b/packages/language/src/pli.langium
@@ -12,21 +12,15 @@
 grammar Pl1
 
 entry PliProgram:
-    (statements+=TopLevelStatement)*;
-
-TopLevelStatement: Package | PackageLevelStatements | Directives;
-
-Directives: SkipDirective | PopDirective | PushDirective | LineDirective | NoteDirective | PageDirective | PrintDirective | IncludeDirective | NoPrintDirective | ProcessDirective | ProcincDirective;
+    (statements+=Statement)*;
 
 // Defined on p. 90
 Package:
-    prefix=ConditionPrefix?
-    name=LabelPrefix
     'PACKAGE'
     exports=Exports? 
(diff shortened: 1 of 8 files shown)


packages/language/src/pli.langium
```diff
@@ -628,10 +652,15 @@ ExpExpression infers Expression: PrimaryExpression ({infer ExpExpression.left=cu
...
     Literal
-    | '(' Expression ')'
+    | ParenthesizedExpression
     | UnaryExpression
     | LocatorCall;
 
+// Performs special handling as a recursive descent parser otherwise cannot handle the 
+// special data-list type-3-DO loop
+// See also language reference pp. 301
+ParenthesizedExpression infers Expression: {infer Parenthesis} '(' value=Expression ('DO' do=DoType3<false>)? ')' ({infer Literal.multiplier=current} value=LiteralValue)?;
```
montymxb, line 662: Not sure about this part w/ the syntactic predicate. Parse results look okay, but I'm not sure how well we test this case, probably need to cover that in more detail coming up.

montymxb (approved):
:octocat: Looks good to add based on our tests so far. We'll still need to follow up necessary validations to guard against bad placement of package & procedure statements, and maybe some tests to ensure the new parenthesized expression case works as expected, but beyond that no objections here.

PR #65, opened 2025-03-04
Fix `development` playground link

msujew:
The files for the `development` branch are actually stored in the `dev` directory. Therefore, the link should point to that directory. See also [here](https://zowe.github.io/zowe-pli-language-support/dev/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@
 PL/I Language Support by Zowe Community
 
 See it in an interactive playground:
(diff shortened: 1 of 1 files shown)


montymxb (approved):
I was wondering if there would be one more thing, thanks for catching this.

I swear I also saw an issue with development not being able to write to to the preview branch, but I don't see it here, appears to be resolved.

PR #68, opened 2025-04-01
[Feature] Preprocessor

msujew:
Based on the work of @Lotes, I integrated everything into the new language server infrastructure.

Note that this is still missing a few features. However, we might want to add those in a separate PR.
1. Right now, the `CompilationUnit#uri` (i.e. the entry point into a program) is based purely on `%INCLUDE` relationships of open files. This should be able to include closed files as well.
2. LSP features are not implemented for preprocessor code
3. Linking inside of preprocessor code works, but is using a separate linking infrastructure from the rest of the language server. This should be consolidated.

Also, some preprocessor features are still missing. These will be implemented in a separate PR.

### DIFF
--- a/packages/language/src/language-server/connection-handler.ts
+++ b/packages/language/src/language-server/connection-handler.ts
@@ -9,21 +9,17 @@
  *
  */
 
-import { Connection, TextDocumentSyncKind } from "vscode-languageserver";
 import { CompletionUnitHandler } from "../workspace/compilation-unit";
-import { TextDocuments } from "./text-documents";
-import { definitionRequest } from "./definition-request";
 import {
   Connection,
   DocumentHighlight,
   TextDocumentSyncKind,
 } from "vscode-languageserver";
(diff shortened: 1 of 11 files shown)


packages/language/src/language-server/connection-handler.ts
```diff
@@ -45,46 +46,62 @@ export function startLanguageServer(connection: Connection): void {
...
+        const doc = TextDocuments.get(def.uri);
+        if (doc) {
+          const range = rangeToLSP(doc, def.range);
+          lspDefinitions.push({
+            uri: def.uri,
+            range,
+          });
+        }
+      }
+      return lspDefinitions;
```
montymxb, line 92: Just a note this duplicates the code above on 55 - 67. We should consider factoring this pattern out later.

montymxb (approved):
:octocat: initial check looks good. Let's pull this into dev!

PR #80, opened 2025-04-03
Reintroduce linking tests

didrikmunther:
The linking test system is using hand-picked utility functions from the Langium project to enable easy definition test.

Let me know if you prefer that some of these utility functions be moved to other files.

### DIFF
--- a/packages/language/test/linking.test.ts
+++ b/packages/language/test/linking.test.ts
@@ -9,258 +9,229 @@
  *
  */
 
-import { test } from "vitest";
+import { describe, test } from "vitest";
+import { expectGotoDefinition as expectGotoDefinitionRoot } from "./utils";
 
-test("TODO", () => {});
+/**
+ * Helper function to parse a string of PL/I statements,
+ * wrapping them in a procedure to ensure they are valid
+ */
+function expectGotoDefinition(
+  params: Parameters<typeof expectGotoDefinitionRoot>[0],
+) {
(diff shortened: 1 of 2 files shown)


packages/language/test/linking.test.ts
```diff
@@ -9,258 +9,229 @@
...
+          index: 0,
+          rangeIndex: 0,
+        });
+      });
+    });
+  });
+
+  describe("Unstructured tests", async () => {
+    // IMPORTANT: These tests are currently skipped. Unskip when scoping is resolved.
+    // https://github.com/zowe/zowe-pli-language-support/issues/29#issuecomment-2623842079
```
msujew, line 53: Please use `test.skip` instead of `describe` for skipped tests.

msujew (approved):
LGTM, just a comment, see below.

packages/language/test/linking.test.ts
didrikmunther: I'm using `test.skip` below when the testing is actually done: https://github.com/zowe/zowe-pli-language-support/blob/cde5b4f34348e6ac697181bc2bcc15afb58c3ed3/packages/language/test/linking.test.ts#L130C11-L130C20

Is this what you mean?

msujew: Ah, nevermind then. I thought that the `describe` in the line below was the test case.

PR #83, opened 2025-04-04
Add rename feature

### LINKED ISSUE #28: [feature-lsp] Support rename request

didrikmunther:
Fixes: #28 

Let me know if the utility functions should be structured/put somewhere else!


packages/language/src/language-server/rename-request.ts
```diff
@@ -0,0 +1,29 @@
...
+import { referencesRequest } from "./references-request";
+import { groupBy } from "../utils/group-by";
+import { DocumentUri } from "vscode-languageserver-types";
+import { Location } from "./types";
+
+export function renameRequest(
+  sourceFile: SourceFile,
+  offset: number,
+): Record<DocumentUri, Location[]> {
+  const references = referencesRequest(sourceFile, offset);
```
didrikmunther, line 22: I'm a bit unsure if I should be using the actual `referencesRequest` to get references. Perhaps another abstraction layer for getting references should exist? The `referencesRequest` currently contains quite a bit of specialized code for reference/name handling.

packages/language/src/utils/map-values.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Map the values of an object using a function.
+ */
+export function mapValues<K extends string | number | symbol, T, U>(
```
didrikmunther, line 4: Let me know if these utility functions are too unintelligible. I normally prefer to abstract small functional utility functions that make the main business logic readable, but I'd like to learn how you prefer to structure this!

packages/language/src/language-server/rename-request.ts
msujew: We should add an abstraction to it - we do something similar in Langium as well. Ideally, it should return all tokens (including the declaring token) that are somehow related to the targeted element.

packages/language/src/utils/map-values.ts
msujew: IMO, it might make sense to group these functions in a single file.

packages/language/src/language-server/rename-request.ts
didrikmunther: I added an abstraction to reference lookups!

msujew (approved):
LGTM, thank you 👍

PR #84, opened 2025-04-04
Add document highlight

didrikmunther:
This feature currently reuses the existing `referencesRequest`, since it contains a lot of specialized name/reference handling logic. Let me know if this is wrong.

### DIFF
--- a/packages/language/src/language-server/connection-handler.ts
+++ b/packages/language/src/language-server/connection-handler.ts
@@ -36,6 +36,7 @@ export function startLanguageServer(connection: Connection): void {
         },
         definitionProvider: true,
         referencesProvider: true,
+        documentHighlightProvider: true,
         semanticTokensProvider: {
           legend: semanticTokenLegend,
           full: true,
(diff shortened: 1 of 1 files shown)


packages/language/src/language-server/connection-handler.ts
```diff
@@ -91,5 +92,22 @@ export function startLanguageServer(connection: Connection): void {
...
+    const sourceFile = sourceFileHandler.getSourceFile(URI.parse(uri));
+    if (textDocument && sourceFile) {
+      const offset = textDocument.offsetAt(position);
+      const definition = referencesRequest(sourceFile, offset);
+      return definition.map((def) => {
+        return {
+          uri: def.uri,
+          range: rangeToLSP(textDocument, def.range),
+        };
+      });
```
msujew, line 108: Suggestion: `uri` is not part of the expected return type `DocumentHighlight`. Also please make sure to filter based on the current URI. Otherwise, you will get document highlights spilling into the current file from other files.

msujew: In fact, I would recommend to use `DocumentHighlight.create`, to ensure that everything is typed correctly (we should probably also do so for the other LSP services as well!)

msujew (approved):

Issue #76, opened 2025-04-02
[feature-lsp] Support LSP features for preprocessor code


msujew:
Depends on https://github.com/zowe/zowe-pli-language-support/issues/71 and https://github.com/zowe/zowe-pli-language-support/issues/70. Afterwards, we should be able to easily support the preprocessor with LSP features.

Issue #71, opened 2025-04-02
[Feature] Adjust preprocessor AST to use a modified "normal" AST


msujew:
Currently, the preprocessor uses a completely separate AST structure compared to the one we use for the rest of the document lifecycle. We should attempt to unify those AST structures so that we can reuse them for the LSP services.

msujew:
Closed https://github.com/zowe/zowe-pli-language-support/pull/92.

Issue #62, opened 2025-03-03
[discussion] AST/CST structure for code with preprocessor statements


msujew:
# PL/I AST Structure

For programs with preprocessor statements, we cannot rely on the AST/CST of Langium, as we lose all preprocessor code before getting to the Langium parser. I.e. the following code:

```
%dcl A char;
%A = 'B';
dcl A%C fixed bin(31);
```

Simply becomes in Langium:

```
dcl BC fixed bin(31);
```

Not only does the reference to the preprocessor variable `A` get lost, but all preprocessor code is no longer discoverable by Langium. Meaning that cross references to the variable `A` cannot be resolved, as Langium has no idea that the variable existed in the first place.

# Implementation Approach

In the example above, being able to jump to the A preprocessor variable is very much required. Similarly, code completion should be able to pick up on the structure of existing preprocessor code, meaning that code completion should also be available within preprocessor code.

This requires storing the preprocessor information in the AST/CST. However, we must not ignore the end result of the preprocessor, as the `A%C` variable becomes a variable named `BC` and is referencable as such. As a consequence, we must be able to store both the `A%` and `BC` data in our AST/CST somehow.

## AST

The AST can be written in a way that is very similar to the existing Langium implementation, with the addition to also create AST nodes for preprocessor statements. I.e. the following code:

```
%dcl A char;
%A = 'B';
```

And the corresponding “normal” code:

```
dcl A char;
A = 'B';
```

Can be represented using the same AST node (and can be parsed using the same parser rules), with the additional information that the first code snippet has to be marked as `preprocessor` to ensure that we can correctly differentiate between a variable `A` in code and a preprocessor variable `A` used in preprocessor code.

Some special preprocessor statements could also find representation in the AST that way, such as `%SKIP` as these statements can for example be used to provide syntax coloring (i.e. reducing opacity for skipped lines).

### Regarding preprocessor statements in non-preprocessor code

Let’s take a look at the example above again:

```
%dcl A char;
%A = 'B';
dcl A%C fixed bin(31);
```

Taking the approach as outlined above will result in this structure:

```js
[
    { preprocessor: true, type: "declaration", name: "A", type: "char" },
    { preprocessor: true, type: "assignment", left: "A", right: "'B'" },
    { preprocessor: false, type: "declaration", name: "BC", type: "fixed bin(31)" }
]
```

[...]

Issue #70, opened 2025-04-02
[Feature] Use symbol table in preprocessor parser


msujew:
Currently, the preprocessor uses an adhoc referencing feature. However, this does not play well with the LSP infrastructure. The natural consequence: We need to at least store references and names in the symbol table.

msujew:
With the rewrite of the preprocessor interpreter, this is no longer relevant.

PR #92, opened 2025-04-10
Rewrite preprocessor parser to support LSP features

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/71
Closes https://github.com/zowe/zowe-pli-language-support/issues/76
Closes https://github.com/zowe/zowe-pli-language-support/issues/62

Related to https://github.com/zowe/zowe-pli-language-support/issues/70 (but doesn't resolve it).

This PR rewrites the preprocessor parser to use the same AST types already used by the normal PL/I parser. This allows us to handle all LSP features without adding any special handling for preprocessor nodes.


packages/language/src/preprocessor/pli-preprocessor-parser.ts
```diff
@@ -399,416 +575,583 @@ export class PliPreprocessorParser {
...
-      return {
-        type: "string",
-        value: tokens,
-      };
+      return this.stringLiteral(state);
     } else if (state.canConsumeKeyword(PreprocessorTokens.Id)) {
       if (state.isOnlyInStatement()) {
-        state.consume(PreprocessorTokens.Percentage);
+        state.consume(
+          undefined!,
```
montymxb, line 1136: Noticed this showing up quite a bit, Maybe we should relax `consume` to take `SyntaxNode | undefined` up front.

packages/language/src/preprocessor/pli-preprocessor-program-builder.ts
```diff
@@ -61,14 +61,32 @@ export class PliPreprocessorProgramBuilder {
...
+      return this.createLabel();
+    }
+  }
+
+  resolveLabel(label: ast.LabelReference): Label | undefined {
+    if (label.label?.text) {
+      const resolvedLabel = this.labels[label.label.text];
+      if (resolvedLabel) {
+        return resolvedLabel;
+      }
```
montymxb, line 87: We can return the label as is, undefined or otherwise, no need to check.

packages/language/src/validation/messages/IBM1295IE-sole-bound-specified.ts
```diff
@@ -18,17 +18,17 @@ export function IBM1295IE_sole_bound_specified(
   bound: DimensionBound,
   accept: PliValidationAcceptor,
 ): void {
-  if (bound.bound2 !== undefined) {
+  if (bound.lower !== undefined) {
```
montymxb, line 21: Just to be sure, `bound2` is supposed to be `lower`?

packages/language/test/parser/pli-preprocessor-interpreter-state.test.ts
```diff
@@ -94,8 +94,8 @@ describe("Preprocessor Interpreter State", () => {
       const state = Fixtures.emptyProgram();
       state.declare("VARIABLE", {
         active: true,
-        dataType: "character",
-        scanMode: "noscan",
+        dataType: "CHARACTER",
+        scanMode: "SCAN",
```
montymxb, line 98: Was this intentionally changed from `NOSCAN`?

packages/language/test/parser/pli-preprocessor-interpreter-state.test.ts
```diff
@@ -118,8 +118,8 @@ describe("Preprocessor Interpreter State", () => {
...
```
montymxb, line 122: Same here

packages/language/test/parser/pli-preprocessor-interpreter-state.test.ts
```diff
@@ -137,8 +137,8 @@ describe("Preprocessor Interpreter State", () => {
       //act
       state.declare("VARIABLE", {
         active: true,
-        dataType: "fixed",
-        scanMode: "noscan",
+        dataType: "FIXED",
+        scanMode: "SCAN",
```
montymxb, line 141: Ditto

packages/language/test/parser/pli-preprocessor-interpreter-state.test.ts
```diff
@@ -151,16 +151,16 @@ describe("Preprocessor Interpreter State", () => {
       const state = Fixtures.emptyProgram();
       state.declare("VARIABLE", {
         active: true,
-        dataType: "fixed",
-        scanMode: "noscan",
+        dataType: "FIXED",
+        scanMode: "SCAN",
```
montymxb, line 155: Ditto

packages/language/test/parser/pli-preprocessor-interpreter-state.test.ts
```diff
@@ -151,16 +151,16 @@ describe("Preprocessor Interpreter State", () => {
...
         value: [],
       });
 
       //act
       state.declare("VARIABLE", {
         active: true,
-        dataType: "character",
-        scanMode: "noscan",
+        dataType: "CHARACTER",
+        scanMode: "SCAN",
```
montymxb, line 163: Ditto

montymxb (commented):
Gave this all a read over:

- Can we refactor the ScopeCache so it's just responsible for a single cache, and add something on top that gives/manages access to them in some way? That should be better to work with later on, especially when we need to update the core scoping logic.
- May be good to write up separate issues for some of these extra TODOs, particularly the validation related ones
- Noted quite a few consolidations that could be made to shore things up, but these are non-blocking overall
- Various old blocks of code that are commented out, but exist in a new form elsewhere, also non-blocking at this time
- Suggestion for changing what `consume` accepts

packages/language/src/validation/messages/IBM1295IE-sole-bound-specified.ts
msujew: Not really, the old logic was really complicated. Essentially, the parser parses the bounds as `${bound1}:${bound2}`, with the `:${bound2}` being optional. If `bound2` is specified, it becomes the upper bound. If it's not specified `bound1` is the upper bound. In Langium, we were unable to represent this logic in the AST (which is why we used `bound1`, `bound2` as names in the first place).

For this case: `lower` is only set, in case we actually parse the `${bound1}:${bound2}` construct. In this case, we no longer have just a sole bound. Therefore, the validation does not apply anymore and we return.

montymxb: Makes sense, thanks for the clarification

montymxb (approved):
:octocat: Nice this looks good to go

PR #93, opened 2025-04-10
Run PR previews on target branch

msujew:
PR previews from outside contributors (from forks) currently don't work because the resources required to push to GitHub Pages aren't accessible to them. This PR fixes that by running the PR previews on the target branch (instead of on the source branch).


montymxb (approved):
:octocat: Let's pull this in, main & dev are protected now

Issue #81, opened 2025-04-03
[feature-lsp] Differentiate between procedure and variable tokens


msujew:
Right now, labels for procedures and variable names are handled completely identical in the semantic token handler. However, procedures should receive `function`-like highlighting.

montymxb:
Self-assigned on behalf of Steven, until he's added in.

ssmifi:
Thanks. (Commented to be assignable.)

PR #95, opened 2025-04-10
Add semantic highlighting for procedures

ssmifi:
Fixes: https://github.com/zowe/zowe-pli-language-support/issues/81

Add semantic highlighting for procedures
Add semantic tokens test
Add semantic token decoder to improve token processing, esp. in tests


msujew (approved):
Thank you, that looks good to me. I really like the semantic token tests!

Issue #73, opened 2025-04-02
[feature-lsp] Support navigation of closed files


msujew:
Currently, all navigation features only support navigating to files that are open in the current editor. We should also be able to navigate that aren't open yet.

ssmifi:
I can look into this.

PR #97, opened 2025-04-11
Inconsistency in compilation units

ssmifi:
The marked code location introduces an inconsistency in the handling of compilation units.

- If the _to be included_ file is opened first, it will be added with its unit.
- If the _include_ file is opened first, both will be added with the unit of the _include_ file and used when the _to be included_ files is opened afterwards.
- If the _to be included_ file is then closed again, the unit will be deleted entirely.

This makes it difficult for the LS to react correctly. Let us discuss how this should be handled in the future.

I suggest to remove the included files from the `CompletionUnitHandler` for now to remove the inconsistency and to optimize later if necessary.

Blocks https://github.com/zowe/zowe-pli-language-support/issues/73

### DIFF
--- a/packages/language/src/workspace/compilation-unit.ts
+++ b/packages/language/src/workspace/compilation-unit.ts
@@ -129,9 +129,7 @@ export class CompletionUnitHandler {
         URI.parse(event.document.uri),
       );
       lifecycle(unit, event.document.getText());
-      unit.files.forEach((file) => {
-        this.compilationUnits.set(file.toString(), unit);
-      });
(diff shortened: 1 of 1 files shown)


msujew (changes requested):
Not a big fan of this. I wanted to fix this issue on monday.

msujew:
Let's merge it for now though.

ssmifi:
I can postpone https://github.com/zowe/zowe-pli-language-support/issues/73 until Monday, if you prefer. No worries. ;-)

ssmifi:
Closed due to #99.

Issue #86, opened 2025-04-07
[Feature] Support NOT and OR compiler options


msujew:
Based on https://github.com/zowe/zowe-pli-language-support/pull/44.

montymxb:
Closed by #109

PR #109, opened 2025-04-14
Support OR and NOT compiler options

### LINKED ISSUE #88: AST is undefined during faulty program
This program:

```pl1
 STARTPR: PROCEDURE OPTIONS (MAIN);
 OUTER: PROCEDURE;
   DCL ABC;

   OUTER: PROCEDURE;
      DCL ABC;

      CALL ABC;
   /* END OUTER; */

   CALL ABC;
 END OUTER;
 END STARTPR;
```

Will cause the `compilationUnit.ast` to be undefined, crashing the language server. The `CompilationUnit.ast` type does not reflect this.

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/86
Closes #88

### DIFF
--- a/packages/language/test/compiler/lexing.test.ts
+++ b/packages/language/test/compiler/lexing.test.ts
@@ -1,0 +1,53 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { describe, test } from "vitest";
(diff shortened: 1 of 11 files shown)


packages/language/test/compiler/lexing.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  test("Should lex correct OR without compiler option", () => {
+    const doc = parseStmts(`
+  DECLARE VAR FIXED;
+  VAR = 1 | 2;
+    `);
+    assertNoParseErrors(doc);
+  });
+
+  test("Should lex correct NOT compiler option", () => {
```
montymxb, line 44: Should be an OR case, I'll tack that on in a follow-up commit here.

montymxb (commented):
Test case needs to be modified to show OR. I'll add that on top, and we can take this in.

montymxb (approved):
:octocat: Ready to add in

Issue #94, opened 2025-04-10
[Feature] Implement fully qualified name scoping


msujew:
Related to https://github.com/zowe/zowe-pli-language-support/pull/87.

didrikmunther:
# Qualified name scoping report

By running some code snippets on the mainframe and cross-referencing with the documentation, we try to reverse engineer the scoping/qualified name behavior.

## Background

### <a name="ambiguity"></a>Ambiguity

A reference to an element of structure is viewed as unambiguous if any of the following conditions
applies:
1. The reference is fully qualified, that is, the reference to the element includes the names of all of its
parents.
2. The reference is not a partially or fully qualified reference to any other variable in the block where the
element is declared.

## Results

### 1. Table/Structure Variable Attribute Error

<table>
<tr><td> Code </td> <td> Result </td> <td> Explanation </td></tr>
<tr><td>

```pli
 START: PROCEDURE OPTIONS (MAIN);
  DCL 1 A CHAR(8) INIT("A"),
      2 B CHAR(8) INIT("B");
 END START;
```

</td><td>

`IBM2462I E: The attribute CHARACTER conflicts with the attribute STRUCTURE and is ignored.`
`IBM2462I E: The attribute INITIAL conflicts with the attribute STRUCTURE and is ignored.`

</td>
<td>A structure definition (`A` in this case) cannot itself have a value.</td>
</tr>
</table>

### 2. Implicit qualification of nested structure variables

<table>
<tr><td> Code </td> <td> Result </td> <td> Explanation </td></tr>
<tr><td>

```pli
 START: PROCEDURE OPTIONS (MAIN);
  DCL 1 A,
      2 B CHAR(8) VALUE("B");

  PUT(B);
  PUT(A.B);
 END START;
```

</td><td>

`B B`

</td>
<td>Table variables can be implicitly qualified</td>
</tr>
</table>

### 3. Shadowing of implicitly qualified nested structure variables with root variable

<table>
<tr><td> Code </td> <td> Result </td> <td> Explanation </td></tr>
<tr><td>

```pli
 START: PROCEDURE OPTIONS (MAIN);
  DCL 1 A,
      2 B CHAR(8) VALUE("B");

  DCL B CHAR(8) VALUE("B2");

  PUT (B);
  PUT (A.B);
 END START;
```

</td><td>

`B2 B`

</td>
[...]

PR #112, opened 2025-04-16
Add qualified name support

didrikmunther:
fixes: #94 

## Intro

PL/I structured variables do not need to be fully qualified. The nested declaration `K` in the following structured declaration:

```pli
 DCL 1 A,
       2 B,
         3 K;
```

can be (partially) qualified by: `K`, `B.K`, `A.K`, and fully qualified by `A.B.K`.

There are many more quirks, especially when it comes to solving ambiguity, which are explained more in #94.

## Implementation

Take this PL/I code as an example:

```pli
 DCL 1 A,
       2 B,
         3 K,
       2 C,
         3 K;
```

We turn this into multiple linked list data structures using recursive descent parsing (`QualifiedSyntaxNode`):

```
K -> B -> A
B -> A
K -> C -> A
C -> A
```

When encountering a qualified reference name (`MemberCall`), we turn it into a reversed list of qualifiers, e.g. `C.K` becomes `["K", "C"]`.

We search for the first name `K`, and get two possibilities:

```
K -> B -> A
K -> C -> A
```

We step through these linked lists, eliminating elements in the qualified names list when matching a name:

```
K -> B -> A ["K", "C"]
K -> C -> A ["K", "C"]
```

```
B -> A ["C"]
C -> A ["C"]
```

```
A ["C"]
A []
```

```
- ["C"]
A []
```

The first linked list does not qualify correctly, so we ignore that. The second linked list partially qualifies.

## TODO:

1. Handle star names
2. Handle erroneous scoping situations and provide diagnostics for this similar to the compiler errors shown in: #94 
3. Write tests for these erroneous situations
4. Implicit declarations

Examples or erroneous situations: redeclarations, ambiguity, wrong level numbering.

Redeclarations are by always referring to the first declaration.

## Performance

Performance test done on Apple M4 Pro 48 GB.

### `generateSymbolTable`

|Program|LoC|First Runtime Duration|Converging Towards|
|-|-|-|-|
|`ADVNTOPT.pli`|3202|17.23 ms|3.5 ms|
|`MACROS.pli`|2039|30.6 ms|1.9 ms|
|`CHART.pli`|876|4.7 ms|1.3 ms|

### `generateSymbolTable` + `link`

|Program|LoC|First Runtime Duration|Converging Towards|
|-|-|-|-|
|`ADVNTOPT.pli`|3202|28.17 ms|3.8 ms|
|`MACROS.pli`|2039|35.1 ms|2.7 ms|
|`CHART.pli`|876|7.1 ms|1.8 ms|

### `generateSymbolTable` + `link` (after performance fixes [8aa204e](https://github.com/zowe/zowe-pli-language-support/pull/112/commits/8aa204eb57ea6bb76d137496567b4143e89581d3))

|Program|LoC|First Runtime Duration|Converging Towards|
|-|-|-|-|
|`ADVNTOPT.pli`|3202|25.2 ms|3.6 ms|
|`MACROS.pli`|2039|35.3 ms|2.7 ms|
|`CHART.pli`|876|6.3 ms|2.0 ms|

### DIFF
--- a/packages/language/src/linking/resolver.ts
+++ b/packages/language/src/linking/resolver.ts
@@ -12,12 +12,7 @@
 import { IToken } from "chevrotain";
 import { Location, tokenToRange } from "../language-server/types";
 import { TokenPayload } from "../parser/abstract-parser";
-import {
-  MemberCall,
-  Reference,
-  ReferenceItem,
-  SyntaxNode,
-} from "../syntax-tree/ast";
+import { Reference, SyntaxKind, SyntaxNode } from "../syntax-tree/ast";
 import { binaryTokenSearch } from "../utils/search";
 import {
   getNameToken,
@@ -29,18 +24,12 @@ import { URI } from "../utils/uri";
 import { CompilationUnit } from "../workspace/compilation-unit";
 
 export class ReferencesCache {
-  private memberCalls: MemberCall[] = [];
   private list: Reference[] = [];
   private reverseMap = new Map<SyntaxNode, Reference[]>();
 
   clear(): void {
-    this.memberCalls = [];
     this.list = [];
     this.reverseMap.clear();
-  }
-
-  addMemberCall(memberCall: MemberCall): void {
-    this.memberCalls.push(memberCall);
   }
 
   add(reference: Reference): void {
@@ -65,10 +54,19 @@ export class ReferencesCache {
   allReferences(): Reference[] {
     return this.list;
   }
+}
 
-  allMemberCalls(): MemberCall[] {
-    return this.memberCalls;
+// Returns the qualified name in reverse order, e.g. "A.B.C" -> ["C", "B", "A"]
+export function getQualifiedName(reference: Reference): string[] {
+  if (reference.owner.container?.kind === SyntaxKind.MemberCall) {
+    const memberCall = reference.owner.container;
+    if (memberCall.previous?.element?.ref) {
+      const names = getQualifiedName(memberCall.previous.element.ref);
+      return [reference.text, ...names];
+    }
   }
+
+  return [reference.text];
 }
 
 export function resolveReference<T extends SyntaxNode>(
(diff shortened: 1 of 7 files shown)


packages/language/src/linking/symbol-table.ts
```diff
@@ -9,29 +9,201 @@
...
-    return this.symbols.get(name);
+  private peek(): DeclaredItem | undefined {
+    return this.items[0];
+  }
+
+  private pop(): DeclaredItem | undefined {
+    return this.items.shift();
+  }
+
+  _generate(
```
didrikmunther, line 117: Should be private

packages/language/src/linking/resolver.ts
```diff
@@ -56,6 +56,19 @@ export class ReferencesCache {
   }
 }
 
+// Returns the qualified name in reverse order, e.g. "A.B.C" -> ["C", "B", "A"]
+export function getQualifiedName(reference: Reference): string[] {
+  if (reference.owner.container?.kind === SyntaxKind.MemberCall) {
+    const memberCall = reference.owner.container;
+    if (memberCall.previous?.element?.ref) {
+      const names = getQualifiedName(memberCall.previous.element.ref);
+      return [reference.text, ...names];
```
msujew, line 65: Suggestion: Since we don't reuse the array anywhere else, can we do this instead? This performs an in-place modification, which should be faster.
```suggestion
      names.unshift(reference.text);
      return names;
```

packages/language/src/linking/resolver.ts
```diff
@@ -64,21 +77,30 @@ export function resolveReference<T extends SyntaxNode>(
...
-      return undefined;
-    }
+  const qualifiedName = getQualifiedName(reference);
+
+  const scope = unit.scopeCache.get(reference.owner);
+  if (!scope) {
+    return undefined;
+  }
+
+  const [symbol, ...symbols] = scope.getSymbols(qualifiedName);
```
msujew, line 91: Suggestion: I'm not 100% on how this is performing internally, but this probably constructs a new array for `symbols`. We should adjust the logic to just use the returned array and look for `symbols[0]` and whether `symbols.length > 1`.

packages/language/src/linking/symbol-table.ts
```diff
@@ -172,25 +325,17 @@ function iterate(
         ...node,
         end: null,
       };
-      forEachNode(procedure, iterateChild);
+      forEachNode(procedure, iterateChild(scope));
       if (node.end) {
-        iterate(node.end, scopeCache, parentScope, references);
+        iterateChild(parentScope)(node.end);
         node.end.container = node;
```
msujew, line 331: Suggestion: I would assume this `container` assignment is no longer necessary?

packages/language/src/linking/symbol-table.ts
```diff
@@ -96,68 +268,49 @@ export function iterateSymbols(compilationUnit: CompilationUnit): void {
...
-  // to build their symbol table.
-  // CAUTION: This function has two side effects:
-  // * it sets the container property on the child node.
-  // * it may overwrite the scope variable if a child node returns a new scope.
-  // This is done to enable redeclarations of local variables.
-  const iterateChild = (child: SyntaxNode) => {
-    // Set the parent node on our first iteration through the file
-    child.container = node;
+  // Are we entering a containing node? Then we should create a new scope for all our children.
+  const scope = shouldNodeCreateNewScope(node)
```
msujew, line 307: Question: In a lot of programming languages every "block" generates a new scope. Does this also include `DO`, `IF` and `BEGIN` statements in PLI? Or do those statements simply inherit the scope of their procedure?

msujew (commented):
Looking really good. A few comments and then it should be good to go 👍

packages/language/src/linking/symbol-table.ts
didrikmunther: I haven't tested the scoping mechanisms of `DO`, `IF`, and `BEGIN` yet. I was thinking that this PR lays a foundation for a generic scoping system (with an implementation for `PROCEDURE` scoping), and a future PR actually implements the full scoping mechanisms of the language.

What do you think?

packages/language/src/linking/resolver.ts
didrikmunther: That makes sense! Fixed!

packages/language/src/linking/resolver.ts
didrikmunther: That makes sense! I made a change which removed all destructuring operations.

packages/language/src/linking/symbol-table.ts
msujew: Sure, fine with me 👍

packages/language/src/linking/symbol-table.ts
msujew: The `container` property is purely for traversing the AST and is by itself independent of the scoping system. However, as it is, the `node.end.container` value is already `node`. Therefore the assignment here is no longer required.

didrikmunther: I see what you mean now! It's fixed!

msujew (approved):
Looks good, thank you 👍

Issue #96, opened 2025-04-11
[Test] Validate AST container structure


ssmifi:
There should be a dedicated test that validates the container structure of the AST. At the moment, some errors may appear in the lsp tests, but it is not immediately obvious that they are related to the AST.

PR #113, opened 2025-04-17
Add symbol table test and fixes

ssmifi:
Please review the new symbol table test and applied fixes. 

Closes https://github.com/zowe/zowe-pli-language-support/issues/96


packages/language/test/linking/symbol-table.test.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+function attachDebugInfo<T extends { kind: SyntaxKind }>(node: T): T {
+  Object.defineProperty(node, "kindName", {
+    get: () => SyntaxKind[node.kind],
+    enumerable: true,
+    configurable: true,
+  });
+  return node;
+}
```
msujew, line 34: Question: Do we actually need this?

packages/language/test/linking/symbol-table.test.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+});
+
+describe("Symbol Table PLI-DOC programs", () => {
+  for (const [index, program] of pliDocPrograms.entries()) {
+    test(`should create valid symbol table for program ${index}`, () => {
+      expectSymbolTable(program);
+    });
+  }
+});
```
msujew, line 99: Suggestion: Instead of running these tests on dedicated new `pliPrograms` and `pliDocPrograms`, could we embed the `expectSymbolTable` and `expectASTNodeContainer` into our existing parsing tests? That way, we don't need to maintain multiple lists of programs for testing. Specifically, we could add them to `assertNoParseErrors`.

packages/language/test/linking/symbol-table.test.ts
ssmifi: I added a property to identify which elements were skipped by the AST iterator, and this one actually turned out to be quite helpful during debugging. ;-)
I can add a comment to clarify its purpose.

packages/language/test/linking/symbol-table.test.ts
ssmifi: The idea was to have a central store for the programs, allowing easy extension by simply adding new ones to the list and eventually running the full lifecycle on them (later on).
That said, I can integrate the new tests into `assertNoParseErrors` if you prefer—though we might want to consider renaming the function in that case?

packages/language/test/linking/symbol-table.test.ts
msujew: I see. I'm a bit unhappy that this requires a copy of the lifecycle function but with attached debug information - whenever we change the lifecycle function, we would also need to adjust the "debug" lifecycle. Do you think there's a way to prevent code duplication?

Note that you can find all elements there were skipped by the AST iterator by looking for the `container: null` element on AST nodes.

packages/language/test/linking/symbol-table.test.ts
ssmifi: I'm still undecided about consolidating the tests. I prefer not to refactor the code while so much is still in flux.
To avoid blocking the parser fixes any further, I went with your initial suggestion and integrated the test into the existing parser test for now.
I've also opened a ticket https://github.com/zowe/zowe-pli-language-support/issues/121 to remind us about consolidating the test cases later, but it's not a high priority at the moment.

montymxb: I would be onboard w/ de-duping programs for the various testing levels, but I would also be okay to defer this to a follow-up PR.

packages/language/test/lsp/semantic-tokens.test.ts
```diff
@@ -20,20 +20,19 @@ import { URI } from "../../src/utils/uri";
 import { SemanticTokenDecoder } from "../../src/language-server/semantic-token-decoder";
 import { SemanticTokenTypes } from "vscode-languageserver-types";
 
-function formatTestPLI(code: string) {
+function formatPLICode(code: string, padding: number = 0): string {
```
montymxb, line 23: Should we have a default padding of `1` here instead of zero?

montymxb (commented):
One small point (partial question), as is looks like we might be able to run with it as is. Also on board with factoring out programs into a corpus which can run through parsing/linking/validation tests later on.

Issue #77, opened 2025-04-03
[Feature] Indicate "skipped" code in the editor


msujew:
Preprocessor statements like `IF` and `SKIP` allow to ignore certain parts of the code. We should highlight those with a lower opacity.

ssmifi:
I'll have a look.

PR #114, opened 2025-04-22
Add skipped code block visualization

ssmifi:
Add custom LS message to display skipped code blocks due to %skip and %if preprocessor instructions.

To separate concerns, the lexer result returns a list of statements (that are not considered pure PLI) and the LS continues from there. Please let me know if you agree with this approach or if you prefer a more explicit approach w.r.t. skipped code blocks inside the preprocessor before I continue with tests and the playground. 
We also can discuss alternatives in 1on1.

Closes https://github.com/zowe/zowe-pli-language-support/issues/77


ssmifi:
Result of the meeting: Generally, we agreed on this approach. However, we will postpone the merge until https://github.com/zowe/zowe-pli-language-support/pull/92 is completed and are going  to keep a dedicated list of skipped ranges in the preprocessor.

packages/vscode-extension/src/extension/decorators.ts
```diff
@@ -0,0 +1,70 @@
...
+    "pli/skippedCode",
+    (params: { uri: string; ranges: Range[] }) => {
+      if (!settings.skippedCodeEnabled) {
+        return;
+      }
+
+      const editor = vscode.window.visibleTextEditors.find(
+        (e) => e.document.uri.toString() === params.uri,
+      );
+      if (editor) {
```
msujew, line 36: Suggestion: Since multiple editors are possible for the same file, you should use `filter` instead. The decoration can then be applied to every editor instead of just the first.

packages/vscode-extension/src/extension/decorators.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+  );
+
+  vscode.window.onDidChangeActiveTextEditor((editor) => {
+    if (!editor || !settings.skippedCodeEnabled) return;
+    const ranges = skippedRangesByUri.get(editor.document.uri.toString());
+    if (ranges) {
+      editor.setDecorations(skippedCodeDecorationType, ranges);
+    }
+  });
```
msujew, line 56: Question: Why is this required?

packages/vscode-extension/src/extension/decorators.ts
```diff
@@ -0,0 +1,70 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import * as vscode from "vscode";
+import { LanguageClient } from "vscode-languageclient/node.js";
+import { Settings } from "./settings";
+import { Range } from "vscode-languageserver-types";
```
msujew, line 15: Suggestion: The `vscode-languageclient` package exports everything from `vscode-languageserver-types`. We don't need a specific dependency to `vscode-languageserver-types`.
```suggestion
import { LanguageClient, Range } from "vscode-languageclient/node.js";
import { Settings } from "./settings";
```

msujew (commented):
I've noticed a few issues when changing the settings related to how change this applies the decorations: When disabling and re-enabling the feature in the settings, the editor must be focused for the changes to take effect. The same thing happens when changing the opacity. 

This doesn't work well, as users often times use multi-column layouts where multiple editors are visible at a time. Only updating the decorations when the user focuses the editor leads to outdated editor decorations.

packages/vscode-extension/src/extension/decorators.ts
ssmifi: I changed the way the editor react to setting changes to better support multiple editors. However, I still need `onDidChangeVisibleTextEditors` at least to reapply decorators after editors become visible again. Maybe I miss something here? I will discuss this with @montymxb. See you soon!

packages/language/src/preprocessor/pli-preprocessor-parser.ts
```diff
@@ -796,6 +805,7 @@ export class PliPreprocessorParser {
       )
     ) {
       lineCount = parseInt(state.last!.image, 10);
+      statement.lineCount = lineCount;
```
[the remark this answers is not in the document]
ssmifi: Sounds good. So, do you open a ticket with your findings w.r.t. the parser rule issue?

packages/language/src/preprocessor/pli-preprocessor-interpreter-state.ts
```diff
@@ -16,6 +16,8 @@ import { PreprocessorTokens } from "./pli-preprocessor-tokens";
 import { CompilationUnitTokens } from "../workspace/compilation-unit";
 import * as ast from "../syntax-tree/ast";
 
+export type EvaluationResults = Map<ast.IfStatement, boolean | undefined>;
```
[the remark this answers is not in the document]
ssmifi: Right. We could store the evaluation result in the corresponding node. We should consult Mark about his plans regarding the PP integration.

packages/language/src/preprocessor/pli-preprocessor-parser.ts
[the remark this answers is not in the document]
montymxb: Spun this off into #122

montymxb (approved):
:octocat: Awesome, let's pull this in.

Issue #19, opened 2025-01-20
Scoping of PL/I Structure


asatklichov:
Currently PL/I structures works fine with hard coded levels 1-2,..., and each variable isolated (hover over not shows structure)
It should work with any level order and  should be recognized as a single entity (each hover-over should show correct structure - shows its subleveles) 

Picture-1:
<img width="469" alt="Image" src="https://github.com/user-attachments/assets/2b41c728-e564-43f5-8e7c-37eaea0e3454" />

Picture-2:
<img width="625" alt="Image" src="https://github.com/user-attachments/assets/be62668b-e1aa-4501-adbf-3e28298f6c99" />

msujew:
Yes, this has been resolved for a while now.

PR #116, opened 2025-04-23
Improve linking

didrikmunther:
**This PR is getting quite big and is creeping, so I'll mark this as ready and continue working on the remaining issues in another PR**

Relates to #19 

Currently, we're failing on a couple of test cases:

### ~~1. Must infer partially qualified names~~ ✅

In this example, `b2` should link correctly because of where its child qualifier `c` links.

```
 DCL 1 A,
        2 <|b1:B|>,
          3 <|c:C|> CHAR(8) VALUE("C");

 DCL 1 A2,
        2 <|b2:B|>,
          3 <|d:D|> CHAR(8) VALUE("D");

 PUT (<|b1>B.<|c>C);
 PUT (<|b2>B.<|d>D);
```

### 2. Should not qualify for factorized structured names

In this example, `d` should never be able to be referenced, since it's nested behind a factorized name in a structure. However, it does link currently.

```
 DCL 1 A,
       2 (B,C),
          3 <|d:D|>;

 PUT(<|d>D);
```

### 3. Star name in structure should be qualifiable

In this example, `b` is fully qualified, but still links to the `B` nested inside the start, which is wrong.

```
 DCL 1 A,
        2 *,
          3 B CHAR(8) VALUE("B"),
        2 <|b:B|> CHAR(8) VALUE("B2");

 PUT(A.<|b>B);
```

### 1. Redeclaration must fail

In this example, the redeclaration should throw an error (IBM1306I E: Repeated declaration of A is invalid and will be ignored.)

```
 DCL A CHAR(8) INIT("A");
 DCL A CHAR(8) INIT("A2");
```

### 2. Redeclaration of label must fail

In this example, the redeclaration should throw an error (IBM1916I S: The PROCEDURE/ENTRY OUTER has already been defined.)

```
 OUTER: PROCEDURE;
 END OUTER;
 OUTER: PROCEDURE;
 END OUTER;
```

### 3. Repeated declaration of label is invalid

Should produce: IBM1306I E: Repeated declaration of A is invalid and will be ignored.

```
 A: PROCEDURE;
 END A;
 DCL A CHAR(8) INIT("A");
```

### 4. Factoring of level numbers into declaration lists containing level numbers is invalid

Should produce: IBM1376I E: Factoring of level numbers into declaration lists containing level numbers is invalid. The level numbers in the declaration list will be ignored.

```
 DCL 1 A,
       2(B, 3 C, D) (3,2) binary fixed (15);
```

### ~~5. Structure level greater than 255 is invalid~~ ✅ 

Should produce: IBM1363I E: Structure level greater than 255 specified. It will be replaced by 255.

```
 DCL 1 A,
       256 B;
```

### ~~6. Ambiguous reference should report error~~

Should produce: IBM1881I S: The reference B is ambiguous.

```
 DCL 1 A1,
     2 B CHAR(8) VALUE("B1");
 DCL 1 A2,
     2 B CHAR(8) VALUE("B2");

 PUT(B);
```

### ~~7. Unused procedure label should warn~~ ✅

Should produce: IBM1213I W: The PROCEDURE OUTER is not referenced.

```
 OUTER: PROCEDURE;
 END OUTER;
```

### DIFF
--- a/packages/language/src/language-server/types.ts
+++ b/packages/language/src/language-server/types.ts
@@ -56,6 +56,10 @@ export function rangeToLSP(
   };
 }
 
+export function tokenToUri(token: IToken): string {
+  return token.payload.uri._formatted; // TODO: Is this safe?
+}
+
 export function tokenToRange(token: IToken): Range {
   return {
     start: token.startOffset,
--- a/packages/language/src/language-server/types.ts
+++ b/packages/language/src/language-server/types.ts
@@ -56,8 +56,8 @@ export function rangeToLSP(
   };
 }
 
-export function tokenToUri(token: IToken): string {
-  return token.payload.uri._formatted; // TODO: Is this safe?
+export function tokenToUri(token: IToken): string | undefined {
+  return token.payload?.uri?.toString();
 }
 
 export function tokenToRange(token: IToken): Range {
--- a/packages/language/src/linking/resolver.ts
+++ b/packages/language/src/linking/resolver.ts
@@ -10,12 +10,7 @@
  */
 
 import { IToken } from "chevrotain";
-import {
-  Diagnostic,
-  Location,
-  Severity,
-  tokenToRange,
-} from "../language-server/types";
+import { Diagnostic, Location, tokenToRange } from "../language-server/types";
 import { TokenPayload } from "../parser/abstract-parser";
 import {
   MemberCall,
@@ -37,7 +32,6 @@ import {
   PliValidationAcceptor,
   PliValidationBuffer,
 } from "../validation/validator";
-import * as PLICodes from "../validation/messages/pli-codes";
 
 export class ReferencesCache {
   private list: Reference[] = [];
@@ -163,16 +157,21 @@ function resolveReference(
   }
 
   const symbols = scope.getSymbols(qualifiedName);
-  // Ambiguous reference, emit a severe diagnostic.
-  // TODO: Currently only emitting on the last member call symbol (`reference.token`)
-  // We want to underline the entire qualified name.
-  if (symbols.length > 1) {
-    const fullName = qualifiedName.toReversed().join(".");
-    acceptor(Severity.S, PLICodes.Severe.IBM1881I.message(fullName), {
-      code: PLICodes.Severe.IBM1881I.fullCode,
-      range: tokenToRange(reference.token),
-      uri: reference.token.payload?.uri?.toString() ?? "",
-    });
+
+  const isAmbiguous = symbols.length > 1;
+  if (isAmbiguous) {
+    /**
+     * @didrikmunther
+     * @WILLFIX: We're colliding with the DEACTIVATE functionality in the preprocessor.
+     */
+    // const fullName = qualifiedName.toReversed().join(".");
+    // // TODO: Currently only emitting on the last member call symbol (`reference.token`)
+    // // We want to underline the entire qualified name.
+    // acceptor(Severity.S, PLICodes.Severe.IBM1881I.message(fullName), {
+    //   code: PLICodes.Severe.IBM1881I.fullCode,
+    //   range: tokenToRange(reference.token),
+    //   uri: reference.token.payload?.uri?.toString() ?? "",
+    // });
   }
 
   // We take the first symbol, even if there are multiple matching.
--- a/packages/language/src/linking/resolver.ts
+++ b/packages/language/src/linking/resolver.ts
@@ -12,7 +12,12 @@
 import { IToken } from "chevrotain";
 import { Location, tokenToRange } from "../language-server/types";
 import { TokenPayload } from "../parser/abstract-parser";
-import { Reference, SyntaxKind, SyntaxNode } from "../syntax-tree/ast";
+import {
+  MemberCall,
+  Reference,
+  SyntaxKind,
+  SyntaxNode,
+} from "../syntax-tree/ast";
 import { binaryTokenSearch } from "../utils/search";
 import {
   getNameToken,
@@ -22,6 +27,7 @@ import {
 } from "./tokens";
 import { URI } from "../utils/uri";
 import { CompilationUnit } from "../workspace/compilation-unit";
+import { QualifiedSyntaxNode } from "./symbol-table";
 
 export class ReferencesCache {
   private list: Reference[] = [];
@@ -57,7 +63,7 @@ export class ReferencesCache {
 }
 
 // Returns the qualified name in reverse order, e.g. "A.B.C" -> ["C", "B", "A"]
-export function getQualifiedName(reference: Reference): string[] {
+function getQualifiedName(reference: Reference): string[] {
   if (reference.owner.container?.kind === SyntaxKind.MemberCall) {
     const memberCall = reference.owner.container;
     if (memberCall.previous?.element?.ref) {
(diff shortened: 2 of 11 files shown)


packages/language/src/language-server/types.ts
```diff
@@ -56,6 +56,10 @@ export function rangeToLSP(
   };
 }
 
+export function tokenToUri(token: IToken): string {
+  return token.payload.uri._formatted; // TODO: Is this safe?
+}
```
msujew, line 61: Suggestion: You can just use `toString()`. `_formatted` is just the cache for the `toString()` method inside of [`Uri`](https://github.com/microsoft/vscode-uri/blob/edfdccd976efaf4bb8fdeca87e97c47257721729/src/uri.ts#L410-L420). Note that as of my latest change, the URI is now optional and might return `undefined`.
```suggestion
export function tokenToUri(token: IToken): string | undefined {
  return token.payload?.uri?.toString();
}
```

packages/language/src/linking/symbol-table.ts
```diff
@@ -276,22 +304,38 @@ export class ScopeCache {
...
+      ...d,
+    });
+  };
+
+  // Iterate over the PLI program creating the symbol table.
+  new SymbolTableIterator(scopeCaches.preprocessor, references, acceptor)
+    .iterate(unit.preprocessorAst, scope);
+
+  new SymbolTableIterator(scopeCaches.regular, references, acceptor)
+    .iterate(unit.ast, scope);
```
msujew, line 336: Question: Does it really make sense to have a dedicated class instance here that we instantly dispose of after running one function? IMO we could factor this into a simple function.

didrikmunther: You are right! Just felt quite verbose with a lot of parameters

packages/language/src/language-server/types.ts
didrikmunther: Ok! Should I just pass along an empty URI to the acceptor then?

montymxb: Yep that's acceptable. When we promote this up to a Diagnostic we just drop the ones that lack an associated URI.

packages/language/test/linking.test.ts
```diff
@@ -120,30 +136,76 @@ describe("Linking tests", () => {
...
+          3 K, // Should be skipped in qualification
+            4 <|d:D|> CHAR(8) VALUE("D");
+
+ PUT (<|b1>B.<|c>C);
+ PUT (<|b2>B.<|d>D);`));
+
+    /**
+     * TODO: This should fail
+     */
+    test.skip("Should error when using factorized names in structures", () => {
```
montymxb, line 198: Can add a WillFix if we want to resolve this later.

packages/language/test/linking.test.ts
```diff
@@ -181,17 +243,119 @@ describe("Linking tests", () => {
 
  PUT(A.<|b>B);`));
 
-    /**
-     * TODO: Implement star handling in structured names
-     */
-    test.skip("Star name in structure should be qualifiable", () =>
+    test.skip("Star name in structure should result in partial qualification", () =>
```
montymxb, line 246: Same here, would be good to add a WillFix and a very short description of the limitation

packages/language/test/linking.test.ts
```diff
@@ -181,17 +243,119 @@ describe("Linking tests", () => {
...
+ END LABL;
+
+ PUT(<|a>A);
+ CALL LABL;
+
+ DCL <|a:A|> CHAR(8) INIT("A");`));
+  });
+
+  /**
+   * These test probably should not in the linking tests, but I'm putting them here for reference for now.
```
montymxb, line 277: Mark @didrikmunther here for ref

packages/language/test/linking.test.ts
```diff
@@ -181,17 +243,119 @@ describe("Linking tests", () => {
...
+ OUTER: PROCEDURE;
+ END OUTER;`);
+      assertDiagnostic(doc, {
+        code: PLICodes.Warning.IBM1213I.fullCode,
+        severity: Severity.W,
+      });
+    });
+
+    /**
+     * TODO: Should this be tested here or in the validation step?
```
montymxb, line 303: Given the diagnostic is available at this stage, it's ok to check it here as well.

packages/language/test/linking.test.ts
```diff
@@ -181,17 +243,119 @@ describe("Linking tests", () => {
...
+     2 B CHAR(8) VALUE("B2");
+ PUT(B);
+ `);
+      assertDiagnostic(doc, {
+        code: PLICodes.Severe.IBM1881I.fullCode,
+        severity: Severity.S,
+      });
+    });
+
+    test.skip("Repeated declaration of label is invalid", () => {
```
montymxb, line 329: willfix

packages/language/test/linking.test.ts
```diff
@@ -181,17 +243,119 @@ describe("Linking tests", () => {
...
+ A: PROCEDURE;
+ END A;
+ DCL A CHAR(8) INIT("A");`);
+      assertDiagnostic(doc, {
+        code: PLICodes.Error.IBM1306I.fullCode,
+        severity: Severity.E,
+      });
+    });
+
+    test.skip("Factoring of level numbers into declaration lists containing level numbers is invalid", () => {
```
montymxb, line 340: willfix

montymxb (commented):
Some points here and there, and a question or two. After changes we can pull this in and continue with revisions on top.

packages/language/src/linking/resolver.ts
```diff
@@ -71,45 +91,119 @@ export function getQualifiedName(reference: Reference): string[] {
...
+
+function resolveReference(
+  unit: CompilationUnit,
+  reference: Reference,
+  acceptor: PliValidationAcceptor,
+) {
+  if (
+    reference === null ||
+    reference.node === null ||
+    reference.node !== undefined
```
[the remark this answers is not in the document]
didrikmunther: It is the same logic as in the original version: https://github.com/zowe/zowe-pli-language-support/blob/cf1ac2ed9a007f09f7d758526b086465a42d8d06/packages/language/src/linking/resolver.ts#L78-L84

But you're right, the first check `reference === null` is no longer necessary, I've removed this now.

My interpretation is this:

1. If `reference.node === null`, we've explicitly marked that this reference does not resolve to anything.
2. If `reference.node !== undefined`, we've already linked this in a previous step, and do not need to relink it.

Technically we could just check for `reference.node !== undefined`, since that includes case (1) as well.

packages/language/src/linking/resolver.ts
```diff
@@ -71,45 +91,119 @@ export function getQualifiedName(reference: Reference): string[] {
...
+  // TODO: Currently only emitting on the last member call symbol (`reference.token`)
+  // We want to underline the entire qualified name.
+  if (symbols.length > 1) {
+    const fullName = qualifiedName.toReversed().join(".");
+    acceptor(Severity.S, PLICodes.Severe.IBM1881I.message(fullName), {
+      code: PLICodes.Severe.IBM1881I.fullCode,
+      range: tokenToRange(reference.token),
+      uri: reference.token.payload?.uri?.toString() ?? "",
+    });
+  }
```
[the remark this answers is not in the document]
didrikmunther: It doesn't work in the preview branch, but it works locally for some reason...

CleanShot 2025-05-08 at 08 20 22@2x

packages/language/src/linking/symbol-table.ts
```diff
@@ -301,16 +369,26 @@ function recursivelySetContainer(node: SyntaxNode) {
   });
 }
 
-// This function iterates over the syntax tree and builds the symbol table.
-function iterate(
-  node: SyntaxNode,
+// This function determines if a node should create a new scope.
+function shouldNodeCreateNewScope(parent: SyntaxNode): boolean {
+  switch (parent.kind) {
```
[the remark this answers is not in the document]
didrikmunther: I suspect we might find more cases of scope-creating nodes in the future. The switch statement is intended as a structure of how the logic could be expanded. What do you think?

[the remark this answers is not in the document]
montymxb: I see, good point. Let's leave it as is then for now. There's also the matter of package & begin blocks which also create scopes.

packages/language/src/linking/resolver.ts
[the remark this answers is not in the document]
montymxb: I should clarify I'm referring to the compiler output, which differs from the LS here.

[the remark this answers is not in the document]
didrikmunther: Yes of course, the example program I showed already suffers from a redeclaration. This is a better example for ambiguity:

```
 DCL 1 A1,
     2 B CHAR(8) VALUE("B1");
 DCL 1 A2,
     2 B CHAR(8) VALUE("B2");

 PUT(B);
 ```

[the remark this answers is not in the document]
Comment (montymxb) [reply packages/language/src/linking/resolver.ts:176]: Ahh yeah that one matches up as expected, nice! However, there's one more issue here with the **include.pli** example:
```c
   %DECLARE PAYROLL CHARACTER;
   %PAYROLL = 'PAY_ROLL'; // <-- flags as ambiguous
   %INCLUDE "included.pli";
   %DEACTIVATE PAYROLL; // <-- and here
   %INCLUDE "included.pli";
   PAY_ROLL = 0; 
   PAYROLL = 1; // <-- and here
```
Dropping the 2nd include removes the problem. It appears we're conflicting the pre-processor ref w/ the pli reference somewhere.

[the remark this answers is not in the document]
didrikmunther: This will require quite a bit of work I think. I'm deactivating ambiguity for now.

PR #118, opened 2025-04-23
Add closed file navigation support

ssmifi:
Closes https://github.com/zowe/zowe-pli-language-support/issues/73

Should probably be verified and merged after https://github.com/zowe/zowe-pli-language-support/pull/117


packages/language/test/lsp/definition-requests.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  afterEach(() => {
+    FileSystemProviderInstance.readFileSync = originalFs;
+  });
+
+  function expectDefinition(
+    programs: string[],
+    includeURIs: string[],
+    expectAllRanges: boolean = true,
+  ) {
```
msujew, line 33: Suggestion: We already have a `expectGotoDefinition` request testing utility in `test/utils.ts`. If it doesn't work for multiple files yet, we should adjust that function, instead of creating a new one.

packages/language/src/language-server/connection-handler.ts
```diff
@@ -54,116 +63,183 @@ export function startLanguageServer(connection: Connection): void {
...
+
+  const offset = textDocument.offsetAt(params.position);
+  const renameLocations = renameRequest(unit, parsedUri, offset);
+  const changes = mapValues(renameLocations, (locations, key) => {
+    let doc = TextDocuments.get(key);
+
+    if (!doc) {
+      const content = FileSystemProviderInstance.readFileSync(URI.parse(key));
+      doc = TextDocument.create(key, "pli", 0, content);
+      TextDocuments.set(doc);
```
msujew, line 228: Suggestion: We should have some kind of try-catch for this, as any file-system error will prevent the request from completing.

packages/language/test/lsp/definition-requests.test.ts
ssmifi: I assume that function was refactored in the meantime. For now, I’d prefer not to add it to the existing linking tests, mainly because this PR needs to test the complete callback handler, not just the `definitionRequest` function. 

That said, I’d generally like to test all LSP callbacks directly, rather than only the functions they invoke—or perhaps even start the language server for the tests. But that’s probably something to address later. I can add that to https://github.com/zowe/zowe-pli-language-support/issues/121.

packages/language/src/language-server/text-documents.ts
```diff
@@ -208,8 +208,28 @@ export class NormalizedTextDocuments<T extends { uri: string }>
...
-  public get(uri: string | URI): T | undefined {
-    return this._syncedDocuments.get(UriUtils.normalize(uri));
+  public get(
+    uri: string | URI,
+    options?: { loadFromURI?: boolean },
+  ): T | undefined {
+    let syncedDocument = this._syncedDocuments.get(UriUtils.normalize(uri));
+    if (syncedDocument === undefined && options?.loadFromURI) {
+      try {
+        const uriName = typeof uri === "string" ? uri : uri.toString();
```
msujew, line 218: Suggestion: We are already calling `UriUtils.normalize`. This is exactly for creating a string from an `string | URI` typed variable. You should reuse the result from that instead of doing `uri.toString()` again.

packages/language/test/lsp/definition-requests.test.ts
```diff
@@ -0,0 +1,170 @@
...
+      onDocumentHighlight: () => {},
+      onDocumentSymbol: () => {},
+      onRenameRequest: () => {},
+      onWorkspaceSymbol: () => {},
+      listen: () => {},
+      dispose: () => {},
+      sendDiagnostics: () => {},
+    };
+
+    startLanguageServer(mockConnection as any);
```
msujew, line 57: Question: Why do we start the language server and then do nothing with the connection?

packages/language/test/lsp/definition-requests.test.ts
msujew: I don't really follow. Both the new `expectDefinition` test and the already existing `expectLinks` functions use the same implementation.

> That said, I’d generally like to test all LSP callbacks directly, rather than only the functions they invoke—or perhaps even start the language server for the tests.

Yes, you would likely need to start the language server and provide a language client implementation. I.e. you would need to write a vscode frontend tests. That's much more involved than whatever we're doing here. We're not even doing that in Langium, because generally, the fact that we're wiring the `Connection` object up correctly is pretty trivial to test in manual testing and rarely breaks.

montymxb (commented):
Some general feedback. Beyond that testing this out looks really nice, great to see that we can jump to declarations in other files without needing to open them first.

packages/language/test/lsp/definition-requests.test.ts
msujew: My bad, the original function is `expectLinks`, but it uses the `definitionRequest` internally.

ssmifi: The rationale was to actually test the `onDefinition` callback that is used in the language server. But I will simplify this and merge it with `expectLinks` then.

msujew (approved):
Looks quite good, thank you. Just a short question before the merge, see below.

packages/language/test/utils.ts
```diff
@@ -425,13 +442,24 @@ export class TestBuilder {
...
+      this.files["file:///main.pli"] = replaceNamedIndices(textOrFiles);
+    } else {
+      for (const file of textOrFiles) {
+        this.files[file.uri] = replaceNamedIndices(file.content);
+      }
+
+      for (const [uri, file] of Object.entries(this.files)) {
+        const doc = TextDocument.create(uri, "pli", 1, file.output);
+        TextDocuments.set(doc);
+      }
```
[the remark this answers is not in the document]
ssmifi: I added the possibility to provide a `FileSystemProvider` to the `TestBuilder` so that tests can use a `VirtualFileSystem` to provide the files. The Preprocessor then loads the files from the file system and does not rely on `TextDocuments` any longer.

montymxb (approved):
:octocat: Approving additional changes for adding `writeFileSync` to the base fs provider interface.

Issue #78, opened 2025-04-03
[Feature] Indicate margin lines in editor


msujew:
Currently, it's rather hard to tell where the actual margin lines are (i.e. where parsing starts and ends on each line). We should somehow highlight those in the editor.

msujew:
No longer blocked :)

PR #123, opened 2025-04-28
Add margin indicators

ssmifi:
Closes https://github.com/zowe/zowe-pli-language-support/issues/78

I implemented two complementary methods to visualize margins:
- VS Code rulers: These rulers are visible across the entire document and are configured globally for the language. As a result, they appear in all open PL/I editors, even if the margin settings differ between files.
- Decorators: This method applies individual decorators when the configured margins are exceeded, allowing for file-specific margin visualization.

Both features are toggleable via the settings and are compatible with each other.
If working with PL/I files that have differing margins turns out to be rare, we can consider disabling the less-used feature.

### DIFF
--- a/packages/language/src/language-server/margin-indicator.ts
+++ b/packages/language/src/language-server/margin-indicator.ts
@@ -1,0 +1,50 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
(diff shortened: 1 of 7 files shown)


ssmifi:
Updated according to discussion: default and auto settings with rulers are sufficient (for now).

montymxb (approved):
:octocat: Looks good to pull in!

Issue #122, opened 2025-04-28
Pre-processor %SKIP Parser Issue


montymxb:
Pre-processor `%SKIP` is parsing incorrectly, i.e. without parentheses wrapping the sole argument. We expect this to parse, `%SKIP (3);`, but instead we parse this `%SKIP 3;`.

Back checking against the compiler confirms this won't pass, just to be certain. A patch will need to address the parser discrepancy + impacted tests, but not much more than that.

PR #124, opened 2025-04-29
Fix missing parentheses at skip directive

ssmifi:
Closes https://github.com/zowe/zowe-pli-language-support/issues/122

Fixed the missing parentheses and added two tests to confirm behavior.


montymxb (approved):
:octocat: Superb, good to go.

PR #125, opened 2025-05-07
fix: Explicitly copy declared items for shifting

didrikmunther:
Because we're using `.shift()` in the `pop()` method, `DeclaredItem`s were actually disappearing from the AST after the symbolization step. We now explicitly copy the items array.


ssmifi (approved):
LGTM. Thanks for the quick response!

Issue #121, opened 2025-04-28
[Test] Consolidate lifecycle tests


ssmifi:
As mentioned in [#113](https://github.com/zowe/zowe-pli-language-support/pull/113), we should consolidate all PL/I programs into a single, extendable list that can be used across the lifecycle tests, including both parser and linking tests.

ssmifi:
Since we added a powerful way to integrate PLI programs into our tests (https://github.com/zowe/zowe-pli-language-support/pull/169) and migrating the tests as topics come up, I'm closing this one.

PR #126, opened 2025-05-07
Improve symbol table test

ssmifi:
Should be merged after https://github.com/zowe/zowe-pli-language-support/pull/125

Improves the symbol table test:

1. Adds a sanity check to catch issues where a node has a non-null container but is no longer reachable by the AST iterator—an indication that symbol table generation may have corrupted the AST structure.
2. Renames `assertValidSymbolTable` to `generateAndAssertValidSymbolTable` to clarify that the function not only validates but also generates the symbol table. This helps prevent confusion with other validation tests that also use `assertNoParseErrors` but operate over the entire lifecycle. The function is now explicitly included in the parser and parser-doc tests. While this can be refactored later (see https://github.com/zowe/zowe-pli-language-support/issues/121), the current approach is cleaner for now.

### DIFF
--- a/packages/language/test/extracted-doc-cases.test.ts
+++ b/packages/language/test/extracted-doc-cases.test.ts
@@ -20,7 +20,12 @@
  */
 
 import { test } from "vitest";
-import { assertNoParseErrors, parse, parseStmts } from "./utils";
+import {
+  assertNoParseErrors,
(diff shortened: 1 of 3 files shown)


montymxb (approved):
:octocat: approved!

PR #128, opened 2025-05-08
Add AST debug annotator

didrikmunther:
An AST node's `kind` is a number. While performant, it is hard to debug. This PR adds a debuggable `_kind`, containing the string name of the AST node kind:

CleanShot 2025-05-08 at 11 30 35

### DIFF
--- a/packages/language/test/utils.ts
+++ b/packages/language/test/utils.ts
@@ -18,6 +18,22 @@ interface AssertNoDiagnosticsOptions {
   ignoreSeverity?: Severity[];
 }
 
+const SyntaxKindReverseLookup: Map<SyntaxKind, string> = new Map(
+  Object.values(SyntaxKind)
+    .filter((key) => typeof key === "number")
+    .map((key) => [key, SyntaxKind[key]]),
+);
+
+/**
+ * This function assigns a `_kind: string` to each node that
+ * alleviates debugging by allowing the user to see the kind
+ * of the node in the debugger.
+ */
+export function assignDebugKind(node: SyntaxNode) {
+  (node as any)._kind = SyntaxKindReverseLookup.get(node.kind);
+  forEachNode(node, assignDebugKind);
+}
+
 function expectNoDiagnostics(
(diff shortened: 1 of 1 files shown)


ssmifi:
I added something similar to https://github.com/zowe/zowe-pli-language-support/pull/113, but it was eventually removed, because of code duplication of the lifecycle before we agreed on https://github.com/zowe/zowe-pli-language-support/issues/121 and it was not really required anymore for the actual test.
Since your solution only interacts with the existing test functions, I would appreciate something like this. 👍

packages/language/test/utils.ts
```diff
@@ -18,6 +18,22 @@ interface AssertNoDiagnosticsOptions {
...
+    .map((key) => [key, SyntaxKind[key]]),
+);
+
+/**
+ * This function assigns a `_kind: string` to each node that
+ * alleviates debugging by allowing the user to see the kind
+ * of the node in the debugger.
+ */
+export function assignDebugKind(node: SyntaxNode) {
+  (node as any)._kind = SyntaxKindReverseLookup.get(node.kind);
```
montymxb, line 33: Why not call it `_debugKind`? In case we see if somewhere else that would be a dead giveaway that it's somewhere we don't expect it to be.

packages/language/test/utils.ts
```diff
@@ -18,6 +18,22 @@ interface AssertNoDiagnosticsOptions {
...
+    .filter((key) => typeof key === "number")
+    .map((key) => [key, SyntaxKind[key]]),
+);
+
+/**
+ * This function assigns a `_kind: string` to each node that
+ * alleviates debugging by allowing the user to see the kind
+ * of the node in the debugger.
+ */
+export function assignDebugKind(node: SyntaxNode) {
```
montymxb, line 32: assignDebugKinds

montymxb (approved):
:octocat: A couple general non-blocking points as suggestions. For testing purposes I would also appreciate this, rather than having to back-check the kinds to their corresponding type in each case.

PR #129, opened 2025-05-08
Add ambiguity error

didrikmunther:
The preprocessor and PL/I linker instances were sharing the same root scope, leading to their symbols being linked to each other. This PR separates their root scope, and enables checking regular PL/I variables for ambiguity.

### DIFF
--- a/packages/language/test/linking.test.ts
+++ b/packages/language/test/linking.test.ts
@@ -287,6 +287,14 @@ describe("Linking tests", () => {
  %DEACTIVATE <|3:PAYROLL|>;
  PUT(<|1><|2><|3>PAYROLL);`),
     );
+
+    test("Must link correctly when collision with preprocessor variables", () =>
+      expectLinks(`
+ %DECLARE ABC CHARACTER;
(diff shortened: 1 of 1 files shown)


packages/language/test/linking.test.ts
```diff
@@ -274,6 +274,21 @@ describe("Linking tests", () => {
  DCL <|a:A|> CHAR(8) INIT("A");`));
   });
 
+  describe("Preprocessor", () => {
+    /**
+     * This test intentionally fails in order to test that a native
+     * PL/I variable does not link to a preprocessor variable.
+     */
+    test.fails("Must not link between PL/I and preprocessor variables", () =>
```
montymxb, line 282: Anti-case is good to avoid to avoid a regression on this issue, but we should also have another case that asserts correct linking for both the pre-processor & pli vars in the same example.

montymxb (commented):
One request for an additional test, rest looks good!

montymxb (approved):
:octocat: looks good!

PR #130, opened 2025-05-08
Add redeclaration error reporting

didrikmunther:
- Adds two new error reportings
  - IBM1916I
  - IBM1306I
- Adds a test utility `TestBuilder` to allow for checking specific location of diagnostics.

### DIFF
--- a/packages/language/src/validation/validator.ts
+++ b/packages/language/src/validation/validator.ts
@@ -17,11 +17,12 @@ import {
   Range,
   Severity,
   tokenToRange,
+  tokenToUri,
 } from "../language-server/types";
 import { ReferencesCache } from "../linking/resolver";
 import { isValidToken } from "../linking/tokens";
 import { TokenPayload } from "../parser/abstract-parser";
-import { LabelPrefix, SyntaxKind, SyntaxNode } from "../syntax-tree/ast";
+import { SyntaxKind, SyntaxNode } from "../syntax-tree/ast";
 import { forEachNode } from "../syntax-tree/ast-iterator";
 import {
   PliValidationChecks,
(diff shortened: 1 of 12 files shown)


packages/language/src/validation/validator.ts
```diff
@@ -163,36 +166,58 @@ export function parserErrorsToDiagnostics(
...
+     *
+     * TODO @didrikmunther: A LabelPrefix without a procedure option should throw a IBM1911I error instead.
+     * Currently, we don't have a way to know if a label is a statement label or a procedure label.
+     */
+    if (symbol.node.kind === SyntaxKind.LabelPrefix) {
+      diagnostics.push({
+        severity: Severity.S,
+        message: PLICodes.Severe.IBM1916I.message(symbol.name),
+        code: PLICodes.Severe.IBM1916I.fullCode,
+        uri: tokenToUri(nameToken) ?? "",
```
msujew, line 197: Minor: Since we cannot show the diagnostic, if we don't know about the `uri`, can we have a guard against that, similar to the `if (!nameToken)` above (or in the same `if`)? That way we don't produce the diagnostic in the first place.

msujew (approved):
Looks very nice. I specifically like the new `TestBuilder`. Just a minor comment, see below.

packages/language/src/validation/validator.ts
didrikmunther: We should abstract this into some kind of `DiagnosticGenerator`, which doesn't produce a diagnostic if the URI is unavailable. This way, we could also avoid duplicating the `PLICode` multiple times (which I think is unnecessarily verbose). I'll fix this in another PR, but I have put a guard against the URI for the time being.

Issue #119, opened 2025-04-24
[feature-lsp] Support document symbol requests


msujew:
Related to providing support for breadcrumbs & outlines.

Issue #120, opened 2025-04-24
[feature-lsp] Support workspace symbol requests


msujew:
Related to providing support for breadcrumbs & outlines.

PR #131, opened 2025-05-08
Add document/workspace symbol request handler

ssmifi:
Closes https://github.com/zowe/zowe-pli-language-support/issues/119 https://github.com/zowe/zowe-pli-language-support/issues/120

- Implemented the handler for document symbol requests. The current implementation supports **procedures**, **variables**, **constants**, **structs**, and other **label**-prefixed symbols. The design allows for easy extension via the symbol builder.
- Added corresponding tests for document symbol handling.
- Additionally, updated (and simplified) the symbolic link tests to use the new `replaceNamedIndices` function.
---
- Implemented the handler for workspace symbol requests. 
- Added test for workspace symbol requests.
- Added caching for both requests (and margins and skipped code) to avoid recomputation.
---
Preview:
image


packages/language/src/workspace/compilation-unit.ts
```diff
@@ -155,8 +173,7 @@ export class CompletionUnitHandler {
           diagnostics: fileDiagnostics ?? [],
         });
       }
-      skippedCode(connection, event.document.uri, unit, document);
-      marginIndicator(connection, event.document.uri, unit, document);
+      unit.requestCaches.revalidateAll({ connection, unit });
```
montymxb, line 176: I presume this should be `invalidateAll`?

montymxb (commented):
General comments & feedback. Overall the cache approach appears to be a bit overkill for what we need in this regard, but a pared down version should be fine.

Additionally the calculation seems to be off, especially for synthesized elements from the pre-processor. Such as for the preprocessor *do.pli* example if you check that one out.

packages/language/src/language-server/document-symbol-request.ts
```diff
@@ -0,0 +1,347 @@
...
+  new ProcedureSymbolBuilder(),
+  new DeclareSymbolBuilder(),
+  new LabelSymbolBuilder(),
+];
+
+export function documentSymbolRequest(
+  compilationUnit: CompilationUnit,
+  uri: URI,
+): DocumentSymbol[] {
+  if (!compilationUnit.requestCaches.get("documentSymbols").isCached()) {
```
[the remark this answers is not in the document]
montymxb: Just revising this a bit after review. Even if the caching logic is simplified, I would still hide the cache used by the compileUnit from outside access, and abstract it away as much as possible.

packages/language/src/language-server/margin-indicator.ts
```diff
@@ -26,25 +25,18 @@ export const MarginIndicatorNotification =
...
 
-  if (compilationUnit) {
-    if (compilationUnit.compilerOptions.margins) {
-      indicator.m = compilationUnit.compilerOptions.margins.m;
-      indicator.n = compilationUnit.compilerOptions.margins.n;
-    } else {
-      indicator.m = 2;
-      indicator.n = 72;
-    }
+  const cachedMargins = compilationUnit.requestCaches.get("margins").get();
```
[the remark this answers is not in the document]
ssmifi: Margins are stored per uri in the client. Besides, the rulers in vs code, which we use for the margins, are set per language, so if different units use different margins, they are all visible in all editors (unless set to no or default margins in the settings).

ssmifi:
> Additionally the calculation seems to be off, especially for synthesized elements from the pre-processor. Such as for the preprocessor _do.pli_ example if you check that one out.

The calculation is based on the information in the ast. For synthesized elements, there is currently no position data, but the positions for included elements seem off. I'm going to have a closer look at that and we can discuss later.

packages/language/test/lsp/workspace-symbol-request.test.ts
```diff
@@ -13,9 +13,8 @@ import { describe, test, expect } from "vitest";
 import { replaceNamedIndices } from "../utils";
 import { TextDocument } from "vscode-languageserver-textdocument";
 import { URI } from "../../src/utils/uri";
-import { workspaceSymbolRequest } from "../../src/language-server/workspace-symbol-request";
 import { TextDocuments } from "../../src/language-server/text-documents";
-import { createCompilationUnit } from "../../src/workspace/compilation-unit";
+import { CompletionUnitHandler } from "../../src/workspace/compilation-unit";
```
montymxb, line 17: Just a side note. I keep seeing this one, I've already corrected the naming in the plugin PR, it'll apply once that's been added

montymxb (approved):
Alright I think we can pull this into dev now.

The outstanding aspect is related to the ranges associated with outlines, but I believe we should follow up on that separately. @msujew thoughts on that?

Assuming we pull this in, @ssmifi can you make an issue noting the current limitation on the doc/workspace symbol ranges in the outlines?

msujew:
@montymxb I've discussed a few things offline with @ssmifi. We will probably refactor some parts of the PR, in order to better align to the LSP types (and the lack of cross-file support in the document symbol request).

ssmifi:
Refactored to process only the tokens of the requested document. @montymxb please have another look when you have a moment.

- Removed all symbols without a valid range, effectively filtering out preprocessor-generated symbols.
- Removed document symbols from the cache, since it’s cleared on every document change.
  - Workspace symbols remain cached, as the request is triggered with each query update.
  - The infrastructure for document symbol caching is still in place, so re-adding support would be straightforward if needed.
    - This could be useful when switching frequently between large documents without making changes. Please let me know what you think! (cc @msujew)
- Split the document symbol request into two files for better readability.

packages/language/test/lsp/workspace-symbol-request.test.ts
ssmifi: Saw it already in your PR. 👍 
I didn't want to interfere. :)

packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -0,0 +1,326 @@
...
+    elementTokens: IToken[],
+    textDocument: TextDocument,
+    childSymbols: DocumentSymbol[],
+  ): DocumentSymbol[];
+}
+
+class ProcedureSymbolBuilder implements SymbolBuilder {
+  canHandle(token: IToken): boolean {
+    return (
+      token.tokenTypeIdx == CstNodeKind.EntryStatement_Variable &&
```
msujew, line 40: Suggestion: The `tokenTypeIdx` is a lexer internal variable (i.e. it is dynamically assigned while initializing the lexer) and is completely unrelated to the `CstNodeKind` enum. If you want to search for the respective enum value in the token, try `token.payload.kind` instead.

packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -0,0 +1,326 @@
...
+      tokenToRange(elementTokens[elementTokens.length - 1]),
+    ).end;
+    return [symbol];
+  }
+}
+
+class DeclareSymbolBuilder implements SymbolBuilder {
+  canHandle(token: IToken): boolean {
+    if (
+      token.tokenTypeIdx !== CstNodeKind.DefineOrdinalStatement_Unsigned0 ||
```
msujew, line 79: Suggestion: Same issue here, see above.

packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -0,0 +1,326 @@
...
+    const levelSymbols: LevelSymbol[] = [];
+
+    let kind: SymbolKind = SymbolKind.Variable;
+    if (
+      // Type is lower case in the ast but upper case in the definition.
+      item.attributes.some(
+        (a) =>
+          a.kind === SyntaxKind.ComputationDataAttribute &&
+          a.type?.toString() === "constant",
+      )
```
msujew, line 133: Suggestion: We perform case-insensitive parsing, so this comparison also needs to be done case-insensitive.

packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -0,0 +1,326 @@
...
+  getRootSymbols(): DocumentSymbol[] {
+    return this.rootSymbols;
+  }
+}
+
+function isValidToken(token: IToken): boolean {
+  // Will fail for tokens with no end offset.
+  // E.g., when generated by the preprocessor.
+  return token.endOffset !== undefined && token.endOffset > 0;
+}
```
msujew, line 226: Suggestion: We already have a `isValidToken` function in `tokens.ts` which does this check.

packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -0,0 +1,326 @@
...
+
+class LabelSymbolBuilder implements SymbolBuilder {
+  canHandle(token: IToken): boolean {
+    if (
+      token.tokenTypeIdx !== CstNodeKind.Reserves_Variables1 ||
```
msujew, line 231: Suggestion: Same issue here, see above.

packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -0,0 +1,326 @@
...
+  name: string,
+  kind: SymbolKind,
+  selectionRange: Range,
+  children: DocumentSymbol[],
+  textDocument: TextDocument,
+): DocumentSymbol {
+  const selectionLSPRange = rangeToLSP(textDocument, selectionRange);
+  const range = {
+    start: selectionLSPRange.start,
+    end: getLastPositionOrDefault(children, selectionLSPRange.end),
```
msujew, line 286: Suggestion: We should make use of the `END` statements (and assign a token to them) to return more accurate results where appropriate (for procedures for example). Though I'm fine with deferring this to another PR.

packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -0,0 +1,326 @@
...
+    childSymbols: DocumentSymbol[],
+  ): DocumentSymbol[] {
+    const labelPrefixStatement = token.payload?.element.container as Statement;
+    const procedureName =
+      labelPrefixStatement.labels.map((label) => label.name).join(" ") ??
+      "<unknown>";
+    const range = getSyntaxNodeRange(labelPrefixStatement.labels[0]) ?? {
+      start: 0,
+      end: 0,
+    };
```
msujew, line 58: Suggestion: We shouldn't fall back to `0:0` as the range for the syntax node. Can we use the `token.offset` position instead?

packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -0,0 +1,326 @@
...
+    const levelSymbols: LevelSymbol[] = [];
+    const hierarchyBuilder = new LevelHierarchyBuilder();
+
+    for (const item of token.payload?.element.items.filter(
+      (i: DeclaredItem) => i.kind === SyntaxKind.DeclaredItem,
+    )) {
+      levelSymbols.push(
+        ...this.retrieveLevelSymbols(item, null, childSymbols, textDocument),
+      );
+    }
```
msujew, line 107: Suggestion: We should try to keep the amount of `any` to a minimum. Try to cast the `token.payload.element` object first, so we get better IDE support.

packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -0,0 +1,326 @@
...
+  ): DocumentSymbol[] {
+    const labelPrefixStatement = token.payload?.element.container as Statement;
+    if (!labelPrefixStatement?.labels?.length) {
+      return [];
+    }
+
+    const documentSymbols: DocumentSymbol[] = [];
+    for (const label of labelPrefixStatement.labels) {
+      const name = label.name ?? "<unknown>";
+      const range = getSyntaxNodeRange(label) ?? { start: 0, end: 0 };
```
msujew, line 259: Suggestion: Use `token.offset` here as well.

packages/language/src/workspace/compilation-unit.ts
```diff
@@ -135,6 +150,34 @@ export class CompletionUnitHandler {
...
+          "workspaceSymbols",
+          workspaceSymbolRequestForCompilationUnit(this, unit),
+        );
+      }
+      const symbols = unit.requestCaches.get("workspaceSymbols") ?? [];
+      return symbols.filter((symbol) =>
+        symbol.name.toLowerCase().includes(query.toLowerCase()),
+      );
+    });
+  }
```
msujew, line 179: Suggestion: The `CompletionUnitHandler` was designed as a storage for `CompilationUnit` in addition to containing the lifecycle/update logic. I wouldn't try to also add LSP logic in there. Can we refactor this out of the handler?

msujew (commented):
Works pretty well. However, there are a few issues wrt. maintainability in the code.

packages/language/src/workspace/compilation-unit.ts
ssmifi: The parts regarding caching were refactored into the CU after the last review suggestions. For me, both is fine. We should just align w.r.t. this question.

packages/language/src/language-server/document-symbol-builder.ts
ssmifi: Actually, the procedure builder uses the end statement now. This part is still included for the structs.

packages/language/src/workspace/compilation-unit.ts
ssmifi: As discussed, moved to the LS.

packages/language/src/language-server/document-symbol-builder.ts
msujew: Ah, I see. Yeah, that makes sense 👍

packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -0,0 +1,329 @@
...
+
+      documentSymbols.push(
+        createDocumentSymbol(
+          name,
+          SymbolKind.Key,
+          range,
+          childSymbols,
+          textDocument,
+        ),
+      );
```
msujew, line 272: Suggestion: Thinking about this again, will `getSyntaxNodeRange` ever return `undefined` for a valid token? And if the token is invalid, shouldn't we prevent the creation of the document symbol? IMO:
1. `getSyntaxNodeRange` should also check for `isValidToken` and return `undefined` if the token is not valid
2. If `getSyntaxNodeRange` returns `undefined`, skip the respective element, since we cannot navigate to it anyway.

packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -0,0 +1,329 @@
...
+    textDocument: TextDocument,
+    childSymbols: DocumentSymbol[],
+  ): DocumentSymbol[];
+}
+
+class ProcedureSymbolBuilder implements SymbolBuilder {
+  canHandle(token: IToken): boolean {
+    return (
+      token.payload.kind == CstNodeKind.ProcedureStatement_PROCEDURE &&
+      token.payload?.element.kind === SyntaxKind.ProcedureStatement
```
msujew, line 47: Minor: Either test for `undefined` on none of the payloads, or only the first one. The second test cannot be undefined, if the first is already defined.

msujew (commented):
Looks good, just one more thing wrt. invalid token ranges. See below.

msujew (approved):
Thank you, works great!

ssmifi:
Thanks again for the thorough review!

Issue #133, opened 2025-05-15
Find references on qualified name parts does not work


msujew:
1. Declare a nested object like `DCL 1 A, 2 B;`
2. Reference it somewhere via `A.B`.
3. Use `Find References` on `A`.
4. No references will be found.

PR #135, opened 2025-05-15
Fix reference lookup for qualified names

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/133

We didn't add the inverse references at the correct spot. This led us to miss the inverse references for parts of fully qualified variable declarations. This change ensures that all references end up in the correct spot.

This change also adds tests to the reference search feature.

### DIFF
--- a/packages/language/test/utils.ts
+++ b/packages/language/test/utils.ts
@@ -25,6 +25,7 @@ import { forEachNode } from "../src/syntax-tree/ast-iterator";
 import { IntermediateBinaryExpression } from "../src/parser/abstract-parser";
 import { IToken } from "@chevrotain/types";
 import { escapeRegExp } from "../src/parser/tokens";
+import { referencesRequest } from "../src/language-server/references-request";
 import { completionRequest } from "../src/language-server/completion/completion-request";
 
 interface AssertNoDiagnosticsOptions {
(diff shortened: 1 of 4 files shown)


packages/language/test/utils.ts
```diff
@@ -453,6 +454,48 @@ export function expectLinks(text: string) {
...
+      rangeIndex: ranges[index],
+    })),
+  );
+
+  const unit = parseAndLink(output);
+
+  for (const { label, offset, rangeIndex } of requests) {
+    const result = referencesRequest(unit, unit.uri, offset);
+
+    expectedFunction(
```
montymxb, line 473: Unrelated, we should rename this to something like `expectedStrictEquals`.

montymxb (approved):
:octocat: Approved. Looks good, can verify associated issue has been corrected as well.

packages/language/test/utils.ts
```diff
@@ -453,6 +454,48 @@ export function expectLinks(text: string) {
   }
 }
 
+export function expectReferences(text: string) {
```
didrikmunther, line 457: There seems to be new "meta"-syntax introduced in the tests, e.g.: `<|1:<|1>A|>`?

We should make a doc comment here similarly to `expectLinks` to document this syntax.

Issue #26, opened 2025-01-22
[feature-lsp] Support "Completion" request


msujew:

msujew:
Reopened due to https://github.com/zowe/zowe-pli-language-support/pull/67.

PR #137, opened 2025-05-15
Initial code completion feature

msujew:
Related to https://github.com/zowe/zowe-pli-language-support/issues/26.

Adds initial support for the completion feature of PL/I. Currently only allows for completion of local and fully qualified references within expressions.

Eventually this infrastructure should support more completion for keywords of the language and snippets.

### DIFF
--- a/packages/language/src/language-server/completion/completion-generator.ts
+++ b/packages/language/src/language-server/completion/completion-generator.ts
@@ -1,0 +1,96 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { CompletionItemKind } from "vscode-languageserver-types";
+import { Statement, SyntaxKind, SyntaxNode } from "../../syntax-tree/ast";
+import { CompilationUnit } from "../../workspace/compilation-unit";
+import { CompletionItem } from "../types";
+import { FollowElement, FollowKind } from "./follow-elements";
+import { getQualifiedName } from "../../linking/resolver";
+
+export interface SimpleCompletionItem extends Omit<CompletionItem, "edit"> {
+  text: string;
+}
+
+export function generateCompletionItems(
+  unit: CompilationUnit,
+  context: SyntaxNode | undefined,
+  followElement: FollowElement,
+): SimpleCompletionItem[] {
+  const items: SimpleCompletionItem[] = [];
+  if (followElement.kind === FollowKind.CstNode) {
+    // TODO: implement completion for keywords
+    return [];
+  }
+  if (followElement.kind === FollowKind.QualifiedReference) {
+    context = followElement.previous;
+  } else if (!context) {
+    return [];
+  }
+  const scopeCache = isPreprocessorNode(context, unit)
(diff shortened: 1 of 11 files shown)


packages/language/src/language-server/completion/completion-generator.ts
```diff
@@ -0,0 +1,96 @@
...
+        return CompletionItemKind.Function;
+      }
+      return CompletionItemKind.Variable;
+    }
+  }
+  return CompletionItemKind.Variable;
+}
+
+function isPreprocessorNode(node: SyntaxNode, unit: CompilationUnit): boolean {
+  let parent = node;
```
montymxb, line 91: I would have thought we would have a faster way to resolve the node type than this, like an additional prop or link to the root. Our container chains won't be so deep here, but if they are we may want to revisit this later.

packages/language/src/utils/search.ts
```diff
@@ -4,23 +4,57 @@ export function binaryTokenSearch(
...
+  const end = token.endOffset!;
+  // If the offset is right after the end of a word token, return that token
+  return offset - end === 1 && /\w$/u.test(token.image);
+}
+
+function isBeforeTokenEnd(token: IToken, offset: number): boolean {
+  return token.endOffset! >= offset || isAtTokenEnd(token, offset);
+}
+
+export function completionTokenSearch(
```
montymxb, line 44: Can call this `completionTokenIndexSearch`, matching the naming done above for the modified binary token search.

montymxb (commented):
Couple questions here and there, a clarification, and a suggestion. Rest looks pretty good for starting completion support.

packages/language/src/parser/parser.ts
```diff
@@ -1570,12 +1577,15 @@ export class PliParser extends AbstractParser {
...
-      });
+    this.MANY1({
+      DEF: () => {
+        this.SUBRULE_ASSIGN1(this.LabelPrefix, {
+          assign: (result) => {
+            element.labels.push(result);
+          },
+        });
+      },
+      GATE: () => this.canParseLabelPrefix(),
```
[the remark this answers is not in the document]
msujew: I actually forgot about this - we can drop it again. I thought this might fix some weird behavior in the error recovery, but it doesn't.

packages/language/src/language-server/completion/completion-request.ts
```diff
@@ -0,0 +1,98 @@
...
+    generateCompletionItems(unit, context, element),
+  );
+  return convertSimpleToItem(items, offset, cursorToken);
+}
+
+function getTokenContext(
+  tokens: IToken[],
+  index: number,
+): SyntaxNode | undefined {
+  for (let i = 0; i < 5; i++) {
```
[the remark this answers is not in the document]
msujew: Not in particular, I just wanted to make sure we can find a bit of context, but not search too long, and a lookbehind of 5 tokens should always be enough. I'll add a comment explaining this behavior.

packages/language/src/language-server/completion/completion-generator.ts
msujew: Right now we don't. I don't want to add a prop to the preprocessor types, since they are just "normal" AST nodes - but sourced from a different parser. I also think it should be generally fine, since it will always be fast enough and we perform this check only once per completion request.

montymxb: That was my thought as well, if it were in a hot path sure, but it's just one-of completions. Should be fine.

montymxb (approved):
:octocat: Good to go

PR #138, opened 2025-05-15
Flag Duplicate & Mutually Exclusive Compiler Opts

### LINKED ISSUE #105: Compiler Options - Can define Repetitive Options
If the compiler option is given with "OR" option, e.g. `EXIT | NOEXIT`, you can only choose one, not both, or repetitively
e.g.  here you can define any combination.
```

*PROCESS NOEXIT, EXIT('HATCHI'), EXIT, EXPORTALL, NOEXPORTALL; 

 MYPROC: PROCEDURE OPTIONS (MAIN);
    DCL TRUE BIT(1) INIT(1);
    DCL FALSE BIT(1) INIT(0);
    DCL OR_VALUE; 
    OR_VALUE = TRUE | FALSE;
    DCL NOT_VALUE;  

 END MYPROC;
```

montymxb:
Closes #105 by adding:
- warning diagnostic for duplicate options `(AG, AG)`
- warning diagnostic for mutually exclusive options `(NOEXIT, EXIT)`
- warning for dupes in short & long form `(AG, AGGREGATE)`
- warning for mutually exclusive options in long & short form `(NAG, AGGREGATE)`
- updated the existing diagnostic for unrecognized options to conform to PLI's compiler code for that
- updated `assertNoDiagnostics` to factor in compiler option diagnostics as well

In all cases the compiler only emits a warning rather than an error. And in cases of duplicate / mutex options the compiler always appears to take the last option seen (verified in listings output).

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/translator.ts
+++ b/packages/language/src/preprocessor/compiler-options/translator.ts
@@ -1147,6 +1147,7 @@ export function translateCompilerOptions(
   input: AbstractCompilerOptions,
 ): CompilerOptionResult {
   translator.options = {};
+  translator.optionSet.clear();
   translator.issues = [...input.issues];
   for (const option of input.options) {
     translator.translate(option);
--- a/packages/language/src/preprocessor/compiler-options/translator.ts
+++ b/packages/language/src/preprocessor/compiler-options/translator.ts
@@ -35,10 +35,23 @@ interface TranslatorRule {
 
 type Translate = (option: CompilerOption, options: CompilerOptions) => void;
 
+/**
+ * Tracks how a rule is applied, either positively or negatively
+ */
+enum Applied {
+  POSITIVE = 1,
+  NEGATIVE = -1,
+}
+
 class Translator {
   options: CompilerOptions = {};
-  optionSet: Set<string> = new Set();
   issues: CompilerOptionIssue[] = [];
+
+  /**
+   * Translator rules that have been applied to the current options,
+   * along w/ info on whether they were applied positively or negatively
+   */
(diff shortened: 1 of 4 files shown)


packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -103,9 +108,36 @@ class Translator {
...
+      });
+    }
+  }
+
+  /**
+   * Reports duplicate or mutually exclusive compiler options.
+   */
+  checkDuplicateOption(option: CompilerOption, name: string): void {
+    const modifiedName = name.startsWith("NO") ? name.slice(2) : `NO${name}`;
+    if (this.optionSet.has(name)) {
```
ssmifi, line 122: _Question:_ Works for the cases you've listed in the PR. However, I was wondering, if we also want to support long vs short form and short form negation? E.g., AGGREGATE, AG and AG, NAG

ssmifi (commented):
LGTM. One question if we should support the short forms of compiler options and their negations.

packages/language/src/preprocessor/compiler-options/translator.ts
montymxb: Good point. I haven't double checked those short hand aliases yet, or there non-standard forms. I take care of that in a second pass, and get some tests in for those edge cases as well.

montymxb:
Refactored to rely on the pre-existing TranslatorRules for tracking which options have been applied. Updated tests to check for short & long hand forms in turn, and updated the `assertNoDiagnostics` function to fail on compiler option diagnostics as well.

msujew (approved):
Great, looks good to me 👍

PR #139, opened 2025-05-19
Prevent infinite loops in preprocessor code

msujew:
Right now, writing a statement such as `%DO %FOREVER; %END;` will keep the language server in an infinite loop (as the preprocessor instruction itself also is an infinite loop).

To prevent such infinite loop, this change limits the amount of times we can jump to an individual instruction to 500 (subject to change in the future). This prevents the language server from locking up while writing code.


packages/language/src/preprocessor/pli-preprocessor-interpreter-state.ts
```diff
@@ -54,11 +58,18 @@ export class PliPreprocessorInterpreterState
     const newCounter = next(this.plainState.programCounter);
     if (newCounter >= 0 && newCounter < this.plainState.program.length) {
       this.plainState = { ...this.plainState, programCounter: newCounter };
+      const instruction = this.goToCounter.get(newCounter) ?? 0;
```
ssmifi, line 61: _Question_: Do we have any data on what would be a reasonable value for real-world programs? As you said, the value is subject to change. A limit of 500 seems low, especially considering nested loops. That said, any number we pick now might be speculative. 
As a _suggestion_ for a future PR: should we consider making this limit user-configurable? If that sounds reasonable, would it make sense to open an issue to track it?
_Tiny remark_ by the way: I’d probably name the variable `instructionCount` or `Calls` for clarity.

packages/language/src/preprocessor/pli-preprocessor-interpreter-state.ts
```diff
@@ -54,11 +58,18 @@ export class PliPreprocessorInterpreterState
...
+      this.goToCounter.set(newCounter, instruction + 1);
+      // Ensure that we don't run into an infinite loop
+      if (instruction > MAX_GO_TO_COUNT) {
+        this._halt = true;
+        console.log("Long running preprocessor code detected. Stopping.");
```
ssmifi, line 66: _Question_: Would it make sense to surface this as a diagnostic instead of logging to the console? That way, users would get direct feedback if they accidentally introduce an infinite loop.

packages/language/test/parser/pli-lexer.test.ts
```diff
@@ -476,6 +476,18 @@ describe("PL/1 Lexer", () => {
     ).toStrictEqual([]);
   });
 
+  test("DO FOREVER without LEAVE", () => {
```
ssmifi, line 479: _Question_: This test currently fails by running indefinitely, which might be confusing if you are unfamiliar with the behavior. Would it be feasible to make it fail more gracefully?

ssmifi (commented):
Looks good, just a couple of questions to make sure we're all on the same page.

packages/language/test/parser/pli-lexer.test.ts
msujew: I'm confused, the test doesn't fail?

packages/language/src/preprocessor/pli-preprocessor-interpreter-state.ts
msujew: > Do we have any data on what would be a reasonable value for real-world programs?

Unfortunately, no. I'll turn it up a bit, but making this configurable eventually is probably preferable.

packages/language/test/parser/pli-lexer.test.ts
ssmifi: Sorry for the confusion. I just meant that if a bug is introduced and the test _should_ fail, it would run indefinitely rather than failing with a clear error message.

ssmifi (approved):
Looks good. Thanks for the answers and the note in https://github.com/zowe/zowe-pli-language-support/issues/142.

Issue #103, opened 2025-04-14
Compiler Options - CASERULES missing validation


asatklichov:
Compiler option: CASERULES KEYWORD has only four options:  LOWER | MIXED | START |  UPPER 
but you can define CASERULES wrongly with any options, as like below. 

`*PROCESS CASERULES(KEYWORDsszz(START));  /*NO KEYWORD VALIDATION*/ `
`*PROCESS CASERULES(KEYWORD(dddSTART));  /*NO KEYWORD options VALIDATION*/ `

montymxb:
Closed by #143

PR #143, opened 2025-05-21
CASERULES validation

ssmifi:
Closes https://github.com/zowe/zowe-pli-language-support/issues/103

Straigtforward. However, I swapped the logic of `ensureType`, because the parsed type was reported as expected. Please double-check.


msujew (approved):
Looks good to me 👍

PR #145, opened 2025-05-21
Refactor linking system

didrikmunther:
The linking system was getting hard to reason about, so this PR refactors it.

This PR:

- Adds the Langium `MultiMap`
- Moves scope, qualified syntax nodes, and declared item parsers, from `symbol-table.ts`, into their own files
- Changes flow of deep linking code (`iterateSymbolTable`), but keeps the same behavior.

### DIFF
--- a/packages/language/src/linking/qualified-syntax-node.ts
+++ b/packages/language/src/linking/qualified-syntax-node.ts
@@ -27,30 +27,6 @@ export class QualifiedSyntaxNode {
   level: number;
   nameToken: IToken;
 
-  /**
-   * The symbol is redeclared in the same scope.
-   * For example:
-   *
-   * ```pli
-   * // Case 1
-   * DCL A;
-   * DCL A;
-   *
-   * // Case 2
-   * DCL A, A;
-   *
-   * // Case 3
-   * A: PROCEDURE;
-   * A: PROCEDURE;
-   *
-   * // Case 4
-   * DCL 1 A,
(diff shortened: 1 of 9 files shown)


msujew:
Quick question before looking closer into this: Does this modify the data structures used for the scoping? I rely on their shape right now for the new completion feature in https://github.com/zowe/zowe-pli-language-support/pull/137.

didrikmunther:
> Quick question before looking closer into this: Does this modify the data structures used for the scoping? I rely on their shape right now for the new completion feature in #137.

From what I see, it doesn't. I just did a quick rebase of your branch on top of this, and there was no issue with the structure!

packages/language/src/linking/qualified-syntax-node.ts
```diff
@@ -0,0 +1,114 @@
...
+   * A: PROCEDURE;
+   * A: PROCEDURE;
+   *
+   * // Case 4
+   * DCL 1 A,
+   *       2 B,
+   *       2 B;
+   * ```
+   */
+  // isRedeclared: boolean;
```
msujew, line 52: Question: Do we still need this?

msujew (approved):
Looks good to me 👍

packages/language/src/linking/qualified-syntax-node.ts
didrikmunther: No, it's dead! Fixed!

PR #154, opened 2025-05-22
Support factorized variable declarations

msujew:
Adds support for declarations such as `DCL (I) CHAR(8)` or even more complicated factorized structure variables. We enable this by pushing declarations inside of factorized variable declarations back onto the stack for immediate processing.

### DIFF
--- a/packages/language/src/linking/declared-item-parser.ts
+++ b/packages/language/src/linking/declared-item-parser.ts
@@ -10,7 +10,7 @@
  */
 
 import { Severity, tokenToRange, tokenToUri } from "../language-server/types";
-import { DeclaredItem } from "../syntax-tree/ast";
+import { DeclaredItem, SyntaxKind } from "../syntax-tree/ast";
 import { PliValidationAcceptor } from "../validation/validator";
 import * as PLICodes from "../validation/messages/pli-codes";
 import { SymbolTable } from "./symbol-table";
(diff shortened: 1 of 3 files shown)


packages/language/src/linking/declared-item-parser.ts
```diff
@@ -118,31 +131,40 @@ export class DeclaredItemParser {
...
+              uri: tokenToUri(nameToken) ?? "",
+            });
+          }
+
+          // Otherwise, we can add the node to the symbol table.
+          const node = new QualifiedSyntaxNode(nameToken, child, parent, level);
+
+          nodes.add(name, node);
+          table.addSymbolDeclaration(name, node);
+          this._generate(table, node, level);
```
didrikmunther, line 163: Just noting this for an extension to this logic.

Trying this code on the mainframe:

```
DCL 1 A, 2 (B, 3 C);
```

Generates the errors:

```
IBM1376I E: Factoring of level numbers into declaration lists containing level numbers is invalid. The level numbers in the declaration list will be ignored.

IBM1483I E: RULES(NOLAXDCL) violation: the structure member B is declared without any data attributes. A level number may be incorrect.

IBM1483I E: RULES(NOLAXDCL) violation: the structure member C is declared without any data attributes. A level number may be incorrect.
```

The first most error should be generated here.

packages/language/test/linking.test.ts
```diff
@@ -196,6 +196,35 @@ describe("Linking tests", () => {
  PUT (<|b1>B.<|c>C);
  PUT (<|b2>B.<|d>D);`));
 
+    test("Factorized names can be used as variables", () =>
```
didrikmunther, line 199: This test case should be added, which links on the mainframe but not here:

```
 DCL 1 A, (2 B);
 PUT(A.B);
```

didrikmunther, line 199: This test case should be added, which produces an ambiguity error (`IBM1881I S: The reference C is ambiguous.`) on the mainframe, but not here:

```
 DCL 1 A, (2 B, 3 C CHAR(10), 2 C CHAR(10));
 PUT(C);
```

didrikmunther (commented):
Nice! I've provided some extra test cases which shows deviating behavior from the mainframe.

didrikmunther (approved):

msujew:
FYI for future me: There are still quite a few edge cases that aren't well supported in this PR. We will fix those in a separate PR.

PR #162, opened 2025-05-23
Fix parenthesized factorized declaration

didrikmunther:
Reworks the factorized declaration logic by introducing a preprocessing step for declared items which unrolls factorized declarations before parsing them in the `DeclaredItemParser`.

We now properly link these declarations:

```
 DCL 1 A, (2 B);
 PUT(A.B);
```


msujew (approved):
Thanks, that approach makes much more sense 👍

PR #169, opened 2025-05-30
Add test harness framework

didrikmunther:
# Harness Test Framework

Inspired by the [TypeScript Fourslash testing system](https://github.com/microsoft/TypeScript/tree/main/tests/cases/fourslash).

Note: I'm very open to suggestions on the testing DSL syntax.

The point of this is to consolidate testing, making it more readable and maintainable, as well as supporting a new business requirement: verifying that tests configurations are correct against the mainframe. The plan for this is to provide different "harness implementations" (https://github.com/zowe/zowe-pli-language-support/pull/169/files#diff-8f1fc5ed58a7a9f307beb10fa61538a8ab015ed6c6db29b06db6932aaf600a5aR23), one for the extension and one for the mainframe. This way, tests will fail on the remote if the expected behavior is not actually the case on the mainframe. Since the PL/I compiler is much less granular than our language server, we need to experiment with how to make the declarations granular enough to test our language server, but also be translated to the mainframe's courser granularity.

Problem: vitest does not seem to have any kind of way to overwrite watching/running behavior, so upon saving one of these test files, vitest does not rerun. Adding the directory to the `included` array in the vitest configuration causes vitest to try to run the actual test file, which does not work as it needs to run in a sandbox in the `harness-runner.ts`. What would be the best solution is to be able to inject/hook into/overwrite vitest watching/running behavior, so that we can specify that these kinds of files need to be run in a VM. Alas, vitest does not support this and does not seem to have a discussion around this feature either.

## Interesting facts

- You can debug files run by `vm`, iff you add the `filename` field, and the code content is identical to the actual file.

## Example

```ts

/**
 * Redeclaration of label must fail
 */

// @wrap: main
//// OUTER: PROCEDURE;
//// END OUTER;
//// <|1:OUTER|>: PROCEDURE;
//// END OUTER;
//// CALL OUTER;

verify.expectExclusiveErrorCodesAt(1, code.Severe.IBM1916I.fullCode);

```

## Technical details

- Uses the `vm` package in node.
- Has a light parser for the harness files, supporting tags and comments.
- Is very similar to the fourslash syntax.

## Usage

All test files must include the framework reference.

```ts
/// <reference path="framework.ts" />
```

### Virtual File System

A single test file can describe multiple PL/I files.
The test runner will create a virtual file system.

A file is defined by the `// @filename: ...` tag on a file definition, e.g.:

```ts
// @filename: file1.pli
//// DCL A;
//// PUT(A);

// @filename: file2.pli
//// DCL B;
//// PUT(B);
```

### Wrap

The `// @wrap: ...` tag wraps a file in a predefined wrap template.
The wrap template are defined in the `wrappers/` directory.
The `<...>` tag is used to declare where content is inserted to the wrap template.

### Imperative Testing Commands

After the file definitions, the commands follow.

## Contributing

When adding a test command, it first needs to be declared in the `tests/framework.ts` file to be exposed to the test files.
Following that, the semantics needs to be implemented in the `harness-implementation.ts` file.

## Testing language suggestion

CleanShot 2025-06-02 at 13 15 09@2x
CleanShot 2025-06-02 at 13 12 58@2x
CleanShot 2025-06-02 at 10 09 55
CleanShot 2025-06-02 at 10 07 37
CleanShot 2025-06-02 at 10 07 14
CleanShot 2025-06-02 at 14 17 38@2x
CleanShot 2025-06-02 at 14 17 55@2x

We could add syntax highlighting to template string, similar to the Apollo GraphQL extension:

CleanShot 2025-06-02 at 09 39 10 (1)


didrikmunther:
That makes more sense, thanks!

PR #182, opened 2025-06-04
Add builtins

didrikmunther:


msujew (commented):
Very nice! Adding this sort of `DocumentConsolidator` works very well. This also allows us to improve on a minor issue I had with the new closed-document feature:

Currently the `RealDocuments` also store all documents that are read from disk for the purpose of providing LSP results. While this works well, I want to separate those from the documents that are open in the editor. This will become more useful later, once we start listening to file changes (compared to only listening to editor changes). Can we introduce a third `FileDocuments` instance for this, which acts as a storage for on-disk documents that we need for the purpose of answering LSP requests? These then act similar to the `NormalizedTextDocuments` with the `loadFromURI` option enabled.

msujew (approved):
Thank you, looks good to me 👍

PR #184, opened 2025-06-05
Consolidate preprocessor and fix some parser issues

msujew:
Fixes a bunch of issues in the preprocessor code:

1. Consolidates the tokens used by the preprocessor parser to be the same as the tokens used by the normal code parser.
2. Fixes an issue with `%GO TO label;` statements. Now `%GOTO label;` is also valid.
3. Fixes an issue related to parsing expressions within `%IF`, `%WHILE` and assignment statements. Inside of the expressions, it is not necessary to use `%` (i.e. `%IF %X > 0` is wrong, `%IF X > 0` is correct).
4. Removes some of the preprocessor directives from the normal parser - the normal parser was unable to parse them anyway, since the tokens where never generated by the lexer.

This is only one part of the preprocessor changes - but I wanted to split them to make it easier to review them. The second part of the preprocessor improvements will allow to integrate preprocessor statements into "normal" statements like:

```
PCK: PACKAGE EXPORTS(
%IF EXPORT_ALL %THEN %DO
  A,
  B,
  C
%END;
  )
```

This is currently not possible and leads to unexpected parser errors, as the preprocessor can currently only correctly parse statements if they are not inside of normal code statements.


packages/language/src/preprocessor/pli-preprocessor-lexer.ts
```diff
@@ -42,7 +43,16 @@ export class PliPreprocessorLexer {
...
+        !token.PATTERN.sticky
+      ) {
+        // Sticky patterns are required for the lexer to work correctly
+        // Usually we shouldn't need to warn, but just in case
+        console.warn(
+          `Token ${token.name} has a non-sticky pattern. This may lead to unexpected behavior.`,
+        );
         token.PATTERN = new RegExp(
           token.PATTERN.source,
           token.PATTERN.flags + "y",
```
ssmifi, line 58: _Question_: Aren't all RegExp tokens sticky now? Should we move this check into a dedicated unit test and remove it at runtime?

ssmifi, line 58: _Question_: By the way, there are still some tokens left in `packages/language/src/preprocessor/compiler-options/parser.ts` that are not sticky. Should they be sticky as well?

ssmifi (approved):
Looks good. Just a couple of questions/suggestions about the capture groups and sticky.

> I wanted to split them to make it easier to review them.

Appreciated! :)

packages/language/src/parser/tokens.ts
```diff
@@ -165,1416 +165,1446 @@ function tokenizeWithCompilerOption(
...
   name: "RECURSIVE",
-  pattern: /RECURSIVE/i,
+  pattern: /RECURSIVE/iy,
   categories: [ID, SimpleOptions],
   longer_alt: ID,
 });
 export const PROCEDURE = createToken({
   name: "PROCEDURE",
-  pattern: /(X)?PROC(EDURE)?/i,
+  pattern: /(X)?PROC(EDURE)?/iy,
```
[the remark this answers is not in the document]
msujew: Internally, we simply use `string.test`, which doesn't really care about capture groups and just returns a `boolean` result. So I don't really think it really matters?

packages/language/src/preprocessor/pli-preprocessor-lexer.ts
msujew: Good question. Though, these are unrelated to the actual preprocessor parsing, and will be made sticky automatically by Chevrotain.

PR #193, opened 2025-06-11
Fixing more preprocessor and normal parser issues

msujew:
As mentioned in #184, this change enables parsing of embedded preprocessor code (see new preprocessor tests).

It also fixes ordinal parsing, since I've noticed that this is also broken.

### DIFF
--- a/packages/language/src/preprocessor/pli-preprocessor-parser-state.ts
+++ b/packages/language/src/preprocessor/pli-preprocessor-parser-state.ts
@@ -181,6 +181,7 @@ export class PliPreprocessorParserState implements PreprocessorParserState {
       element,
     };
     this.index++;
+    this.skipHiddenTokens();
     return true;
   }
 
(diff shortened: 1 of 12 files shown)


packages/language/src/preprocessor/pli-preprocessor-parser-state.ts
```diff
@@ -204,9 +205,16 @@ export class PliPreprocessorParserState implements PreprocessorParserState {
...
+    this.skipHiddenTokens();
     return token;
   }
 
+  private skipHiddenTokens() {
+    while (this.tokens[this.index]?.tokenType.GROUP) {
+      this.index++;
+    }
+  }
+
```
ssmifi, line 217: _Question_:

There is already a remark in `pli-preprocessor-lexer.ts` in `tokenizePliTokensUntilSemicolon()`, which is called by the parser-state:
> // TODO: Ignore hidden tokens already here?

Should we address this TODO by clarifying why hidden token handling is still needed separately in parser-state?
It appears that most of the hidden token handling is already performed in `pli-preprocessor-lexer`.

msujew:
Superseded by https://github.com/zowe/zowe-pli-language-support/pull/221.

PR #195, opened 2025-06-12
Support Config Compiler Options

### LINKED ISSUE #110: [Feature] Add language server configuration
Similar to the HLASM and Cobol extension, we want to add a configuration format that allows to specify which files belong to one compilation unit, what the compiler options are, etc.

montymxb:
Finally closes #110 by adding in the remaining compiler option handling from the .pliplugin config folder. Currently compiler options that come from the config are pulled out around the same time we extract them from the program itself, but we run a separate translation step on those options 1st, and then on the program sourced config options 2nd. The results are then merged together once they're done.

This helps to avoid mixing up the sourcing of diagnostics, as the config sourced options have diagnostics reported on the *PROCESS statement when present, and fallback to a `console.log` otherwise. The handling of existing compiler options remains unchanged.

There is an outstanding issue with reporting duplicates across the two different sets of compiler options. Since they're processed independently, we do _not_ get reports of duplicates or mutually exclusive options across the two option sets. This still needs to be amended. But depending on demand we can either rectify it here or in a follow-up.

Lastly this resolves an issue where changes to the .pliplugin folder were only picked up at start. Meaning changes were being ignored from both the program & process group configuration files. This led to some confusing situations where the state was not correct until restarting everything. Note that since the watching is client side, we'll need an FS-free implementation to get the same effect in a browser-context.

### DIFF
--- a/packages/vscode-extension/src/extension/main.ts
+++ b/packages/vscode-extension/src/extension/main.ts
@@ -177,8 +177,14 @@ function watchPlipluginFolder(
   workspaceFolder: string,
   context: vscode.ExtensionContext,
 ): void {
-  const folderPattern = new vscode.RelativePattern(workspaceFolder, ".pliplugin");
-  const filePattern = new vscode.RelativePattern(workspaceFolder, ".pliplugin/*.json");
+  const folderPattern = new vscode.RelativePattern(
+    workspaceFolder,
+    ".pliplugin",
+  );
+  const filePattern = new vscode.RelativePattern(
+    workspaceFolder,
+    ".pliplugin/*.json",
+  );
 
   const folderWatcher = vscode.workspace.createFileSystemWatcher(folderPattern);
   const fileWatcher = vscode.workspace.createFileSystemWatcher(filePattern);
--- a/packages/language/src/workspace/plugin-configuration-provider.ts
+++ b/packages/language/src/workspace/plugin-configuration-provider.ts
@@ -138,9 +138,7 @@ class PluginConfigurationProvider {
 
     // clear otherwise, no valid program config to use
     this.programConfigs.clear();
-    console.warn(
-      "No program config found, clearing existing configurations."
-    );
+    console.warn("No program config found, clearing existing configurations.");
   }
 
   /**
@@ -180,7 +178,7 @@ class PluginConfigurationProvider {
     // clear otherwise, no valid PG to use
     this.processGroupConfigs.clear();
     console.warn(
-      "No process group config found, clearing existing configurations."
+      "No process group config found, clearing existing configurations.",
     );
   }
 


packages/language/src/language-server/connection-handler.ts
```diff
@@ -216,5 +216,9 @@ export function startLanguageServer(connection: Connection): void {
       compilationUnitHandler.getAllCompilationUnits(),
     );
   });
+  connection.onNotification("workspace/didChangePlipluginConfig", () => {
```
msujew, line 219: Suggestion: Can we store these as a constant somewhere?

packages/language/src/preprocessor/compiler-options-processor.ts
```diff
@@ -11,44 +11,154 @@
...
+        issue.range = range
+          ? {
+              start: range.start,
+              end: range.start + CompilerOptionsProcessor.PROCESS_TOKEN_LENGTH,
+            }
+          : { start: 0, end: 0 };
+        issue.message = `PLI Plugin Config: ${issue.message}`;
+      } else {
+        // report as is, don't adjust
+        console.warn(`PLI Plugin Config: ${issue.message}`, issue.range);
```
msujew, line 157: Suggestion: As it is, this will print a warning for every issue on every keystroke - even if the config hasn't changed. This will pollute the output log **very quickly**. We should print those only when the config has actually changed.

packages/language/src/preprocessor/compiler-options-processor.ts
```diff
@@ -11,44 +11,154 @@
...
+        const configAbstractOptions =
+          parseAbstractCompilerOptions(configCompilerOpts);
+        configCompilerResult = translateCompilerOptions(configAbstractOptions);
+        // adjust all issues from config
+        this.adjustIssuesFromConfig(configCompilerResult.issues, range);
+        if (!range) {
+          // don't report config issues if we don't have a *PROCESS directive to attach them to
+          configCompilerResult.issues = [];
+        }
+      }
```
msujew, line 87: Suggestion: I think we should parse the config outside of the compiler options extraction (that happens on every parse of every compilation unit). This makes more sense conceptually (since the compiler options come from outside of the file anyway) and should also save cycles, as we don't need to reparse the compiler options every time.

We can store the parsed and configured compiler options separately on the compilation unit and then merge them for use in the tokenizer, etc. I.e. have three versions of the compiler options available, config, text and merged. This would also resolve the issue from above (printing the warnings).

msujew (commented):
Works well, but I'm a bit concerned about the coupling of the systems at play here: The config should be passed into the lifecycle, not be parsed as part of the lifecycle. See comments below.

packages/vscode-extension/src/extension/main.ts
```diff
@@ -161,3 +167,32 @@ function startLanguageClient(context: vscode.ExtensionContext): LanguageClient {
...
+  const watcher = fs.watch(plipluginPath, (_eventType, filename) => {
+    if (filename === "pgm_conf.json" || filename === "proc_grps.json") {
+      client.sendNotification(WorkspaceDidChangePlipluginConfigNotification);
+    }
+  });
+
+  watcher.on("error", (error) => {
+    console.error("Error watching .pliplugin folder:", error);
+    watcher.close();
+  });
```
msujew, line 195: Suggestion: This breaks pretty easily, for example if the `.pliplugin` folder doesn't exist during startup of the extension but is only added in later.

Please use the file watching options provided by VSCode, i.e. `vscode.workspace.createFileSystemWatcher`, which are much more robust. An alternative might be to instruct the language client to watch the folder, that way we don't actually need to base our implementation on the exact language client. See how [Langium is doing this](https://github.com/eclipse-langium/langium/blob/b9cd5299af415cab6045ad0a79e05526cdc62bce/packages/langium/src/lsp/document-update-handler.ts#L86-L120).

msujew (commented):
Looks pretty good. I just have one comment about the file watching implementation, which might be prone to breakage.

packages/vscode-extension/src/extension/main.ts
montymxb: Good point, just patched both those up (folder case + watcher abstraction), thanks for looking that over again. Also had no idea about the watcher abstraction under the vscode extension api, that's _much_ better in comparison.

Didn't opt to go the full route to use the LS to check client capabilities for dynamic registration, but that's good to know as well. I'm assuming we can set that up later as an improvement.

msujew (approved):
Alright, thank you 👍

PR #197, opened 2025-06-12
Add binary expression and assignment completion support

didrikmunther:


packages/language/src/language-server/completion/follow-elements.ts
```diff
@@ -114,43 +147,47 @@ export function getFollowElements(
...
-        kind: FollowKind.CstNode,
-        types: AllStatementStartKeywordsArray,
-      });
-      break;
+      return [
+        {
+          kind: FollowKind.CstNode,
+          types: AllStatementStartKeywordsArray,
+        },
+      ];
```
msujew, line 155: Minor: I've noticed that we're actually missing the assignment statement follow node here. So a `FollowKind.LocalReference` element needs to be returned in the array as well.

code_samples/DIDRIK3.pli
```diff
@@ -0,0 +1 @@
+ DCL Y FIXED;
```
msujew, line 1: Leftover from testing?

msujew (approved):
Looks good to me 👍

PR #204, opened 2025-06-18
Add missing builtin functions

msujew:
Adds the rest of the missing builtin functions to the list of builtins.

Also adjusts our semantic token highlighting to ensure that functions are highlighted as expected.
Fixes another very minor bug related to the `POP` token (it was missing its `longer_alt`.


didrikmunther (approved):
LGTM!

PR #208, opened 2025-06-23
Fix invalid assertion in document symbols

msujew:
In some cases, it seems like the `element?.container` property is actually `null` or `undefined`. Alternatively, some procedures didn't have a valid label. In these cases, the document symbol provider will throw an error which will be propagated to the user.

This change just adds an additional check to ensure that we actually have a non-null statement obj at hand.


ssmifi (approved):
LGTM. Please run pretty and merge.

Besides, is an element withtout container something that is desired? Should we add a ticket for this and investigate?

msujew:
@ssmifi It's a result of an invalid parse tree. In that sense, it is expected behavior and does IMO not require a change.

PR #210, opened 2025-06-23
Improve symbol table and linking performance

msujew:
Performs three main changes:

1. Caches the builtin scope to prevent re-computation on every change.
2. Due to 1. we need to overlay the builtins scope with a new scope for every compilation unit.
3. Instead of iterating over each scope of **each node** in the symbol table, we now only iterate over each unique scope once. This is the main performance improvement of this PR.

The changes cannot be independently tested. Large files should simply be evaluated much faster by the language server. All tests should still pass, since nothing changed on a functional level.


didrikmunther (approved):
Nice!

PR #214, opened 2025-06-24
Further Include + Config Improvements

montymxb:
Supplements upon our existing include/config work with some general improvements:
- renames the config extension option to `include-extensions` from `copybook-extensions` per request
- adds glob matching for program config `program` entries with [minimatch](https://www.npmjs.com/package/minimatch)
- disables relative & absolute include resolution + test updates
- adds go to def support for include items when resolvable


msujew (approved):

montymxb:
Added one more commit, forgot to remove the wildcard lib resolution that I marked for removal.

Issue #211, opened 2025-06-24
[Improvement] Semantic tokens for procedure parameters


msujew:
Currently, procedure parameters are not colored at all. We should add semantic token highlighting for them.

See also https://github.com/zowe/zowe-pli-language-support/issues/209.

PR #217, opened 2025-06-26
Add semantic token highlighting for parameter

ssmifi:
https://github.com/zowe/zowe-pli-language-support/issues/211


montymxb (approved):
:octocat: Looks good

PR #220, opened 2025-06-26
Auto-Set Doc Language for PLI Lib & Program Files

montymxb:
This resolves an issue where extension-less PLI files (like `MYLIB`) or files with different extensions than .pli/pl1 would not be recognized as PL/I files within vscode. This change optimistically assigns PL/I as the language for a document when the file is:
- located in a known PL/I libs folder declared in any process group with any extension (ex. `MYLIB.pli`, `MYLIB.txt`, or `MYLIB`)
- or is reachable as a program config entry (by literal or glob match) as a an extension-less file (ex. `MYLIB.pli` or `MYLIB`)

Assignment is done lazily when the file is opened, so up to that point the file may still be recognized with a different language type.

The differentiation on the later case is to avoid accidentally picking up other legitimate program types that may live alongside pli files, such as for COBOL. However we'll likely to have to observe how usage is impacted by this, and will adjust as needed.

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -97,7 +97,8 @@
     "stream-browserify": "^3.0.0",
     "util": "^0.12.5",
     "vscode-languageclient": "~9.0.1",
-    "vscode-languageserver": "~9.0.1"
+    "vscode-languageserver": "~9.0.1",
+    "minimatch": "^10.0.3"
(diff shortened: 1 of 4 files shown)


montymxb:
Closing out in favor of revising this post-release, and in the likelihood we will probably chip away at this problem in later PRs.

PR #226, opened 2025-06-30
telemetry activation

asatklichov:


packages/vscode-extension/src/extension/main.ts
```diff
@@ -192,23 +215,43 @@ function watchPlipluginFolder(
   // watch for folder create/delete events
   folderWatcher.onDidCreate(() => {
     client.sendNotification(WorkspaceDidChangePlipluginConfigNotification);
+    sendTelemetryEvent(
+      "pli.language.support.onDidCreate.folder",
+      telemetryReporter,
+    );
```
msujew, line 221: The `folderWatcher` and `fileWatcher` only target the `.pliplugin` folder. Other folders are not watched. Is that the expected behavior?

msujew (approved):
Alright then, fine with me 👍

PR #228, opened 2025-06-30
Add pli file detection for plain text documents

ssmifi:
Proposal to detect pli files and set the language automatically. 

Since the firstLine feature is relatively useless in combination with pli, we could add a minimal listener for specific pli keywords if no language was determined yet. This also works for new documents, which are also not covered by firstLine. We can of course add both or only one of the alternatives. Please let me know what you think. 

Maybe we should restrict the test to the first line, since `DECLARE` is not that uncommon.


montymxb (commented):
General note about the firstLine check to limit what is picked up.

montymxb (approved):
:octocat: Nice good to go!

PR #229, opened 2025-06-30
Fix vulnerable and outdated dev dependencies

msujew:


asatklichov (approved):
Thanks for update, let's merge it to dev, and see the new scan.

PR #238, opened 2025-07-01
Refactor symbol table iterator

didrikmunther:
Make it a bit more readable


msujew (approved):
Thanks, that makes it easier to understand 👍

PR #239, opened 2025-07-02
Support `%INSCAN` statements

msujew:
Supports both the `%INSCAN` and `%XINSCAN` statements. Also, simplified `XINCLUDE` handling to just use one token and renames the `xInclude` property to `idempotent` throughout the whole preprocessor stack.

### DIFF
--- a/scripts/merge-tmlanguage.mjs
+++ b/scripts/merge-tmlanguage.mjs
@@ -50,6 +50,9 @@ const controlKeywords = [
     'begin',
     'process',
     'include',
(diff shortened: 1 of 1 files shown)


montymxb (approved):
:octocat: approved, looks good!

A small suggestion is to update the syntax highlighting to pickup XINCLUDE & XINSCAN as well (seems like INSCAN is also not present), but it's not blocking.

Idempotent prop is a nice touch by the way 👍

PR #244, opened 2025-07-03
Add pli file detection listener for plain text documents

ssmifi:
Split https://github.com/zowe/zowe-pli-language-support/pull/228 in two PRs, so that we can merge the lightweight firstLine immediately and keep this for after the release, so that we can test/extend it during the next cycle.

Proposal to detect pli files and set the language automatically.

Since the firstLine feature is relatively useless in combination with pli, we could add a minimal listener for specific pli keywords if no language was determined yet. This also works for new documents, which are also not covered by firstLine. We can of course add both or only one of the alternatives. Please let me know what you think.

We can extend the keywords or detection mechanism.


ssmifi:
We rely on the simple firstline check atm, so, I'm closing this one for now.

PR #245, opened 2025-07-03
Move semantic token tests to harness

didrikmunther:


msujew (approved):
Thanks, makes sense 👍

PR #253, opened 2025-07-03
Add reference of preprocessor variables in PL/I code

msujew:
Allows users to use "Find Definition" and "Find References" on PL/I tokens that are actually references to preprocessor variables.


montymxb (approved):
:octocat: Approved, good to go!

PR #255, opened 2025-07-03
Bugfix: Don't Parse Standalone Lib Files

montymxb:
Standalone libs files should not be parsed outside of an existing compile unit's context (from a known program that matches a known program config). If another program's compile unit already includes this lib file, we use that one, but if none exists we don't attempt to parse the file

In the case where a given file is not a lib file, and not a known program entry point, we should still parse it (but we won't handle any include resolution).

### DIFF
--- a/packages/language/src/workspace/compilation-unit.ts
+++ b/packages/language/src/workspace/compilation-unit.ts
@@ -31,6 +31,7 @@ import {
   TextDocuments,
 } from "../language-server/text-documents.js";
 import { Builtins, BuiltinsUri, BuiltinsUriSchema } from "./builtins.js";
+import { PluginConfigurationProviderInstance } from "./plugin-configuration-provider.js";
 import { EvaluationResults } from "../preprocessor/instruction-interpreter.js";
 
 /**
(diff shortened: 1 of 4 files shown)


packages/language/src/workspace/compilation-unit.ts
```diff
@@ -177,15 +178,23 @@ export class CompilationUnitHandler {
...
-    return (
-      this.compilationUnits.get(uri.toString()) ||
-      this.createAndStoreCompilationUnit(uri)
-    );
+  getOrCreateCompilationUnit(uri: URI): CompilationUnit | undefined {
+    if (this.compilationUnits.has(uri.toString())) {
+      // existing compilation unit
+      return this.compilationUnits.get(uri.toString());
+    } else if (
+      !PluginConfigurationProviderInstance.isLibFileCandidate(uri.toString())
```
msujew, line 190: We need to skip the encoding here, otherwise the `c:/` on windows gets encoded into `c%3A/`

```suggestion
      !PluginConfigurationProviderInstance.isLibFileCandidate(uri.toString(true))
```

PR #257, opened 2025-07-04
Add ambiguous reference error on entire reference

didrikmunther:


packages/language/src/linking/error.ts
```diff
@@ -11,15 +11,64 @@
...
 import { PliValidationAcceptor } from "../validation/validator";
+import { CompilationUnit } from "../workspace/compilation-unit";
 import { QualifiedSyntaxNode } from "./qualified-syntax-node";
 
+function mergeRanges(range1: Range, range2: Range): Range {
+  return {
+    start: Math.min(range1.start, range2.start),
+    end: Math.max(range1.end, range2.end),
+  };
+}
```
msujew, line 31: Suggestion: We can make this more general (by using an array), and we need to shield against `NaN`:
```suggestion
function mergeRanges(...ranges: Range[]): Range {
  return {
    start: Math.min(ranges.map(r => r.start).filter(n => !isNaN(n)),
    end: Math.max(ranges.map(r => r.end).filter(n => !isNaN(n)),
  };
}
```

packages/language/src/syntax-tree/ast.ts
```diff
@@ -1965,13 +1965,15 @@ export interface ReferenceItem extends AstNode {
   kind: SyntaxKind.ReferenceItem;
   ref: Reference<NamedElement> | null;
   dimensions: Dimensions | null;
+  token: Token | null;
```
msujew, line 1968: Suggestion: Can we reuse the `ref.token` value instead? I don't think it's necessary to store the reference item token twice.

msujew (approved):
Nice, looks good. Just two comments, see below.

Issue #247, opened 2025-07-03
PL/I Findings3


asatklichov:
In provided (internally) source files, please fix the bugs related to compiler options, macros, include files and labels

1. [**DONE**] compiler options validations on: 

BLANK, DDSQL, DEFAULT, CHECK, DECIMAL

2. [**DONE**]  include support - missing validations and nested include support

3. [**DONE**] Label issue 

```
 RGT005: PACKAGE EXPORTS(RGT005);      
  RGT005: PROCEDURE() OPTIONS(MAIN);  

    DCL SYSNULL BUILTIN;
    DCL  P1     POINTER;              
    P1 = SYSNULL;                     
    DCL (I,J) CHAR(32);               
    I = 'This is a test string';      
    J = SUBSTR(I,1,10);               
    PUT SKIP LIST(I);                 
    PUT SKIP LIST(J);                 
  END RGT005;                         
 END RGT005;    
```
 
4. [**OPEN**] Macro, margin issues are defined in [Findings](https://github.com/zowe/zowe-pli-language-support/issues/186)

PR #258, opened 2025-07-04
Fixed compiler option issues

ssmifi:
Fixed compiler option with empty token name
Made compiler option value comma optional
Fixed validation on BLANK, BRACKETS, CHECK, DECIMAL, DEFAULT, DDSQL (https://github.com/zowe/zowe-pli-language-support/issues/247)


msujew (commented):
Looks mostly good. Just one improvement on the parser changes 👍

PR #259, opened 2025-07-04
Fix package scoping

didrikmunther:


packages/language/src/syntax-tree/ast.ts
```diff
@@ -1294,10 +1296,15 @@ export interface ExecStatement extends AstNode {
 export interface ExitStatement extends AstNode {
   kind: SyntaxKind.ExitStatement;
 }
+export interface ExportsItem extends AstNode {
+  kind: SyntaxKind.ExportsItem;
+  name: string | null;
```
msujew, line 1301: Suggestion: You can use `reference.text` to get the name of the exported item.

packages/language/src/syntax-tree/ast.ts
```diff
@@ -1965,13 +1972,15 @@ export interface ReferenceItem extends AstNode {
   kind: SyntaxKind.ReferenceItem;
   ref: Reference<NamedElement> | null;
   dimensions: Dimensions | null;
+  token: Token | null;
```
msujew, line 1975: Please remove, see other PR.

PR #260, opened 2025-07-07
Add procedure parameter linking

didrikmunther:


msujew (dismissed):

msujew (commented):
I've noticed an issue with this approach wrt. the scope of the procedure parameters:

```
DCL MAINPARM;
 PLEA:
    PROCEDURE(MAINPARM) OPTIONS(MAIN) REORDER;
```

When declaring the variable outside of the procedure, it still links to the procedure - however, I would expect that the linking fails. I'm not sure how doable this is via our current linking system. Note that I haven't tested whether this is valid in the compiler - it might actually be valid to do this. Can you test this?

didrikmunther:
> I've noticed an issue with this approach wrt. the scope of the procedure parameters:
> 
> ```
> DCL MAINPARM;
>  PLEA:
>     PROCEDURE(MAINPARM) OPTIONS(MAIN) REORDER;
> ```
> 
> When declaring the variable outside of the procedure, it still links to the procedure - however, I would expect that the linking fails. I'm not sure how doable this is via our current linking system. Note that I haven't tested whether this is valid in the compiler - it might actually be valid to do this. Can you test this?

Good catch! I tested this on the mainframe:

```pli
 MAIN: PROC OPTIONS(MAIN);

  DCL 1 P1 CHAR(32);

  MYPROC: PROCEDURE(P1);
    PUT(P1);
  END MYPROC;

  CALL MYPROC(1);

 END MAIN;
```

And it does fail with a `IBM1373I E: RULES(NOLAXDCL) violation: variable P1 is implicitly declared.` on the procedure parameter. I will see if there is a way to solve this in the linking system.

msujew (approved):
Looks good to me, thank you very much! 🎉

PR #268, opened 2025-07-09
Avoid missing file tokens

ssmifi:
There are cases when document symbols are requested in which the text document is not included in the current compilation unit.  This occurred when changing an include in the code and than switching between different libraries in the lib folder, which should be investigated further. Nevertheless, the document symbol request should not try to access file tokens that are not present and fail.


msujew (approved):
Looks good, thank you

PR #269, opened 2025-07-09
Restrict Unrecognized Extension Include Resolution

montymxb:
Closes an observed issue where having a file like `cpy/File.xyzabc` would still be included by its extension-less reference, such as `%INCLUDE FILE`. This was due to a missing secondary check where we attempted to resolve a potential match for that include, which has now been added. Also put in a test case to try and sneak around this with a few different include approaches.


msujew (approved):
Makes sense, thank you 👍

PR #272, opened 2025-07-10
Add library chaining

ssmifi:
Proposal for enabling library chaining as requested.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -154,7 +154,8 @@ export interface EvaluationResults {
 }
 
 interface InterpreterContext {
-  uri?: URI;
+  currentUri?: URI;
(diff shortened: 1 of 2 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -857,7 +859,8 @@ function runInclude(item: IncludeItem, context: InterpreterContext): void {
     const instruction = generateInstructions(subProgram.statements);
     const newContext: InterpreterContext = {
       ...context,
-      uri: uri,
+      currentUri: uri,
+      entryUri: context.entryUri,
```
msujew, line 863: Minor: We don't need to explicitly state this, since we use the `...context` spread operation, which already sets `entryUri` to `context.entryUri`.

msujew (approved):
Thank you 👍

PR #284, opened 2025-07-16
add display name and fix gif

asatklichov:
add display name change and fix gif


zeibura:
tweaked Code4z icon
otherwise good to go!

zeibura (approved):

Issue #287, opened 2025-07-21
Too much CPU  consumed


asatklichov:
When I use PL/I extension in VSCode, i see my CPU usage increased and when I disabled it it drops down with high margin. 

<img width="734" height="212" alt="Image" src="https://github.com/user-attachments/assets/d01ea1a6-24b8-4c01-a88e-134f8eca884a" />

PR #288, opened 2025-07-21
Fix endless event loop in ruler settings

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/287

The code as it is currently results in the VSCode extension host calling our settings update routine all the time, as we update the settings in there, which results in VSCode calling ..., etc. you know the drill. This essentially results in the thread that hosts the VS Code extension host to endlessly call our code.

This change ensures that we don't actually change the setting in case nothing has changed.


ssmifi (approved):
LG, thanks!

PR #290, opened 2025-07-22
Sets up PP INCLUDE Option Handling (Plus Others)

### LINKED ISSUE #171: New ++INCLUDE support and %INCLUDE improvements
1. %INCLUDE preprocessor should work (include) by default with copybooks reside in configured folder (e.g. INC  SYLIB etc.)   in configuration file 

`%INCLUDE abc;     `

2. Define it with compiler option in compound configuration file to make ++INCLUDE work  in source file. 
e.g. once you define below in config file

`PP(INCLUDE('ID(++INCLUDE)'))`   

then you can able to start using 

`++INCLUDE abc     `     (no semicolon at the end)

Both type of INCLUDE can work together

Image

montymxb:
Closes #171 by adding remaining handling for the `PP(INCLUDE('ID(++INCLUDE)'))` option case. Other cases are now picked up as well so we have access to them, but they're not yet handled (i.e. PPINCLUDE, PPCISC, PPLIST, etc.). But the pattern is pretty clear on how to proceed with these.

### DIFF
--- a/packages/language/src/parser/tokens.ts
+++ b/packages/language/src/parser/tokens.ts
@@ -192,13 +192,13 @@ export function escapeRegExp(value: string): string {
 export function setCompilerOptions(options: CompilerOptions): void {
   const orChars = escapeRegExp(options.or || "|");
   const notChars = escapeRegExp(options.not || "¬^");
-  let includeAlt: string | undefined = '++include';
+  let includeAlt: string | undefined = "++include";
   // take last option if multiple are present
   if (Array.isArray(options.pp)) {
     for (let x = options.pp.length - 1; x >= 0; x--) {
       const item = options.pp[x];
       const val = item.value;
-      if (item.name.toLowerCase() === 'include' && val) {
+      if (item.name.toLowerCase() === "include" && val) {
         // check for the ID(...) syntax, denoting a new include alternative
         const match = val.match(/ID\(([^\)]+)\)/);
         if (match) {
--- a/packages/language/src/parser/tokens.ts
+++ b/packages/language/src/parser/tokens.ts
@@ -192,7 +192,23 @@ export function escapeRegExp(value: string): string {
 export function setCompilerOptions(options: CompilerOptions): void {
   const orChars = escapeRegExp(options.or || "|");
   const notChars = escapeRegExp(options.not || "¬^");
-  const includeAlt = "++include";
+  let includeAlt: string | undefined = '++include';
+  // take last option if multiple are present
+  if (Array.isArray(options.pp)) {
+    for (let x = options.pp.length - 1; x >= 0; x--) {
+      const item = options.pp[x];
+      const val = item.value;
+      if (item.name.toLowerCase() === 'include' && val) {
+        // check for the ID(...) syntax, denoting a new include alternative
+        const match = val.match(/ID\(([^\)]+)\)/);
+        if (match) {
+          includeAlt = match[1].toLowerCase();
+          break;
+        }
+      }
+    }
+  }
+
   if (includeAlt) {
     includeAltRegex = new RegExp(escapeRegExp(includeAlt), "y");
   } else {
--- a/packages/language/src/parser/tokens.ts
+++ b/packages/language/src/parser/tokens.ts
@@ -192,7 +192,7 @@ export function escapeRegExp(value: string): string {
 export function setCompilerOptions(options: CompilerOptions): void {
   const orChars = escapeRegExp(options.or || "|");
   const notChars = escapeRegExp(options.not || "¬^");
-  let includeAlt: string | undefined = "++include";
+  let includeAlt: string | undefined = undefined;
   // take last option if multiple are present
   if (Array.isArray(options.pp)) {
     for (let x = options.pp.length - 1; x >= 0; x--) {
--- a/packages/language/src/parser/tokens.ts
+++ b/packages/language/src/parser/tokens.ts
@@ -200,9 +200,10 @@ export function setCompilerOptions(options: CompilerOptions): void {
       const val = item.value;
       if (item.name.toLowerCase() === "include" && val) {
         // check for the ID(...) syntax, denoting a new include alternative
-        const match = val.match(/ID\(([^\)]+)\)/);
-        if (match) {
-          includeAlt = match[1].toLowerCase();
+        const match = val.match(/ID\(([^\)]+)\)/g);
+        if (match && match.length > 0) {
+          // take the last match & trim off the ID(...) part
+          includeAlt = match[match.length - 1].slice(3, -1).toLowerCase();
           break;
         }
       }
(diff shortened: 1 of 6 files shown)


packages/language/src/parser/tokens.ts
```diff
@@ -192,7 +192,24 @@ export function escapeRegExp(value: string): string {
...
+        // check for the ID(...) syntax, denoting a new include alternative
+        const match = val.match(/ID\(([^\)]+)\)/g);
+        if (match && match.length > 0) {
+          // take the last match & trim off the ID(...) part
+          includeAlt = match[match.length - 1].slice(3, -1).toLowerCase();
+          break;
+        }
+      }
+    }
+  }
```
msujew: Suggestion: The pp options are actually well known (even if they are just strings from a syntactical perspective, which is why I initially modelled them as `{ name, value }` arrays). IMO, we should have a `PPInclude { value: string }` type that is part of the `PP` type that is already parsed during the translation process. Doing the parsing adhoc in here looks like a code smell.

packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -1456,12 +1456,112 @@ translator.rule(
 );
 
 /** {@link CompilerOptions.pp} */
+translator.rule(["PP"], (option, options) => {
+  // 1 or more pre-processor options to collect
+  ensureArguments(option, 1, Infinity);
```
ssmifi: _Minor_: You can discard the upper bound, if there is none. But Infinity makes sense, too. :)

packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -1456,12 +1456,112 @@ translator.rule(
...
+
+    if (value.values.length !== 1) {
+      throw new TranslationError(
+        value.token,
+        `Expected exactly one value for the ${value.name} option, but received ${value.values.length}.`,
+        1,
+      );
+    }
+
+    ensureType(value.values[0], "string");
```
ssmifi: _Suggestion_: Also validate PP value. According to spec only `CICS`, `INCLUDE`, `MACRO` and `SQL` are currently supported.

We could also check for the `ID` inside the pp-string, but not sure if we want to validate strings in the options currently.

packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -1456,12 +1456,112 @@ translator.rule(
...
+  options.pp.push(...ppOptions);
+});
+
 /** {@link CompilerOptions.ppCics} */
+translator.rule(["PPCICS"], (option, options) => {
+  ensureArguments(option, 1, 1);
+  const value = option.values[0];
+  ensureType(value, "string");
+  options.ppCics = value.value;
+});
```
ssmifi: _Suggestion_: Also add `NOPPCICS`.

packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -1456,12 +1456,112 @@ translator.rule(
...
 /** {@link CompilerOptions.ppTrace} */
+translator.rule(
+  ["PPTRACE"],
+  (option, options) => {
+    ensureArguments(option, 0, 0);
+    options.ppTrace = true;
+  },
+  ["NOPPTRACE"],
+  (_, options) => {
+    options.ppTrace = false;
```
ssmifi, line 1600: _Suggestion_: Also ensure 0 arguments at the negative alternatives (also see rules above).

ssmifi (commented):
Looks good. I have some minor suggestions to improve the reporting of the options translator.

msujew (commented):
Reviewed this already yesterday but forgot to click "submit" :)

packages/language/src/parser/tokens.ts
montymxb: Right, the options themselves are known, but the formatting of the arguments to those options is what's in the value. So we could be getting something like `PP(INCLUDE('ID(-inc) ID(++include) ID(+inc)'))`. I may be overlooking something, but I don't believe we have any handling for value strings like this yet, which is what this implementation is seeking to handle.

However, I do agree that this isn't really ideal in its current form, it's too targeted and does constitute a smell. The goal here is more to focus on checking the include alternative feature & ensuring that it works separately, without touching the rest of the logic, at least not yet. A bit of a calculated move to try and prepare this for an upcoming minor release first.

The rest is a bit more involved, and could be done here, but then I should put this back into draft. There's points such as the option `PPINCLUDE('ID(-inc)')` not being semantically equivalent to `PP(INCLUDE('ID(-inc)'))`, which sets & activates the alternative. My plan was to break that out into a follow-up issue & PR separate from the PP/INCLUDE case since that needs more handling, but what do you think?

msujew: Sure, fine with me 👍

packages/language/src/preprocessor/compiler-options/translator.ts
montymxb: Good idea to validate on the name as well, but I'm planning to leave the string part alone for now. There's a bit more involved in terms of how that needs to work overall.

packages/language/src/parser/tokens.ts
montymxb: As mentioned earlier today I re-read this & changed my mind, I'm going to put the suggestion in, makes a lot of sense. But the handling for how `PP` & `PPINCLUDE` interact will be something to do later. I'll add another issue for that after this is done.

montymxb: Opened up #300 to continue on this one.

packages/language/test/compiler/options-parser.test.ts
```diff
@@ -354,15 +354,23 @@ describe("CompilerOptions translator", () => {
     const issues = result.issues;
     expect(issues).toHaveLength(0);
 
-    const ppItems = result.options.pp;
-    expect(ppItems).toBeDefined();
-    expect(ppItems).toHaveLength(1);
+    const pp = result.options.pp;
+    expect(pp).toBeDefined();
+    expect(pp !== false).toBeTruthy();
+    const items = (pp as any).items as CompilerOptions.PPItem[];
```
msujew: Why do we need to cast to `any` here?

msujew (approved):
Looks pretty good to me. Just a minor testing issue, see below.

ssmifi (approved):
LGTM, thanks!

packages/language/test/compiler/options-parser.test.ts
montymxb: The type isn't narrow enough to access the items prop, there's also `false` as part of the union. This is a lazy workaround, but honestly it's a good question why I'm using `any` in the first place, it's almost always an issue...let me put up one more change to remove that.

montymxb: Done.

PR #293, opened 2025-07-24
Improve document symbol LSP request

msujew:
Improves the type safety of our document/workspace symbol requests. This prevents errors such as:

```
[Error - 11:33:19 AM] Request textDocument/documentSymbol failed.
Error: selectionRange must be contained in fullRange
```

By making sure that we correctly specify the `selectionRange` and `range` items and adjusting the `range` to make sure that all children are contained the range before returning the LSP response. Also ensures that all processed tokens are using valid ranges.

Also fixes a very minor parser ambiguity issue that I've noticed - `ENTRY` was never supposed to be a `DefaultAttribute`, it has it's own `EntryAttribute` type.


ssmifi (approved):
Looks good. Thank you.

Issue #275, opened 2025-07-10
Extract builtin functions which need explicit declaration


asatklichov:
This task is a continuation of [builin-functions](https://github.com/zowe/zowe-pli-language-support/issues/173) task. 
Implementation is done correctly, as you see in below code. 
e.g. If **SUBSTR** (or other example **MEMCU12** ) is not explicitly defined, but compiler does not complain about it. 
But, for **SYSNULL** compiler complains if it does not explicitly defined as BUILTIN.   You must declare it  explicitly as BUILTIN. 
Otherwise it gives 'Unknown Identifier 'SYSNULL' error.

Problem is, IBM specification does not differentiate which built-in functions need explicit declaration. 
[Ref] - IBM Ref PLI V6R1,  Euroclear PL/I compiler version 6.1 in batch IBMZPLI, in Unix PLI 
[Ref] - IBM Language Reference Chapter 18, page 377

So, we need to validate them by PL/I compiler, and extract all such built-in functions, and make it work like SYSNULL.   
 

 ```
RGT005: PACKAGE EXPORTS(RGT005);     
  /* DCL SUBSTR          BUILTIN; */  
  DCL SYSNULL         BUILTIN;        
  RGT005: PROCEDURE() OPTIONS(MAIN);  

    DCL  P1     POINTER;              
    P1 = SYSNULL;                     
    DCL (I,J) CHAR(32);               
    I = 'This is a test string';      
    J = SUBSTR(I,1,10);               
    PUT SKIP LIST(I);                 
    PUT SKIP LIST(J);                 
  END RGT005;                         
 END RGT005;            


```

Issue #172, opened 2025-06-02
Configuring MACROs via compound configuration files


asatklichov:
Using [pgm_config.json, proc_grpc.json](url) configuration you should be able to define macros, 
If not defined in the config file,  it should take it from the DEFAULT compiler option.

Reference - ProgGuide page [Chapter 1 - Compiler Options, Chapter2 - PL/I Preprocessors]

Options [schema example with default values](https://github.com/eclipse-che4z/che-che4z-lsp-for-hlasm/blob/development/clients/vscode-hlasmplugin/schema/asm_options.schema.json):  

Some examples, which customer uses more often:

Page 40 - Ref. Prog-Guide
1) `INCAFTER(PROCESS(FWKMACRO))`  - option specifies a file to be included after a particular statement in your source program

~2)  `SYSPARM('PLATFORM(EB),NAME(${PROGRAM_NAME}),TYPE(SRC)')  `  with parameterized support. 
The SYSPARM option specifies the value of the string that is returned by the macro facility built-in function
SYSPARM.~
Broken out into a separate issue, #312 .

3) `SYSTEM` - The SYSTEM option specifies the format used to pass parameters to the MAIN PL/I procedure, and
generally indicates the host system under which the program runs.

<img width="619" height="253" alt="Image" src="https://github.com/user-attachments/assets/17c0edfa-9cf9-4415-a32e-8641f5d1f65a" />

<img width="440" height="230" alt="Image" src="https://github.com/user-attachments/assets/37763adb-2281-41bd-9885-0534463e9aef" />

**Just remember that PP and MACRO option is similar, but differentiate in a way that:** 
With the PP option, you can define preprocessors with order (sequentially), so output from first one can be used as input for the second preprocessor and etc..  e.g. PP(INCLUDE('ID(++INCLUDE)'), macro2, macro3), which you can not manage with MACRO. 
 

<img width="436" height="227" alt="Image" src="https://github.com/user-attachments/assets/deca4734-9ce6-428e-a8ba-74fe3dc77ae6" />

<img width="446" height="208" alt="Image" src="https://github.com/user-attachments/assets/40add5b4-6a45-420d-99d4-bf70eb620d05" />

montymxb:
Please see #312 for continued work on the SYSPARM parametric arg feature.

montymxb:
Closing as resolved by #295 (for the cases outlined here).

PR #294, opened 2025-07-24
Add configure property and validation for implicit builtins

ssmifi:
Concludes https://github.com/zowe/zowe-pli-language-support/issues/275 

In our tests only `SUBSTR` is implicitly available on the system we have access to, but this must not be the case in general. Therefore, developer can add additional implicit builtins in the process group config. 

(@montymxb:) Also added program configurations and process groups to the `TestBuilder` (which can be a starting point for a more sophisticated system in https://github.com/zowe/zowe-pli-language-support/issues/172 if needed.)

Cleaned up validation naming a bit: General stuff in `validator.ts` now usually omits the Pli prefix, whereas Pli specific code in `pli-validator.ts` includes it. Also, it would be beneficial for some checks to not evaluate instances, such process groups, every time, so I made the `PliValidator` persistent. However, in the current state, the functions must be explicitly bound if they require state, which is not optimal. We should improve the validator architecture in this regard in a new PR in the near future.

### DIFF
--- a/packages/language/src/validation/pli-validator.ts
+++ b/packages/language/src/validation/pli-validator.ts
@@ -17,31 +17,28 @@ import {
 } from "../language-server/types";
 import * as AST from "../syntax-tree/ast";
 import { forEachNode } from "../syntax-tree/ast-iterator";
+import { BuiltinsUriSchema } from "../workspace/builtins";
+import { CompilationUnit } from "../workspace/compilation-unit";
+import {
+  PluginConfigurationProviderInstance,
+  ProcessGroup,
+  ProgramConfig,
+} from "../workspace/plugin-configuration-provider";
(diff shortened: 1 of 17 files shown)


packages/language/src/validation/pli-validator.ts
```diff
@@ -273,4 +295,37 @@ export class Pl1Validator {
...
+    // Check if the reference item is a builtin.
+    if (node.ref?.node?.nameToken?.payload.uri?.scheme !== BuiltinsUriSchema) {
+      return;
+    }
+
+    const name = node.ref.node.name.toUpperCase();
+    if (!this.processGroup["implicit-builtins"]?.includes(name)) {
+      acceptor(
+        Severity.E,
+        PLICodes.Error.IBM1373I.message(node.ref?.node?.name),
```
msujew: I believe we don't need the checks.
```suggestion
        PLICodes.Error.IBM1373I.message(node.ref.node.name),
```

msujew (approved):
Looks good to me, thank you 👍

PR #295, opened 2025-07-24
Support Defining Macros in Config + Associated PP Support

montymxb:
WIP for rectifying all issues reported in #172 

- [x] support declaring macros in configs in key/value style

Support initial macro declarations from both the process group & program config files. The program config entries should take precedence over the process group entries. In either case the `pli-options` key is used, following convention in HLASM extension.

Ex. program entries can be like so
```json
{
  "pgms": [
    {
      "program": "*.pli",
      "pgroup": "default",
      "pli-options": {
        "SYSPARM": "'from program config'",
        "SYSTEM": "MVS"
      }
    }
  ]
}

```

And process group entries can be like so:
```json
{
  "pgroups": [
    {
      "name": "default",
      "compiler-options": [
        "AGGREGATE",
        "MARGINS(2,72)"
      ],
      "pli-options": {
        "SYSPARM": "'from process group'",
        "SYSTEM": "MVS",
        "MARGINS": "2,72"
      },
      "libs": [
        "cpy"
      ],
      "include-extensions": [
        ".pli",
        ".cpy",
        ".inc"
      ]
    }
  ]
}
```

When forming compiler options from cofigurations, we factor in pli-options w/ compiler-options in the following order:
1. compiler-options
2. process group pli-options
3. program config pli-options

Later entries in pli-options will replace existing ones.

The above configuration should give:
```
SYSPARM('from program config') SYSTEM(MVS) MARGINS(2,72)
```

- [x] Support SYSPARM and SYSTEM macros, which we don’t handle yet

Invocations of `SYSPARM('some val')` set the builtin value, and later can be retrieved via `SYSPARM` directly. The default value appears to be a _null_ string, but I've also seen `''` mentioned.

SYSTEM indicates the host system & param format to pass to main. It takes a fixed arg set:

- MVS = single char string, no params, or an arbitrary param list _(default for zOS)_
- OS = unix param list
- CICS = pointer(s)
- IMS = pointer(s)
- TSO = pointer to CPPL

Technically we should be changing the function signature of `main` for typechecking based on this, but that should probably come later.

- [x] Support the INCAFTER macro

Pretty striaghtforward, informs the PP to include a given file after the last `*PROCESS`. Can either modify the pre-processor behavior to inject a synthetic %INCLUDE file after the last `*PROCESS`, or just directly invoke the include logic after the last `*PROCESS` directive.

If there is no `*PROCESS` directive this is effectively a noop.

Quick example
```
*PROCESS INCAFTER(PROCESS(myfile));
*PROCESS MARGINS(2,72);
*PROCESS AG;
 // --> %INCLUDE myfile;
```

- [x] Tests for each of these cases (4slash needs to be able to modify configs pre-parse to pick these up)


montymxb:
All set, caveat is that both SYSPARM & SYSTEM are not directly referentiable in code, given we haven't gotten to PP procedures yet.

msujew (approved):
LGTM. Before merging, can you add a test in which `INCAFTER` is used in a program config?

packages/language/test/fourslash/preprocessor/include/incafter-in-config.ts
```diff
@@ -0,0 +1,40 @@
...
+// @filename: cpy/lib.pli
+//// DECLARE LIB_VAR FIXED;
+
+// @filename: main.pli
+////*PROCESS;
+//// DCL TEST FIXED;
+
+preprocessor.expectTokens(`*PROCESS;
+  DCL TEST FIXED;
+`);
```
msujew, line 40: Shouldn't this also contain `DECLARE LIB_VAR FIXED;`?

montymxb: 🤦 yeah, I'll put up another PR shortly

Issue #186, opened 2025-06-10
PL/I findings


asatklichov:
Regarding to below bugs, examples are provided with full description. 

1. [**OK - User acknowledge**] Comments are parsed inconsistently. 
In some files it does not give an error even comment exceeds column 72,  
but in some files it gives 'Symbol not found' error which is non-related.  

2. [**OPEN**] Currently you can start coding PL/I statement at column zero. 
Either PL/I extension or PL/I compiler is not complaining about it.
  
But, problem will arise during runtime (PL/I object module execution).  

So, PL/I program statements must follow below criteria: 
COL1- %PROCESS (or *PROCESS) is the only statement starts from column 1
COL2-72 - PLI STATEMENTS
COL73-80 - SEQUENCE NUMBER USED BY Z/OS 

Give an error message "PL/I statements must start from column 1 or as defined in MARGINS(M,N, C)"

3. [**DONE**] Compiler Options has different errors, sea each file starts with ERR*.pli  
Errors are described in each file. 

4.  [**DONE**]  Macro %ACTIVATE /  %DEACTIVATE have two issues

 a [**DONE**] - coloring is not activated for %ACT, %DEACT 
b[**DONE**] - not supporting NORESCAN option, see macro_activate_or-DEACT.pli

PR #298, opened 2025-07-25
Add diagnostic markers for text outside of margins

ssmifi:
Suggestion for https://github.com/zowe/zowe-pli-language-support/issues/186 (2)

I think we can check for violated boundaries without too much impact when checking each character in `splitLines` when processing the margins anyway. Please let me know if you agree. 

Also, allow tags without value in the test harness. 
Added `@crlf` tag to specifically test for \r\n in the harness to test lines that are processed manually.
(Could be reasonable to automatically test all harness test against both line endings in the future.)

Also, fixed display of the right margin ruler.

### DIFF
--- a/packages/language/src/preprocessor/pli-margins-processor.ts
+++ b/packages/language/src/preprocessor/pli-margins-processor.ts
@@ -9,20 +9,51 @@
  *
  */
 
-import { Range } from "../language-server/types";
+import { Range, Severity } from "../language-server/types";
 import { CompilerOptionsProcessorResult } from "./compiler-options-processor";
+import { LexingIssue } from "./pli-lexer";
+import { URI } from "../utils/uri";
+import { Warning } from "../validation/messages/pli-codes";
+import { PluginConfigurationProviderInstance } from "../workspace/plugin-configuration-provider";
 
 const NEWLINE = "\n".charCodeAt(0);
+const SPACE = " ".charCodeAt(0);
+const PREFIX_PATTERN = /^[0-9\+\- ]+$/;
+const SEQUENCE_PATTERN = /^\s*[A-Z0-9]*\r?\n?$/;
 
 export interface MarginsProcessor {
-  processMargins(input: CompilerOptionsProcessorResult): string;
+  issues: LexingIssue[];
+  processMargins(input: CompilerOptionsProcessorResult, uri: URI): string;
 }
 
 /**
  * Helper class to replace text margins with space characters (characters 2-72 are normal program text)
  */
 export class PliMarginsProcessor implements MarginsProcessor {
-  processMargins(input: CompilerOptionsProcessorResult): string {
+  static readonly MARGIN_ERROR_MESSAGE_LEFT =
+    "PL/I statements must start from column 2 or as defined in MARGINS(M,N).";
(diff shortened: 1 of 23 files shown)


packages/language/src/preprocessor/pli-margins-processor.ts
```diff
@@ -11,18 +11,31 @@
...
 }
 
 /**
  * Helper class to replace text margins with space characters (characters 2-72 are normal program text)
  */
 export class PliMarginsProcessor implements MarginsProcessor {
-  processMargins(input: CompilerOptionsProcessorResult): string {
+  static readonly MARGIN_ERROR_MESSAGE_LEFT =
+    "PL/I statements must start from column 2 or as defined in MARGINS(M,N).";
+  static readonly MARGIN_ERROR_MESSAGE_RIGHT = "Line exceeds right margin!";
```
msujew: Suggestion: There is an actual error code for this, see `IBM1084I`.

packages/language/src/preprocessor/pli-margins-processor.ts
```diff
@@ -11,18 +11,31 @@
...
+  issues: LexingError[];
+  processMargins(input: CompilerOptionsProcessorResult, uri: URI): string;
 }
 
 /**
  * Helper class to replace text margins with space characters (characters 2-72 are normal program text)
  */
 export class PliMarginsProcessor implements MarginsProcessor {
-  processMargins(input: CompilerOptionsProcessorResult): string {
+  static readonly MARGIN_ERROR_MESSAGE_LEFT =
```
msujew, line 33: Question: I can't seem to find an error code for this. Did you confirm the behavior on the mainframe?

ssmifi: There is no specific error code for a left margin violation on the mainframe. The preprocessor truncates the margin and the remaining text is parsed. `ND` was `END`:

> [Severe] IBM3988I at line 10: Statement has invalid syntax.
> [Error] IBM3552I at line 10: The statement element ND is invalid. The statement will be ignored.

I followed @asatklichov's request in https://github.com/zowe/zowe-pli-language-support/issues/186 and report the error message he proposed to give the developer some feedback.

msujew: We should probably try to treat it as a warning and also add a config option (into the `.pliplugin` config) to disable/enable this behavior. There might be valid use cases to add data into the first n columns and users might get annoyed if we flag every character with an error. Perhaps we can add `lsp-options` in there to make sure these are language server specific PLI options?

ssmifi: Sounds good to me. Implemented the `lsp-options` and changed it to warnings.

packages/vscode-extension/schemas/proc_grps.schema.json
```diff
@@ -41,6 +41,14 @@
...
               "type": "string"
             }
+          },
+          "lsp-options": {
+            "type": "object",
+            "properties": {
+              "checkMargins": {
+                "type": "boolean"
+              }
+            }
```
msujew: Minor: Please add descriptions here. These are helpful for adopters to explore the different options available. Also, maybe we should keep everything in kebab-case, i.e. `check-margins`?

msujew (approved):
Thank you, just one more change :)

Issue #132, opened 2025-05-14
[Feature] Support %DO SKIP


ssmifi:
Support the `%DO SKIP;` statement.

msujew:
Note for anyone trying to implement the statement, this is what the language reference says about the statement:

> This statement causes all code through the matching `%END` statement to be ignored (and thus can be useful as a way of "commenting out" code that contains comments).

It essentially acts as a comment statement, but on a preprocessor level.

PR #299, opened 2025-07-28
Add support for %DO SKIP;

ssmifi:
Add support for `%DO SKIP;` (https://github.com/zowe/zowe-pli-language-support/issues/132)

### DIFF
--- a/packages/language/src/preprocessor/pli-preprocessor-parser.ts
+++ b/packages/language/src/preprocessor/pli-preprocessor-parser.ts
@@ -572,7 +572,7 @@ export class PliPreprocessorParser {
         CstNodeKind.DoStatement_Semicolon0,
         PreprocessorTokens.Semicolon,
       );
-      this.doSkip(state);
+      this.statements(state);
       statement.skip = true;
       statement.end = this.endStatement(state);
       state.consume(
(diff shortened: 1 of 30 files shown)


msujew:
@ssmifi Please rebase your PR, `development` has been rebased on `main`.

ssmifi:
> @ssmifi Please rebase your PR, `development` has been rebased on `main`.

Already on it.

packages/language/src/preprocessor/pli-preprocessor-parser.ts
```diff
@@ -544,9 +544,37 @@ export class PliPreprocessorParser {
...
+    );
+
+    return statement;
+  }
+
   doStatement(state: PreprocessorParserState): ast.DoStatement {
     const statement = ast.createDoStatement();
     state.consume(statement, CstNodeKind.DoStatement_DO, PreprocessorTokens.Do);
+
+    if (state.canConsume(PreprocessorTokens.Skip)) {
```
msujew, line 563: Can the `DO SKIP` instruction be merged with others? I.e. `DO SKIP; FOREVER`? Because the way this is written would allow this.

ssmifi: I double checked on the mf (https://github.com/zowe/zowe-pli-language-support/issues/311). There, it can. 
The spec is rather vague about it.

msujew: Interesting, thank you for investigating it!

ssmifi:
Factored in findings from https://github.com/zowe/zowe-pli-language-support/issues/311 regarding skipped code and excess `%`.

packages/language/src/preprocessor/pli-preprocessor-parser.ts
```diff
@@ -728,6 +746,31 @@ export class PliPreprocessorParser {
...
+          state.index--; // Get the % for the end statement.
+          break;
+        }
+        nestingLevel--;
+      }
+      // TODO ssmifi: If a *preprocessor* statement is encountered that is not valid, the parser should throw an error.
+      // For example, %xyz; is not valid, whereas xyz; is ignored.
+      // Assignments to declared pp variables are considered valid.
+      state.index++;
+    }
```
msujew: Suggestion: Why not simply resume with the normal parsing? Doing this kind of lexical analysis is prone to errors, for example:

```
%DO SKIP;
  %IF X THEN;
  %END;
%END;
```

Since the text inside of the `SKIP` needs to be valid PLI code anyway, we should definitely just use the normal parser for that instead of trying to skip to the respective `%END` statement.

ssmifi: Not necessarily. As long as it is not an invalid **pre**-processor statement, invalid statements are ignored. Normal parsing would not allow `%DO SKIP; xyz;`, correct? If so, should we ignore that for now?

msujew: Normal parsing would allow that. `xyz;` will be part of a `TokenStatement`, which will simply be ignored as the `%DO SKIP` statement does not generate an instruction.

ssmifi: Cool, thanks you. Did not know this. Updated.

msujew (approved):
Thank you, looks good to me 👍

PR #302, opened 2025-07-28
Fix builtin validation uri

ssmifi:
The validation should use the uri of the reference item.


msujew (approved):

msujew (commented):
Perhaps our tests should also catch that the diagnostics are shown on the correct URI?

ssmifi:
Yea, I also thought about that. I have to look at the test harness. I think it must be extended for that, but I can be mistaken. I would open a separate issue for that if that is ok with you.

msujew:
@ssmifi Fine with me 👍

PR #307, opened 2025-07-28
[Bugfix] Correct INCAFTER Test

montymxb:
Follows up on #295 w/ a correction to an existing test. The prior case was correct, but misleading, as there was an extra `DCL` statement being checked that was unrelated to the incafter effect under test.

This also adds support for the additional `INCAFTER()` compiler option usage, which allows setting back to the default state, and in turn makes the test a bit cleaner in terms of its expectations.

Small correction in an example file as well.

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/translator.ts
+++ b/packages/language/src/preprocessor/compiler-options/translator.ts
@@ -1354,7 +1354,6 @@ translator.rule(["INCAFTER"], (option, options) => {
   ensureArguments(option, 1, 1);
   const value = option.values[0];
   if (value.kind === SyntaxKind.CompilerOption) {
-    ensureType(value, "option");
     if (value.name.toUpperCase() !== "PROCESS") {
       throw new TranslationError(
         value.token,
--- a/packages/language/src/preprocessor/compiler-options/translator.ts
+++ b/packages/language/src/preprocessor/compiler-options/translator.ts
@@ -1353,21 +1353,26 @@ translator.flag("goff", ["GOFF"], ["NOGOFF"]);
 translator.rule(["INCAFTER"], (option, options) => {
   ensureArguments(option, 1, 1);
   const value = option.values[0];
-  ensureType(value, "option");
-  if (value.name.toUpperCase() !== "PROCESS") {
-    throw new TranslationError(
-      value.token,
-      `Expected "PROCESS", but received '${value.name}'.`,
-      1,
-    );
+  if (value.kind === SyntaxKind.CompilerOption) {
+    ensureType(value, "option");
+    if (value.name.toUpperCase() !== "PROCESS") {
+      throw new TranslationError(
+        value.token,
+        `Expected "PROCESS", but received '${value.name}'.`,
+        1,
+      );
+    }
+    ensureArguments(value, 1, 1);
+    const processValue = value.values[0];
+    ensureType(processValue, "plain");
+    options.incAfter = {
+      process: processValue.value,
+      token: processValue.token,
+    };
+  } else {
+    ensureType(value, "plain");
+    options.incAfter = undefined;
   }
-  ensureArguments(value, 1, 1);
-  const processValue = value.values[0];
-  ensureType(processValue, "plain");
-  options.incAfter = {
-    process: processValue.value,
-    token: processValue.token,
-  };
 });
 /** {@link CompilerOptions.incDir} */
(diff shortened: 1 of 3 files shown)


packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -1353,21 +1353,26 @@ translator.flag("goff", ["GOFF"], ["NOGOFF"]);
...
   const value = option.values[0];
-  ensureType(value, "option");
-  if (value.name.toUpperCase() !== "PROCESS") {
-    throw new TranslationError(
-      value.token,
-      `Expected "PROCESS", but received '${value.name}'.`,
-      1,
-    );
+  if (value.kind === SyntaxKind.CompilerOption) {
+    ensureType(value, "option");
```
msujew: Don't we already know that the `kind` is an `option`? We shouldn't need to verify it anymore.

packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -1353,21 +1353,26 @@ translator.flag("goff", ["GOFF"], ["NOGOFF"]);
...
+    ensureArguments(value, 1, 1);
+    const processValue = value.values[0];
+    ensureType(processValue, "plain");
+    options.incAfter = {
+      process: processValue.value,
+      token: processValue.token,
+    };
+  } else {
+    ensureType(value, "plain");
+    options.incAfter = undefined;
```
msujew, line 1373: Question: Is `INCAFTER()` a valid standalone option? The syntax diagram [here](https://www.ibm.com/docs/en/pli-for-aix/3.1.0?topic=options-incafter) specifies that it should always have a `PROCESS` value in there.

packages/language/test/fourslash/preprocessor/include/incafter-in-config.ts
```diff
@@ -33,8 +33,7 @@
 
 // @filename: main.pli
 ////*PROCESS;
-//// DCL TEST FIXED;
 
-preprocessor.expectTokens(`*PROCESS;
-  DCL TEST FIXED;
+preprocessor.expectTokens(`*PROCESS INCAFTER();
```
msujew: Question: Why do we need the `INCAFTER()` here? I don't think it should be required.

packages/language/src/preprocessor/compiler-options/translator.ts
montymxb: It's valid for the [6.1 spec](https://www.ibm.com/docs/en/SSY2V3_6.1/pdf/pg.pdf), and it's the default as well there. It looks like the spec you're referring to is an older one on a different platform, 3.1.0 on AIX.

packages/language/test/fourslash/preprocessor/include/incafter-in-config.ts
montymxb: It is for the moment, but I initially thought the same as well. The updated process group in the test is also impacting the resulting token stream here, so we're getting more tokens than expected. I chatted with @ssmifi about this yesterday, he said he was going to take a look there. I think we all agree, this part should be independent from any config changes we apply to the tests.

Beyond that using the `INCAFTER()` options works now, and will also work as expected after we address this config issue.

packages/language/src/preprocessor/compiler-options/translator.ts
montymxb: Yep you're right, I'll adjust that.

montymxb:
Factored in new test updates brought in by @ssmifi , resetting the modified plugin config before running additional checks that invoke the lifecycle. This avoids the need to have the extra `INCAFTER()` approach to make this test work (although it's still supported), and should hopefully make the test intent much clearer now

packages/language/src/preprocessor/compiler-options/translator.ts
msujew: I see, thanks I didn't notice. That was just the first result that popped up when searching for `INCAFTER pli` :D

packages/language/test/fourslash/preprocessor/include/incafter-in-config.ts
montymxb: Forgot to update here, with the new changes from @ssmifi we don't need to invoke this directly for the test's sake.

msujew (approved):
Looks good to me, thank you 👍

Issue #306, opened 2025-07-28
[Improvement] Add support to the test harness to check for correct URIs in the diagnostics


ssmifi:
Add support to the test harness to check for correct URIs in the diagnostics as mentioned in https://github.com/zowe/zowe-pli-language-support/pull/302.

PR #308, opened 2025-07-29
Test harness extensions

ssmifi:
Several small extensions to the test harness:
- Check range diagnostics in the the `TestBuilder` for correct URIs automatically (https://github.com/zowe/zowe-pli-language-support/issues/306)
- Reset `PluginConfigurationProvider` after harness tests 
- Expose the `TestBuilder` in the test API so that tests can access it
- Expose the non-exclusive expect methods for diagnostics and error codes
- Allow // in the code section of the harness tests to enable comments during debugging
- _Addition_: Actually, reset the `PluginConfigurationProvider` after the `test-builder` has finished its lifecycle, so that potential test methods that invoke lifecycle functions themselves are not affected by the test config.
- _Addition_: Add `expectToThrow` to the harness interface.

@montymxb: Please check regarding the plugin configurations. Thank you!


packages/language/test/fourslash/diagnostic-uris.ts
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="framework.ts" />
+
+// @filename: main.pli
+//// RGT005: PACKAGE EXPORTS(RGT005);
```
montymxb: Package label meant to be the same intentionally?

packages/language/test/test-builder.ts
```diff
@@ -98,13 +99,14 @@ function replaceNamedIndicesWithDocument(file: PliTestFile): TestFile {
...
-  private diagnostics: Diagnostic[];
-  private options: TestBuilderOptions;
+  // We expose all members here, so that tests may access them directly via the harness interface.
+  public unit: CompilationUnit;
+  public files: Map<string, TestFile> = new Map();
+  public output: string;
+  public indices: Record<string, number[]>;
+  public ranges: Record<string, Array<[number, number]>>;
+  public diagnostics: Diagnostic[];
+  public options: TestBuilderOptions;
```
montymxb: Unless we really need all of these public, I would just expose the ones we absolutely need. It might also be better to hide them behind getters if read-only access is all that's needed, rather than exposing all properties directly for r/w access.

packages/language/test/test-builder.ts
```diff
@@ -338,8 +406,12 @@ export class TestBuilder {
 
   expectDiagnosticsAt(
     label: string,
-    diagnostics: Partial<Diagnostic>[],
+    diagnostics: Partial<Diagnostic> | Partial<Diagnostic>[],
```
montymxb, line 413: Function still indicates this should be plural, I would opt to keep the original type to match that.

montymxb (commented):
General feedback. I checked out the plugin behavior, and it looks like it's still persisting. The test file in question is **incafter-in-config.ts**, which should fail in its current state as soon as this is corrected. Right now the program config is still being applied to add in an extra include, same as the file under test. Looks like that's stemming from test functions like `expectPreprocessorTokens` still running under the modified config.

packages/language/test/test-builder.ts
ssmifi: `expectErrorCodesAt`, `expectExclusiveErrorCodesAt`, `expectExclusiveDiagnosticsAt` already had the singular/plural pattern. I just aligned this one. So, should we keep it to be consistent?

ssmifi:
I see (and discussed with @montymxb in more detail). I will work on a solution later today.

packages/language/test/test-builder.ts
montymxb: Ahh, okay then let's just opt to be consistent w/ the other implementations, wasn't aware this was the pattern we were using.

packages/language/test/fourslash/diagnostic-uris.ts
ssmifi: Yes, but I changed the IDs in the test to carry more "meaning". :)

packages/language/test/fourslash/preprocessor/include/incafter-in-config.ts
```diff
@@ -36,5 +36,6 @@
 //// DCL TEST FIXED;
 
 preprocessor.expectTokens(`*PROCESS;
```
montymxb: We can also drop *PROCESS here, it doesn't add anything extra at this point.

montymxb (approved):
:octocat: looks good, just one non-blocking point here

packages/language/test/fourslash/diagnostic-uris.ts
```diff
@@ -0,0 +1,26 @@
...
+////   DCL  P1     POINTER;
+////   P1 = <|1:SYSNULL|>;
+//// END TB0001;
+//// END INTERNAL;
+
+testAPI.testBuilder.getDiagnostics().forEach((diagnostic) => {
+  diagnostic.uri = "file:///sysnull.pli";
+});
+testAPI.testBuilder.checkDiagnosticsURIs({ suppressTestError: true });
+verify.expectErrorCodesAt(1, code.Internal.DiagnosticURIMismatch.fullCode);
```
msujew: Suggestion: IMO adding a synthetic diagnostic in there seems like a workaround for not using something like `verify.expectToThrow(() => testAPI.testBuilder.checkDiagnosticsURIs())`. I think we should do that instead.

msujew (changes requested):

msujew (approved):
Looks good, thanks (after fixing the compiler error)

packages/language/test/fourslash/diagnostic-uris.ts
ssmifi: Sounds good, thanks. I've added `expectToThrow` to the harness interface.

Issue #296, opened 2025-07-25
Implement rest of compiler options together with config option alternative


asatklichov:
For some compiler options currently we have compiler options support via %PROCESS [copm-option]
and via configuration file via element "compiler-options", e.g. "compiler-options":["MARGINS(2,72)"]. 

See the options implemented and un-implemented [here](https://github.com/zowe/zowe-pli-language-support/blob/development/packages/language/src/preprocessor/compiler-options/translator.ts) : 

Follow the compiler-options from specification - Programming Guide, page 3.

PR #314, opened 2025-08-04
Extend compiler options and tests

ssmifi:
I added the first batch of missing compiler options (https://github.com/zowe/zowe-pli-language-support/issues/296): `gonu` - `ign`.
To test the compiler options conveniently in the new harness, I added new error codes for the options and extended the harness. I created the draft so that we can agree on further steps, before I implement the rest of the options.

- Added `gonumber`, `graphic`, `header`, `hgpr`, `ignore` compiler options.
- Added harness functions to test compiler options and specific diagnostics. 
- `expectDiagnosticsAt` also takes arrays for labels now to test for similar diagnostics at different locations (should/could be added for other expect methods as well in the future).
- The test runner now also creates `describe` blocks according to the relative path structure.
- Added compiler option codes for more convenient testing (should be completed over time).

The compiler option task (https://github.com/zowe/zowe-pli-language-support/issues/296) will be split up into several PRs to make reviewing more convenient.

### DIFF
--- a/packages/language/test/fourslash-harness/harness-interface.ts
+++ b/packages/language/test/fourslash-harness/harness-interface.ts
@@ -12,12 +12,13 @@
 import { CompletionKeywords } from "../../src/language-server/completion/keywords";
 import { Diagnostic } from "../../src/language-server/types";
 import { InternalCodes, PLICodes } from "../../src/validation/messages";
-import { ExpectedCompletion, TestBuilder } from "../test-builder";
+import { ExpectedCompletion, Label, TestBuilder } from "../test-builder";
 import { type Severity } from "../../src/language-server/types";
 import { SemanticTokenTypes } from "vscode-languageserver-types";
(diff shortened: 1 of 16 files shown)


packages/language/test/fourslash-harness/harness-interface.ts
```diff
@@ -71,6 +72,12 @@ export interface HarnessTesterInterface {
      */
     noDiagnostics(label?: string | number): void;
 
+    /**
+     * Expect that the compilation unit has no diagnostics apart from the given regexes.
+     * @param regexes The regexes to expect no diagnostics apart from.
+     */
+    noDiagnosticsApartFrom(regexes: RegExp[]): void;
```
msujew: Minor: naming

```suggestion
    noDiagnosticsExcept(regexes: RegExp[]): void;
```

msujew (approved):
Looks pretty good to me. Just one comment wrt. naming. Please replace all the "apart from" occurences with "except".

PR #318, opened 2025-08-05
Extend compiler options and tests #2

ssmifi:
Based on https://github.com/zowe/zowe-pli-language-support/pull/314

- Implemented compiler options `incDir`, `include`, `incPds`, `initAuto`, `initBased`, `initCtl`, `initStatic`, `insource`, `interrupt`,  `json`, `langLvl`, `limits`, `lineCount`, `lineDir`, `list`, `listView`, `lp`, `macro`, `map`, `margini`, `maxbranch`, `maxgen`, `maxmem`, and `maxmsg`.
- Dupes and Mutex are now still reported if the corresponding rule throws another error.
- `noDiagnostics` also accept arrays now.

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/codes.ts
+++ b/packages/language/src/preprocessor/compiler-options/codes.ts
@@ -13,121 +13,303 @@ import { Severity } from "../../language-server/types";
 import { ParametricPLICode } from "../../validation/messages/pli-codes";
 
 export const CompilerOptionsCodes = {
-  WrongParameterCount: {
-    code: "_COOP01",
+  // At the moment, the actual codes are only used internally.
+  // Only the message is presented to the user.
+  InvalidParameterCount: {
+    code: "COOP01",
     severity: Severity.W,
-    message: (received: string, min: string, max?: string) => {
+    message: (received: number, min: number, max?: number) => {
       if (min === max) {
-        return `Expected ${min} arguments, but received ${received}.`;
-      } else if (max === "") {
-        return `Expected at least ${min} arguments, but received ${received}.`;
+        return `Expected ${min} argument${min === 1 ? "" : "s"}, but received ${received}.`;
+      } else if (max === undefined) {
+        return `Expected at least ${min} argument${min === 1 ? "" : "s"}, but received ${received}.`;
       } else {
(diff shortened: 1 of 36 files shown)


packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -1369,15 +1462,122 @@ translator.rule(["INCAFTER"], (option, options) => {
...
+);
+
 /** {@link CompilerOptions.initAuto} */
+translator.rule(
+  ["INITAUTO"],
+  (option, options) => {
+    ensureArguments(option, 0, 1);
+    if (option.values.length === 0) {
+      // TODO ssmifi: The spec does not say which one is the default.
+      options.initAuto = {};
```
ssmifi: Our mf suggests that the default is `FULL`. Should we carry that over?

packages/language/src/preprocessor/compiler-options/codes.ts
```diff
@@ -13,16 +13,16 @@ import { Severity } from "../../language-server/types";
 import { ParametricPLICode } from "../../validation/messages/pli-codes";
 
 export const CompilerOptionsCodes = {
-  WrongParameterCount: {
+  InvalidParameterCount: {
     code: "_COOP01",
```
msujew: Question: Is there a specific reason why these are prefixed with `_`?

msujew (approved):
Looks good to me, just a question, see below.

packages/language/src/preprocessor/compiler-options/codes.ts
ssmifi: Jein. It merely indicates that these are internal codes for now. The codes from the compiler options are not propagated to the user at the moment (just the messages of the codes), but allow simple replacement with meaningful names for certain official codes later on, such as  `CharacterAlreadyDefined: PLICodes.Warning.IBM1108I,` in #342, in which case we maybe want to report these codes as well? 
Nonetheless, I can remove the `_` if you think it's confusing.

msujew: I would remove it for now and add a comment that these codes are for internal usage, yes :)

PR #328, opened 2025-08-12
Rewrite tokenizer

msujew:
This change introduces a high performance lexer (can still be optimized though) for our preprocessor infrastructure. In my tests, it's roughly 60% faster than the old lexer (~850ms per million lines of code now, vs ~2s in the old lexer). This also enables us to then implement an incremental lexer on top of this lexer.

Due to the new infrastructure, we're now able to remove a bunch of old Chevrotain code.

Note that I didn't add any new tests for anything I wrote - everything should work just as it did previously.


packages/language/src/parser/tokenizer.ts
```diff
@@ -0,0 +1,569 @@
...
+      );
+    } else {
+      context.index++;
+      return undefined;
+    }
+  };
+}
+
+const tokenizeString = tokenizeRegex(tokens.STRING_TERM, stringRegex);
+const tokenizeNumber = tokenizeRegex(tokens.NUMBER, numberRegex);
```
ssmifi, line 193: _Minor_: Can we not just use the pattern from the token here?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -936,25 +938,39 @@ function resolveIncludeFileUri(
...
+      //     const ext = matches[0].match(/\.\w+$/);
+      //     if (
+      //       ext &&
+      //       ext?.length &&
+      //       pgroup["include-extensions"]?.includes(ext[0].toLowerCase())
+      //     ) {
+      //       return URI.file(matches[0]);
+      //     }
+      //   }
+      // }
```
ssmifi: Can we remove this?

ssmifi (approved):
Looks nice and clean. Thank you.

packages/language/src/preprocessor/instruction-interpreter.ts
msujew: Yes, I've added a few more comments to the new code and removed the commented out code.

PR #329, opened 2025-08-13
Increased Hover Support

### LINKED ISSUE #206: Hover-over functionality
Currently we are missing hover-over functionality. When I do mouse-hover on any 
- simple   variable
- construct variable  (e.g. If I hover over at level 5, it should show up-to 1 level)
- procedures
- copybooks 

It should show the content of definition with  as like in editor.

montymxb:
Closes #206, by adding in additional support for handling hover requests:

- Hover support for include items (ex. `%INCLUDE abc;` or `%INCLUDE "myfile.pli";`), which will now show the full path of the the include on hover
- Hover support for label references, which are then used to generate a hover value for the associated procedure:
```pli
 MyProc: PROC OPTIONS(MAIN) RECURSIVE;
 ...
 END MyProc; // <-- hover here
 ...
 CALL MyProc; // <-- or here
```
On either label ref, we should see the full procedure declaration there.

There are still additional options that need handling, which I've added a little stub for as well. It's just a matter of extending later on.

- Improved hover support for regular declarations (ex. `DCL MyVar FIXED DECIMAL(2,5);`). These now show the full declaration (sans the decimal params at this time). Looks like before it was mixed up a bit with structure members, giving results like `DCL 1 MyVar`, so that's been cleaned up too
- Structure members were already supported for generating hover values
- Added 4slash tests for each case

### DIFF
--- a/packages/language/src/language-server/hover-request.ts
+++ b/packages/language/src/language-server/hover-request.ts
@@ -20,11 +20,17 @@ import {
 import { Token } from "../parser/tokens";
 import { getAttributes } from "../preprocessor/util";
 import {
+  Bound,
   DeclaredVariable,
+  DimensionBound,
+  Dimensions,
+  Expression,
   IncludeItem,
   LabelPrefix,
+  ProcedureStatement,
   SyntaxKind,
   SyntaxNode,
+  Wildcard,
 } from "../syntax-tree/ast";
 import { formatPliCodeBlock } from "../utils/code-block";
 import { binaryTokenSearch } from "../utils/search";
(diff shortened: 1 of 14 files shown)


packages/language/src/language-server/hover-request.ts
```diff
@@ -59,29 +59,102 @@ function getDeclaredVariableRepresentation(
...
+}
 
-  return formatPliCodeBlock(`DCL ${elements.join(", ")};`);
+/**
+ * Converts an IncludeItem node to a string representation.
+ * Ex. %INCLUDE "file:///path/to/file.pli"
+ */
+function getIncludeItemRepresentation(node: IncludeItem): string {
+  const filePath = node.filePath ?? "<unknown>";
+  return formatPliCodeBlock(`%INCLUDE "${filePath}"`);
```
msujew: Suggestion: Showing the full URI here (including the schema) is probably rather unexpected UX. Can we use `URI.parse(filePath).fsPath` instead? Also, if the include does not resolve, we should probably not show a hover, since the preprocessor will raise an error at that location anyway.

packages/language/src/language-server/hover-request.ts
```diff
@@ -59,29 +59,102 @@ function getDeclaredVariableRepresentation(
...
+      if (option.order) {
+        optionStrings.push(option.order);
+      }
+    } else if (option.kind === SyntaxKind.ProcedureRecursiveOption) {
+      optionStrings.push("RECURSIVE");
+    }
+    // TODO @montymxb add additional procedure cases cases as they come up
+  }
+  const options = optionStrings.length > 0 ? optionStrings.join(" ") : "";
+  return formatPliCodeBlock(`${name}: PROC ${options};`);
```
msujew: Suggestion: IMO we should show procedure parameters as well here.

packages/language/src/language-server/hover-request.ts
montymxb: Ah good catch! Yep that matches what we would be seeing in TS and such. I'm still missing additional content too, apparently the cobol extension attempts to inline the entire file's contents on hover as well. Up to about 15k lines or so from a rough test, but looks like we're going to opt for 15 - 20 lines or so.

packages/language/src/language-server/hover-request.ts
montymxb: I can add that in too.

montymxb:
Windows fail noted, will come back to this in a bit.

packages/language/src/preprocessor/instruction-generator.ts
```diff
@@ -234,7 +234,7 @@ function getAllDeclarations(
   return declarations;
 }
 
-function getAttributes(item: ast.DeclaredVariable): string[] {
+export function getAttributes(item: ast.DeclaredVariable): string[] {
```
msujew: Suggestion: If you need this information somewhere else, please move this out of this file and into some utils. The hover provider should be unaware of the `instruction-generator.ts`

packages/language/src/parser/parser.ts
```diff
@@ -658,6 +662,18 @@ export class PliParser extends AbstractParser {
...
+      const endRange = tokenToRange(token);
+      if (token.uri && startRange) {
+        const textDoc = TextDocuments.get(token.uri);
+        if (textDoc) {
+          const text = textDoc.getText({
+            start: textDoc.positionAt(startRange.start),
+            end: textDoc.positionAt(endRange.end),
+          });
+          element.sourceText = text;
+        }
```
msujew: While this works, isn't this a bit problematic in the presence of preprocessor code? I.e. you can write something like this:

```
%DCL X INIT("HELLO_WORLD: PROC; END HELLO_WORLD;");
X
```

And the hover provider will simply show nothing, since neither start nor end is defined for the tokens. So every reference to `HELLO_WORLD` will have no hover information (which exists in the AST though).

montymxb: Yeah, that's a good point. Hmm, I steered away from depending on the AST initially due to complexity involved in effectively printing out far more than I wanted to, and completely overlooked the impact of the PP.

I'll go back over this, I already had code that was serializing pretty much everything from the sub-tree of the procedure statement, seems that'll be necessary after all.

montymxb:
Went ahead & extended the hover test api to support testing includes separately, to handle the OS-dependent path aspect of the include tests, specifically to handle the fsPath side-effect.

msujew (approved):
Thanks, looks good to me 👍

PR #330, opened 2025-08-13
Add caching for tokenizer + preprocessor

msujew:
A small change that adds a cache for tokenizer and preprocessor results to the lexing infrastructure: As long as the input text (and the important compiler options don't change), the whole tokenization + parsing result for any file (i.e. tokens, parser errors, statements and preprocessor instruction) can be cached. Note that the result of the preprocessor **is not cached**. It is still recomputed on every change since the result of the preprocessor for an individual file is not independent of other files.

This seems to have a rather profound impact on the lexing time :) Tests should still pass just like before.

### DIFF
--- a/packages/language/src/preprocessor/instruction-cache.ts
+++ b/packages/language/src/preprocessor/instruction-cache.ts
@@ -37,20 +37,12 @@ export class InstructionCache {
   private previousCompilerOptions = getDefaultCompilerOptions();
 
   update(compilerOptions: CompilerOptions): void {
-    let update = false;
-    if (compilerOptions.or !== this.previousCompilerOptions.or) {
-      update = true;
-    }
-    if (compilerOptions.not !== this.previousCompilerOptions.not) {
-      update = true;
-    }
     if (
+      compilerOptions.or !== this.previousCompilerOptions.or ||
+      compilerOptions.not !== this.previousCompilerOptions.not ||
       compilerOptions.pp?.ppInclude?.value !==
-      this.previousCompilerOptions.pp?.ppInclude?.value
+        this.previousCompilerOptions.pp?.ppInclude?.value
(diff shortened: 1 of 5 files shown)


packages/language/src/preprocessor/instruction-cache.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (compilerOptions.not !== this.previousCompilerOptions.not) {
+      update = true;
+    }
+    if (
+      compilerOptions.pp?.ppInclude?.value !==
+      this.previousCompilerOptions.pp?.ppInclude?.value
+    ) {
+      update = true;
+    }
+    if (update) {
```
ssmifi: _Minor_: This can be simplified with short-circuit or `else`s.

ssmifi (commented):
Code looks good. 

_Question_: I am still wondering when the cache is actually applied. In the end, the tokenizer is called at `onDidChangeContent` or if the plugin config file changes, right? While possible to send a change content request without change, I'd have thought that this is a rare event, isn't it?

msujew:
> I am still wondering when the cache is actually applied. In the end, the tokenizer is called at onDidChangeContent or if the plugin config file changes, right? While possible to send a change content request without change, I'd have thought that this is a rare event, isn't it?

If you have a compilation unit at hand with multiples `%INCLUDE` instruction, only one file actually changes at a time. All the other files stay the same (assuming the compiler options don't change). The cache is invalidated for the changed file, while the other files haven't changed and can have their whole tokenization phase cached.

ssmifi (approved):
Ok, I see. Thank you for the clarification. 
The loading issues I experienced were unrelated and are addressed in https://github.com/zowe/zowe-pli-language-support/pull/337.

PR #331, opened 2025-08-13
Remove Langium from dev workflow

msujew:
Our current build workflow to generate the textmate syntax file requires us to use the old Langium grammar file to get access to all the keywords in the language. As of #328, this is no longer necessary, since we have all keywords available in a simple list.

This change removes all references to the `langium` package in our repo and uses the new keyword list to generate the textmate file on the fly. The file was previously checked into git, which also has been changed, as it is a purely programmatically generated artifact. Also adjusts the manual part of the textmate syntax to better reflect our syntax.


scripts/generate-tmlanguage.mts
```diff
@@ -54,18 +38,24 @@ const controlKeywords = [
...
     'xinscan',
     'deactivate',
+    'deact',
     'activate',
-];
+    'act',
+].sort();
 
-const storageKeywords = keywords.map(e => e.toLowerCase()).exclude(controlKeywords).toArray();
-storageKeywords.push('scan', 'rescan', 'noscan', 'norescan', 'activate', 'act', 'deactivate', 'deact');
```
montymxb, line 61: What about the rest of these xxscan keywords? Are these okay to drop of do we still need them?

montymxb (approved):
:octocat: looks good! Left one point about the dropped scan keywords. Feel free to proceed if that's a non-issue.

scripts/generate-tmlanguage.mts
msujew: They are now part of the keyword list in the `keywordMap`. We only needed them here, because they didn't exist in the original Langium grammar.

montymxb: Got it. Then we can probably  remove `act` and `deact` as well, they're also in the map.

msujew: No, that's working as intended. We differentiate between `control` and `storage` keywords. `act` and `deact` are control keywords, so they get special handling in the list. Everything else is just `storage`.

The original code was pretty nonsensical.

montymxb: Ahhh I see, thanks for clarifying, that makes more sense.

PR #334, opened 2025-08-14
Normalize PP Include option

msujew:
We're currently normalizing the include option to always be lower cased. This works if the user actually writes a lower case version of the PP include option, but not if they use an upper case version. This change will always normalize both the option as well as the input text to both be upper cased.

### DIFF
--- a/packages/language/src/parser/tokenizer.ts
+++ b/packages/language/src/parser/tokenizer.ts
@@ -410,7 +410,12 @@ function tokenizeIncludeAlt(
     return undefined;
   }
   for (let i = 0; i < includeAlt.length; i++) {
-    if (context.input[context.index + i] !== includeAlt[i]) {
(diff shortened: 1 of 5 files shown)


montymxb (approved):
:octocat: Looks good!

Separate point. Initially a bit odd when I looked farther up to see where this is being invoked in the new tokenizer. Feels strange that it's squeezed in before all the other tokenizing functions. Not something to change now, it works just fine atml, but I'm concerned we'll have other strange cases that may necessitate a similar approach, and it would be good to pattern that out a bit more.

msujew:
> but I'm concerned we'll have other strange cases that may necessitate a similar approach, and it would be good to pattern that out a bit more.

I'm open to that in case we find more of these strange tokens that can be configured by the user.

PR #336, opened 2025-08-14
Make compilation unit lifecycle async

msujew:
There are two main reasons to make the lifecycle async:
1. With a sync lifecycle, we always need to wait until the previous lifecycle is completed until we can start with a new one. This can quickly accumulate and essentially prevent the language server from catching up to the current state of the document.
2. Eventually, we want to perform async file operations in the preprocessor code (i.e. lookups of files on remote filesystems, etc.). This requires us to make the preprocessor async, which in turn requires the lifecycle to already be async.

This fixes the first issue, as it will abort the mutex when the compilation unit is changed. The second part will come when necessary. Most of the changes in here are related to the tests, which all needed to be updated to be async.

### DIFF
--- a/packages/language/src/utils/promises.ts
+++ b/packages/language/src/utils/promises.ts
@@ -1,0 +1,121 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import {
+  CancellationToken,
+  CancellationTokenSource,
+} from "vscode-languageserver";
+
+export type MaybePromise<T> = T | Promise<T>;
+
+/**
+ * Delays the execution of the current code to the next tick of the event loop.
+ * Don't call this method directly in a tight loop to prevent too many promises from being created.
+ */
+export function delayNextTick(): Promise<void> {
+  return new Promise((resolve) => {
+    // In case we are running in a non-node environment, `setImmediate` isn't available.
+    // Using `setTimeout` of the browser API accomplishes the same result.
+    if (typeof setImmediate === "undefined") {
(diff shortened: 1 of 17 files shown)


packages/language/test/validating.test.ts
```diff
@@ -390,7 +390,7 @@ describe("Validating", () => {
   //     });
   //
   //     test('check capital letter validation', async () => {
-  //         document = await parseWithValidations(`
+  //         document = await await parseWithValidations(`
   //             person langium
   //         `);
   //
```
ssmifi: _Minor_: Can we remove these?

packages/language/src/workspace/lifecycle.ts
```diff
@@ -25,19 +25,28 @@ import { LexerResult, PliLexer } from "../preprocessor/pli-lexer";
...
   tokenize(compilationUnit, text);
+  await interruptAndCheck(cancellation);
   parse(compilationUnit);
+  await interruptAndCheck(cancellation);
   generateSymbolTable(compilationUnit);
+  await interruptAndCheck(cancellation);
   link(compilationUnit);
+  await interruptAndCheck(cancellation);
   validate(compilationUnit);
+  await interruptAndCheck(cancellation);
```
ssmifi, line 49: _Question_: I guess it is intended to abort even after the last step to prevent the diagnostics to be send to the client, because the only way for a cancellation is another run that would send the potentially new diagnostics anyway, correct?

packages/language/src/utils/promises.ts
```diff
@@ -0,0 +1,121 @@
...
+  token: CancellationToken,
+): Promise<void> {
+  if (token === CancellationToken.None) {
+    // Early exit in case cancellation was disabled by the caller
+    return;
+  }
+  const current = performance.now();
+  if (current - lastTick >= globalInterruptionPeriod) {
+    lastTick = current;
+    await delayNextTick();
```
ssmifi, line 92: _Question_: Could we return early here and not wait if cancellation is already requested?

ssmifi (approved):
Looks good. Thanks for the 1on1!

PR #337, opened 2025-08-14
Fix uri in case sensitive fs

ssmifi:
Results from the debugging session with @msujew (thanks again):
Fixed uri in case sensitive fs and removed the deletion of the compilation unit on close.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -938,22 +938,22 @@ function resolveIncludeFileUri(
         lib,
         item.fileName,
       );
-      const files: [string, URI][] = [[libFileUri.path, libFileUri]];
+      const files: string[] = [libFileUri.path];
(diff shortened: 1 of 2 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -949,11 +949,11 @@ function resolveIncludeFileUri(
       // Check whether any of the glob patterns match a file in the file system
       // We cannot check directly whether a file exists, as PL/I handles the file system as case insensitive
       // whereas Unix file systems are case sensitive
-      for (const [filePath, fileUri] of files) {
+      for (const [filePath, _] of files) {
```
msujew: I don't think we need the `URI` part at all anymore, right? We can just make it `const files: string[]`.

ssmifi: True. Thx.

msujew (approved):

Issue #335, opened 2025-08-14
Findings - 14.8


asatklichov:
- [ ] 1. Validation for duplication in compiler options should be improved
Currently it complains about: e.g. INCAFTER(PROCESS(XYZ)) and INCAFTER(PROCESS(ABC))  or  FLAG(W) and  FLAG(S)

- [ ] 2. Below issue is existing in Playground as well

<img width="1364" height="360" alt="Image" src="https://github.com/user-attachments/assets/f5187317-d7ff-4e45-929c-62f11f12daa3" />

- [x] 3. Margin issue, see[ 186 ](https://github.com/zowe/zowe-pli-language-support/issues/186)how it should behave 

<img width="389" height="240" alt="Image" src="https://github.com/user-attachments/assets/9280cff0-4386-4c2c-bda5-e9b09ddf05b3" />

- [x] 4. Nested includes has issues with loading directly

<img width="300" height="115" alt="Image" src="https://github.com/user-attachments/assets/f6684005-8c0e-4a8c-9532-2fd09b93a870" />

- [ ] 5. Provide better error message for compiler option %PROCESS OR('|'), NOT ('¬') ; 
See  [304](https://github.com/zowe/zowe-pli-language-support/issues/304)

<img width="413" height="245" alt="Image" src="https://github.com/user-attachments/assets/c87e8260-ae98-4d29-8977-9a5a5bf2d777" />

- [x] 6. When including files starting with number, there is a lexer issue 
%INCLUDE 25k;

<img width="1507" height="509" alt="Image" src="https://github.com/user-attachments/assets/8cdf17a1-c9af-4ba0-87a0-e39604c66e11" />

montymxb:
Marking 6 as resolved due to being a likely non-issue. This was double checked and found to generates an **IBM3841I S** error.

Issue #301, opened 2025-07-28
[Discrepancy] Semicolon in process directive is mandatory


ssmifi:
The semicolon in the process directive, e.g., `*PROCESS MARGINS(2,72);`, (at least on our mainframe) is mandatory. Otherwise, the following statement, even if on the next line, is counted towards the process directive and is not recognized as valid option.

ssmifi:
Current options parser also expects a semicolon.

PR #338, opened 2025-08-15
Fix margin warnings at CR

ssmifi:
Fix for https://github.com/zowe/zowe-pli-language-support/issues/335 3

- Empty lines with CRLF are now ignored.
- Adjusted the process directive extraction to be in line with the mainframe. The semicolon is not yet mandatory in our version though (https://github.com/zowe/zowe-pli-language-support/issues/301).
- Extended the harness to test for \n \r \t.

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/parser.ts
+++ b/packages/language/src/preprocessor/compiler-options/parser.ts
@@ -29,6 +29,7 @@ import {
 import { Token } from "../../parser/tokens";
 
 const commaToken = createToken({ name: "comma", pattern: "," });
+const semicolonToken = createToken({ name: "semicolon", pattern: ";" });
 const stringToken = createToken({
   name: "string",
   pattern: /("(""|\\.|[^"\\])*"|'(''|\\.|[^'\\])*')/,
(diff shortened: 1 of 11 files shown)


packages/language/src/preprocessor/compiler-options/parser.ts
```diff
@@ -261,9 +261,10 @@ export function parseAbstractCompilerOptions(
   input: string,
   offset?: number,
 ): AbstractCompilerOptions {
-  const lexerResult = lexer.tokenize(
-    " ".repeat(offset ?? 0) + input.replace(/;$/, ""),
-  );
+  // Remove everything after the first ;.
+  // *PROCESS MARGINS(2, 72) ; MARGINS(1, 72); is valid, but everything after the first ; is ignored.
+  const validInput = input.split(";")[0];
+  const lexerResult = lexer.tokenize(" ".repeat(offset ?? 0) + validInput);
```
msujew: Suggestion: This breaks in the presence of `;` in a string: `*PROCESS OR(";");`. Perhaps it makes sense to add a `;` to the tokenizer, tokenize the input text and slice the tokens at the index of the `;` token?

msujew (commented):
Looks quite good, just one comment, see below.

packages/language/src/preprocessor/compiler-options/parser.ts
ssmifi: Makes sense. Updated the lexer accordingly. Thanks!

msujew (approved):
Thanks!

PR #340, opened 2025-08-18
Update playground link for `main`

msujew:
Just so we get the playground link on main before our next merge (sometime in September).


VitGottwald (approved):
Thank you @msujew !

PR #341, opened 2025-08-18
Complete support for preprocessor arrays

msujew:
This is a follow up on https://github.com/zowe/zowe-pli-language-support/pull/261, which added preliminary support for arrays in the preprocessor. This finishes this work:

The PR adds support for array values as return types of any computation. Adopters can now assign arrays to other variables.

In the other PR, I've mentioned that we want to support "array slices" (array cross sections in the language reference) as well eventually. I've taken this from the language reference for wildcard expressions in PLI code:

image

However, while this is supported for PLI code in the compiler, the preprocessor does not allow to use cross sections. So they are not supported in our preprocessor implementation.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -31,26 +31,26 @@ import { tokenize } from "../parser/tokenizer";
 
 interface Variable {
   name: string;
-  value: VariableValue;
+  value: Value;
   declarationNode?: ast.SyntaxNode | null;
   mode: inst.ScanMode;
   active: boolean;
 }
 
-type VariableValue = Value | ArrayValue;
+type Value = ScalarValue | ArrayValue;
 
-interface Value {
-  // Preprocessor variables are always strings or numbers.
+interface ScalarValue {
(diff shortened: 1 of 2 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -397,6 +386,10 @@ function runIfInstruction(
   context: InterpreterContext,
 ): inst.InstructionNode | undefined {
   const condition = evaluateExpression(instruction.condition, context);
+  if (!isScalarValue(condition)) {
+    // Condition cannot be evaluated, returning false branch
+    return instruction.falseBranch;
```
ssmifi: _Question_: Should we return any branch at all if the condition cannot be evaluated?

ssmifi (commented):
LGTM, just one question about the branching.

packages/language/src/preprocessor/instruction-interpreter.ts
msujew: Good point, I'll change it to return `undefined`.

ssmifi (approved):
Thanks!

PR #348, opened 2025-08-19
Fixes some issues with the scripts

danieldietrich:


scripts/generate-railroad-diagram.mts
```diff
@@ -10,7 +10,7 @@
  */
 
 import { PliParserInstance } from '../packages/language/src/parser/parser';
-import { createSyntaxDiagramsCode } from 'chevrotain';
+import { createSyntaxDiagramsCode } from '../packages/language/node_modules/chevrotain';
```
danieldietrich: I left the chevrotain dependencies where they currently are. The script is just a script - this will work fine.

scripts/generate-tmlanguage.mts
```diff
@@ -12,7 +12,7 @@
 import * as fs from 'fs/promises';
 import * as tokens from '../packages/language/src/parser/tokens.js';
 
-const keywords = Array.from(tokens.keywordMap, ([key]) => key.toLowerCase()).toSorted();
+const keywords = Array.from(tokens.keywordMap, ([key]) => key.toLowerCase()).sort();
```
danieldietrich, line 15: toSorted() is supported by ES2023+ (which doesn't seem to be used).

Difference: sort() mutates the array in place but Array.from() does already create a new array instance.

scripts/generate-railroad-diagram.mts
danieldietrich: @msujew please wait, there still seems to be a problem. I will convert this PR to draft meanwhile...

danieldietrich: Done!

scripts/generate-railroad-diagram.mts
```diff
@@ -10,7 +10,7 @@
  */
 
 import { PliParserInstance } from '../packages/language/src/parser/parser';
-import { createSyntaxDiagramsCode } from 'chevrotain';
+import { createSyntaxDiagramsCode } from '../packages/language/node_modules/chevrotain/lib/src/api.js';
```
danieldietrich: Fixed missing dependency

msujew: Suggestion: IMO we should add `chevrotain` as a dev dependency to the root project.

danieldietrich: Ok, let me change that...

msujew (approved):

Issue #297, opened 2025-07-25
Preprocessor Implementation Completion  - Core Language


asatklichov:
Currently we have support for  the following preprocessor statements:

Listing control statements (PAGE, SKIP, PRINT, PUSH, POP)
Standard include statements (%INCLUDE, %XINCLUDE, %INSCAN, %INSCAN)
Core preprocessor statements (%DECLARE, %ACTIVATE, %DEACTIVATE, %DO, %SELECT, %IF, %ELSE, %THEN, %GOTO, %END, %ITERATE, %LEAVE, %NOTE, %null, %REPLACE)

And we want to implement for: 
- MACRO preprocessor statement builtins 
- Procedures preprocessor procedures -   %PROCEDURE
- And missing %REPLACE, %TRACE, %USE  etc. 

**Also related bugs** 
--------------- ----
[OPEN] Macro %DCL / %DECARE is not supporting dimensions, attributes correctly,
see language reference page 618

PR #350, opened 2025-08-20
Initial support for preprocessor procedures

msujew:
Related to https://github.com/zowe/zowe-pli-language-support/issues/297. Adds support for preprocessor procedures in preprocessor code. I.e. you can now call a (locally) defined preprocessor procedure:

```
%TEST_PROC: PROC (...) RETURNS (CHAR);
  // DO SOMETHING
%END;
%VAR = TEST_PROC(...);
```

And expect that the correct value is now attached to `VAR`.

Out of scope for this PR: Use preprocessor procedures outside of preprocessor code. This requires some tricky behavior, due to the `STATEMENT` keyword:

grafik

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -19,7 +19,10 @@ import {
 import { FileSystemProviderInstance } from "../workspace/file-system-provider";
 import { PluginConfigurationProviderInstance } from "../workspace/plugin-configuration-provider";
 import { CompilerOptionResult } from "./compiler-options/options";
-import { generateInstructions } from "./instruction-generator";
+import {
+  generateInstructions,
+  InstructionGeneratorResult,
+} from "./instruction-generator";
 import * as inst from "./instructions";
 import * as ast from "../syntax-tree/ast";
 import { LexingIssue } from "./pli-lexer";
@@ -165,16 +168,26 @@ interface InterpreterContext {
   unit: CompilationUnit;
   currentUri?: URI;
   entryUri?: URI;
+  /**
+   * These are global variables that are defined on the root level of the preprocessor
+   */
   variables: Map<string, Variable>;
+  /**
+   * Local variables only exist within the scope of a procedure. Set to `undefined` when not in a procedure.
+   */
+  localVariables?: Map<string, Variable>;
   statements: ast.Statement[];
   result: CompilationUnitTokens;
   errors: LexingIssue[];
+  procedures: Map<string, inst.ProcedureInstructionContainer>;
+  activeProcedures: Set<string>;
   counter: Map<inst.InstructionNode, number>;
   references: ast.Reference[];
   evaluations: EvaluationResults;
   xIncludes: Set<string>;
   uris: string[];
   options: InterpreterOptions;
+  returnValue: Value;
 }
 
 export type InstructionInterpreterResult = CompilationUnitTokens & {
@@ -195,7 +208,7 @@ export interface InterpreterOptions {
 export function runInstructions(
   unit: CompilationUnit,
   uri: URI,
-  start: inst.InstructionNode,
+  instruction: InstructionGeneratorResult,
   options: InterpreterOptions,
 ): InstructionInterpreterResult {
   const context: InterpreterContext = {
@@ -207,6 +220,8 @@ export function runInstructions(
     statements: [],
     xIncludes: new Set(),
     variables: new Map(),
+    procedures: new Map(),
+    activeProcedures: new Set(),
     references: [],
     evaluations: {
       ifStatements: new Map(),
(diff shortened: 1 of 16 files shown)


packages/language/src/preprocessor/pli-preprocessor-parser-state.ts
```diff
@@ -18,7 +18,10 @@ import { URI } from "../utils/uri";
 import { Token } from "../parser/tokens";
 import { tokenize } from "../parser/tokenizer";
 
-type ParserLocation = "in-statement" | "in-procedure";
+export enum ParserLocation {
+  Statement,
```
ssmifi: _Question_: Is the `Statement` information in `Location` used anywhere now that `isOnlyInStatement` relies on `Procedure` or will it be needed in the future?

packages/language/src/preprocessor/pli-preprocessor-parser.ts
```diff
@@ -364,6 +349,32 @@ export class PliPreprocessorParser {
     return statement;
```
ssmifi, line 371: This `state.pop();` is already handled by the `commonStatement` caller, correct?

_Suggestion_: If so, add a test case that catches this behavior.

packages/language/src/preprocessor/pli-preprocessor-parser-state.ts
```diff
@@ -70,7 +71,7 @@ export class PliPreprocessorParserState implements PreprocessorParserState {
   public uri: URI;
   private text: string;
   private location: ParserLocation[] = [];
```
ssmifi: _Question_: Do you need the stack for nesting? 
According to spec, preprocessor procedures cannot be nested (p. 601, lrm). If this is really the case (and `ParserLocation.Statement` is not required elsewhere), we should prohibit nesting and the `inProcedure` logic could probably be simplified.

ssmifi (commented):
Code looks good. However, I am unsure about the nesting of preprocessor procedures (see questions.) Although there seems to be an issue with the scoping(?), the parser currently recognizes nested preprocessor procedures, correct? According to spec (p. 601, lrm), this should not be valid.

msujew:
@ssmifi Thanks, I didn't look too closely at the list of supported statements in a preprocessor procedure. I've adapted the parser to support the correct ones and removed all code for nested procedures (since the parser rejects those anyway now).

I've also added a few more tests that helped me identify another issue in the parsing code related to `END` statements.

packages/language/src/preprocessor/pli-preprocessor-parser-state.ts
msujew: I've removed it, it was no longer required.

packages/language/src/preprocessor/pli-preprocessor-parser.ts
msujew: Thanks, fixed this.

packages/language/src/preprocessor/pli-preprocessor-parser-state.ts
msujew: I've removed it, due to the lack of procedure nesting.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -682,29 +771,44 @@ function runDeclareInstruction(
...
-  context.variables.set(variable.name, variable);
+  // If a declaration like `DCL PROC_NAME ENTRY` appears, simply activate the procedure
+  if (context.procedures.has(instruction.name)) {
+    context.activeProcedures.add(instruction.name);
+    return;
+  }
+  // If the local variables are defined, that means we are in a procedure
+  const variables = context.localVariables ?? context.variables;
+  // Don't override existing variables
+  if (!variables.has(instruction.name)) {
```
ssmifi, line 782: _Question_: Couldn't there be variables in nested scopes that shadow existing local variables? Is this a thing in the pp?

ssmifi (commented):
Thanks, looks good. I have one remaining question about the shadowing of local variables.

packages/language/src/preprocessor/instruction-interpreter.ts
msujew: Can you give an example? I have a hard time imagining how this would look like.

ssmifi: I was thinking about sth like 
```
 %TEST: PROC (VAR) RETURNS (FIXED);
   DCL X FIXED;
   X = 1;
   DO;
     DCL X FIXED;
     X = 2;
   END;
   RETURN (X);
 %END;
```
but I quickly tried that on the normal parser and it is also not valid, so I guess that's not a thing in pli. Sorry for the delay. :)

ssmifi (approved):

packages/language/src/preprocessor/instruction-interpreter.ts
msujew: I see. Yes, PLI does not see `DO`, `IF`, etc. as a new scope. Only entering a procedure qualifies as a new scope. We already handle that correctly both in the interpreter and in the validation 👍

PR #343, opened 2025-08-19
Adds Do-Type 3 preprocessor parser support

danieldietrich:
Fixes #315

This PR implements support for PL/I preprocessor DO Type 3 statements with comprehensive loop control variable parsing. See `code_samples/preprocessor/do-type3.pli` for examples.

Additions:

- Added a script to generate the GBNF file `pli.gbnf`. Run `./scripts/generate-gbnf.mts | less # or > pli.gbnf`
- Added a script to run the preprocessor parser on pli files. Run `./scripts/pre-parse.mts code_samples/preprocessor/do-type3.pli | less # or > do-type3.ast`
- Minor: Made `scripts/*.mts` executable

---

## DO Type 3 Examples - Code to AST Mapping

_(please ignore the start and end offsets, all examples reside in the same file)_

**Basic DO i = 1 TO 10**
<details>

Code:
```pli
%DCL I FIXED;
%DO I = 1 TO 10;
  PUT LIST('Iteration: ' || I);
%END;
```

AST:
```json
{
  "kind": "Statement",
  "value": {
    "kind": "DoStatement",
    "statements": [
      {
        "kind": "Statement",
        "value": {
          "kind": "TokenStatement",
          "tokens": [
            {
              "image": "PUT",
              "startOffset": 110,
              "endOffset": 112
            },
            {
              "image": "LIST",
              "startOffset": 114,
              "endOffset": 117
            },
            {
              "image": "(",
              "startOffset": 118,
              "endOffset": 118
            },
            {
              "image": "'Iteration: '",
              "startOffset": 119,
              "endOffset": 131
            },
            {
              "image": "||",
              "startOffset": 133,
              "endOffset": 134
            },
            {
              "image": "I",
              "startOffset": 136,
              "endOffset": 136
            },
            {
              "image": ")",
              "startOffset": 137,
              "endOffset": 137
            },
            {
              "image": ";",
              "startOffset": 138,
              "endOffset": 138
            }
          ]
        }
      }
    ],
    "end": {
      "kind": "EndStatement",
      "label": null
    },
    "doType2": null,
    "doType3": {
      "kind": "DoType3",
      "variable": {
        "kind": "ReferenceItem",
        "ref": {
          "text": "I",
          "preprocessor": true
        }
      },
      "specifications": [
        {
          "kind": "DoSpecification",
          "expression": {
            "kind": "Literal",
            "value": {
              "kind": "NumberLiteral",
              "value": "1"
            }
          },
          "upthru": null,
          "downthru": null,
[...]

PR #351, opened 2025-08-21
Adds Do-Type 3 support to preprocessor interpreter

### LINKED ISSUE #316: [Feature] Support Do-Type 3 statements in preprocessor interpreter
In #315 we want to support the parsing of the statement. This issue is for tracking the progress on the interpreter side.

danieldietrich:
Fixes #316 
Follow-up of PR #343

This PR implements interpreter support for PL/I preprocessor DO Type 3 statements. The semantics are described in the [PL/I language specification](https://www.ibm.com/docs/en/SSY2V3_6.1/pdf/lrm.pdf) on the pages 212-222 ("DO statement").

The corresponding tests can be run by

```bash
npx vitest run packages/language/test/fourslash-harness/execute.test.ts -t "do-type3-"
```

### DIFF
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -392,10 +392,12 @@ function generateDoInstruction(
     };
   }
   if (node.doType3) {
-    if (!node.doType3.variable) {
+    if (!node.doType3.variable?.element) {
       return undefined;
     }
-    const variable = generateReferenceItemInstruction(node.doType3.variable);
+    const variable = generateReferenceItemInstruction(
+      node.doType3.variable.element,
+    );
     if (node.doType3.specifications.length !== 1) {
       return undefined; // Preprocessor %DO does require exactly one specification
     }
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -392,7 +392,10 @@ function generateDoInstruction(
     };
   }
   if (node.doType3) {
-    const variable = generateReferenceItemInstruction(node.doType3.variable!);
+    if (!node.doType3.variable) {
+      return undefined;
+    }
+    const variable = generateReferenceItemInstruction(node.doType3.variable);
     if (node.doType3.specifications.length !== 1) {
       return undefined; // Preprocessor %DO does require exactly one specification
     }
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -391,6 +391,31 @@ function generateDoInstruction(
       while: whileCond,
     };
   }
+  if (node.doType3) {
+    const variable = generateReferenceItemInstruction(node.doType3.variable!);
+    const specificationItems: inst.DoType3SpecificationItem[] = [];
+    for (const spec of node.doType3.specifications) {
+      const specItem: inst.DoType3SpecificationItem = {
+        expression: generateExpressionInstruction(spec.expression) ?? null,
+        upthru: generateExpressionInstruction(spec.upthru) ?? null,
+        downthru: generateExpressionInstruction(spec.downthru) ?? null,
+        repeat: generateExpressionInstruction(spec.repeat) ?? null,
+        while: spec.whileOrUntil?.while
+          ? (generateExpressionInstruction(spec.whileOrUntil.while) ?? null)
+          : null,
+        until: spec.whileOrUntil?.until
+          ? (generateExpressionInstruction(spec.whileOrUntil.until) ?? null)
+          : null,
+        to: generateExpressionInstruction(spec.to) ?? null,
+        by: generateExpressionInstruction(spec.by) ?? null,
+      };
+      specificationItems.push(specItem);
+    }
+    doInstruction.doType3 = {
+      variable,
+      specificationItems,
+    };
+  }
   if (node.doType2 || node.doType3 || node.doType4) {
     doType1 = false;
   }
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -393,27 +393,24 @@ function generateDoInstruction(
   }
   if (node.doType3) {
     const variable = generateReferenceItemInstruction(node.doType3.variable!);
-    const specificationItems: inst.DoType3SpecificationItem[] = [];
-    for (const spec of node.doType3.specifications) {
-      const specItem: inst.DoType3SpecificationItem = {
-        expression: generateExpressionInstruction(spec.expression) ?? null,
-        upthru: generateExpressionInstruction(spec.upthru) ?? null,
-        downthru: generateExpressionInstruction(spec.downthru) ?? null,
-        repeat: generateExpressionInstruction(spec.repeat) ?? null,
-        while: spec.whileOrUntil?.while
-          ? (generateExpressionInstruction(spec.whileOrUntil.while) ?? null)
-          : null,
-        until: spec.whileOrUntil?.until
-          ? (generateExpressionInstruction(spec.whileOrUntil.until) ?? null)
-          : null,
-        to: generateExpressionInstruction(spec.to) ?? null,
-        by: generateExpressionInstruction(spec.by) ?? null,
-      };
-      specificationItems.push(specItem);
+    if (node.doType3.specifications.length !== 1) {
+      return undefined; // Preprocessor %DO does require exactly one specification
     }
+    const spec = node.doType3.specifications[0];
+    if (spec.upthru || spec.downthru) {
+      return undefined; // upthru and downthru are not supported
+    }
+    const specification: inst.DoType3Specification = {
+      expression: generateExpressionInstruction(spec.expression) ?? null,
+      repeat: generateExpressionInstruction(spec.repeat) ?? null,
+      while: generateExpressionInstruction(spec.whileOrUntil?.while) ?? null,
+      until: generateExpressionInstruction(spec.whileOrUntil?.until) ?? null,
+      to: generateExpressionInstruction(spec.to) ?? null,
+      by: generateExpressionInstruction(spec.by) ?? null,
+    };
     doInstruction.doType3 = {
       variable,
-      specificationItems,
+      specification,
     };
   }
   if (node.doType2 || node.doType3 || node.doType4) {
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -185,6 +185,8 @@ interface InterpreterContext {
   procedures: Map<string, inst.ProcedureInstructionContainer>;
   activeProcedures: Set<string>;
   counter: Map<inst.InstructionNode, number>;
+  doType3SpecIndex: Map<inst.InstructionNode, number>;
+  doType3NeedsReinit: Map<inst.InstructionNode, boolean>;
   references: ast.Reference[];
   evaluations: EvaluationResults;
   xIncludes: Set<string>;
@@ -235,6 +237,8 @@ export function runInstructions(
     },
     options,
     counter: new Map(),
+    doType3SpecIndex: new Map(),
+    doType3NeedsReinit: new Map(),
     returnValue: defaultEmptyValue,
   };
   for (const [key, value] of instruction.procedures.entries()) {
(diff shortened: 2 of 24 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -170,6 +170,8 @@ interface InterpreterContext {
   result: CompilationUnitTokens;
   errors: LexingIssue[];
   counter: Map<inst.InstructionNode, number>;
+  doType3SpecIndex: Map<inst.InstructionNode, number>;
+  doType3NeedsReinit: Map<inst.InstructionNode, boolean>;
```
msujew: Suggestion: Since we need more info on doType3 statements (see comments below) and these maps are always accessed together what do you think about something like this?

```ts
doType3: Map<inst.InstructionNode, DoType3Store>;

interface DoType3Store {
  ...
}
```

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -419,88 +425,94 @@ function runDoType3Instruction(
...
+        stepValue = evaluateExpression(spec.by, context);
+      } else {
+        stepValue = valueToNumber("1");
+      }
+    } else if (spec.upthru) {
+      endValue = evaluateExpression(spec.upthru, context);
+      stepValue = valueToNumber("1");
+    } else if (spec.downthru) {
+      endValue = evaluateExpression(spec.downthru, context);
+      stepValue = valueToNumber("-1");
```
msujew: Suggestion: The language reference says this about do specifications:

image

Note the "is evaluated at entry and saved". When entering the next iteration of the current specification, this code will re-evaluate the spec, which might lead to incorrect results. See this example:

```
X = 10;
DO I = 1 TO X;
  X += 1;
END;
```

In the compiler, X is evaluated at the start to `10`, which is saved, and therefore the loop will run 10 times. If it were to re-evaluate it at every loop, it runs infinitely (until the interpreter decides to stop the loop by itself after 5000 iterations). We need to store the spec expressions, see the `DoType3Store` above.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -319,38 +324,235 @@ function runInstruction(
 function runDoInstruction(
   instruction: inst.DoInstruction,
   context: InterpreterContext,
+  node?: inst.InstructionNode,
```
msujew: Minor: Shouldn't we make this mandatory? We always have a node available that we can pass into this function.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -319,38 +324,235 @@ function runInstruction(
...
+function runDoType3Instruction(
+  doType3: inst.DoType3Instruction,
+  context: InterpreterContext,
+  node: inst.InstructionNode,
+): boolean {
+  const { specificationItems, variable } = doType3;
+
+  if (specificationItems.length === 0) {
+    throw new Error("DoType3 requires at least one specification item!");
+  }
```
msujew: Suggestion: Please don't throw in the preprocessor unless the issue indicates a programming error in our own code. Having no `specificationItems` indicates that the user has somehow written a doType3 loop without any items, which is a user error. Simply return `false` instead, which is probably what is intended.

This applies to essentially all of the `throw` statements in this method.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -319,38 +324,235 @@ function runInstruction(
...
+
+  const iterationCount = context.counter.get(node);
+  if (iterationCount === undefined || iterationCount < 1) {
+    throw new Error("Interpreter can't determine AST node iteration count");
+  }
+  const isFirstIteration = iterationCount === 1;
+
+  // Get current spec index, starting at 0 for first iteration
+  let currentSpecIndex = context.doType3SpecIndex.get(node) ?? 0;
+  const needsReinit = context.doType3NeedsReinit.get(node) ?? false;
```
msujew: Suggestion: I don't think we need the iteration count at all here, if we set the default for `needsReinit` differently.
```suggestion
  // Get current spec index, starting at 0 for first iteration
  let currentSpecIndex = context.doType3SpecIndex.get(node) ?? 0;
  const needsReinit = context.doType3NeedsReinit.get(node) ?? true;
```

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -319,38 +324,235 @@ function runInstruction(
...
+
+  // If we've exhausted all specifications, return false to stop the loop
+  if (currentSpecIndex >= specificationItems.length) {
+    return false;
+  }
+
+  const spec = specificationItems[currentSpecIndex];
+  if (!spec.expression) {
+    throw new Error("DoType3 requires initial value!");
```
msujew: Simply return false.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -319,38 +324,235 @@ function runInstruction(
...
+
+  const varName = variable.variable;
+  let loopVar = context.variables.get(varName);
+
+  // Initialize for first iteration OR when we need to reinitialize for a new spec
+  if (isFirstIteration || needsReinit) {
+    const start = evaluateExpression(spec.expression, context);
+    if (!isScalarValue(start)) {
+      throw new Error("DoType3 initial value must be scalar!");
+    }
```
msujew: Return false. The type system will (eventually) tell the user that this is an invalid operation.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -319,38 +324,235 @@ function runInstruction(
...
+    // Clear the reinit flag and update spec index
+    context.doType3NeedsReinit.set(node, false);
+    context.doType3SpecIndex.set(node, currentSpecIndex);
+    return true; // Always continue after initialization
+  }
+
+  // Normal iteration within current spec - update the loop variable
+  if (!loopVar) {
+    throw new Error("DoType3 loop variable not found on subsequent iteration!");
+  }
```
msujew: Suggestion: If the loop variable is not found, it has been implicitly set by the user. Do not throw, but follow the logic in the assignment instruction for missing variables.

msujew (commented):
Pretty good, but there are two main issues regarding:

1. Re-evaluation of expressions on every loop
2. Errors being thrown

packages/language/src/preprocessor/instruction-interpreter.ts
danieldietrich: Good catch. There might be more specifics (e.g. regarding BY etc.), I will double-check the specs.

danieldietrich:
Note: I resolved all review comments. Please hold on before another review, I will add more tests.

packages/language/src/preprocessor/instruction-generator.ts
```diff
@@ -336,6 +336,28 @@ function generateDoInstruction(
       while: whileCond,
     };
   }
+  if (node.doType3) {
+    const variable = generateReferenceItemInstruction(node.doType3.variable!);
```
msujew: Minor: don't coerce possible null values, simply return `undefined`.
```suggestion
    if (!node.doType3.variable) {
      return undefined;
    }
    const variable = generateReferenceItemInstruction(node.doType3.variable);
```

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -319,40 +327,234 @@ function runInstruction(
...
+    // Store all expressions that must be evaluated at the start of the loop
+    condition &&= _runDoType3Initialization(doType3, context, node);
+  } else {
+    // Subsequent iterations
+    condition &&= _runDoType3UntilCheck(doType3, context);
+    condition &&= _runDoType3VariableUpdate(doType3, context, node);
+  }
+
+  // Check WHILE condition before do-group execution
+  condition &&= _runDoType3WhileCheck(doType3, context);
```
msujew: Minor: Any specific reason why these are prefixed with `_`? These functions are already private, since they are not exported.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -319,40 +327,234 @@ function runInstruction(
...
+  // Initial expression is required
+  if (!spec.expression) {
+    return false;
+  }
+  const start = evaluateExpression(spec.expression, context);
+  if (!isScalarValue(start)) {
+    return false;
+  }
+
+  let loopVar = context.variables.get(varName);
```
msujew: Suggestion: Please rebase on the latest `development`. We now have `context.localVariables` as well to capture procedure variables. If the `localVariables` are set, then the new variable should be set there instead.

danieldietrich:
The next steps will be:

* before merging this PR, I want to test all do-type3 related preprocessor code_samples on the host _(currently waiting for getting access)_
* [in a different PR] additional validation routines should be added because the preprocessor does not support multiple do-sequences, UPTHRU and DOWNTHRU.

@msujew could you please take a look at these points in a second review:

* I made changes to the interfaces you initially created in instruction.ts
* The `instruction-generator.ts` now contains guards/checks regarding unsupported do-type3 features. I thought it would be a good idea to exit early. If the design decision is to do these kinds of checks solely in the `interstuction-interpreter.ts` I would move those checks.

Side notes:

* After refactoring & cleanup, `runDoType3Instruction()` now contains only a high-level view of interpreting do-type3 statements
* In scripts/, all none .mts files are ignore now. If we decide to add other scripts, like .sh, we will need to whitelist those extension in .gitignore

Thx!

packages/language/test/fourslash/preprocessor/do/do-type3-by-to.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/// <reference path="../../framework.ts" />
+
+/**
+ * DO Type 3 - DO with BY before TO
+ */
+
+//// %DCL I FIXED;
+//// %DO I = 5 BY 5 TO 10;
+////   DCL Var%;I FIXED;
```
msujew: Note that the output of the preprocessor doesn't have to be valid output to confirm the behavior. You can also just do `I` within the loop and then use `expectTokens("5 10")`. But I'm ambiguous to that. Whatever you think serves best to test the behavior.

packages/language/test/fourslash/preprocessor/do/do-type3-to-by.ts
```diff
@@ -0,0 +1,31 @@
...
+
+//// %DCL I FIXED;
+//// %DCL X FIXED;
+//// %X = 4;
+//// %Y = 2;
+//// %DO I = 2 TO X BY Y;
+////   %X = X + 1;
+////   %Y = Y + 1;
+////   DCL Var%;I FIXED;
+//// %END;
```
msujew: Minor: Can you add a comment here to explain how this test asserts that the preprocessor follows the spec?

msujew (commented):
This looks really nice! Please rebase on the latest `development`, fix a few of the issues below, and then we should be good to go :)

packages/language/src/preprocessor/instruction-interpreter.ts
danieldietrich: The `_` prefix for private functions is a common style in JS (that I use often) but you are right, it differs from the style in this codebase. I will change that 👍

packages/language/src/preprocessor/instruction-interpreter.ts
danieldietrich: Hi @msujew, could you please take a look at the latest version of instruction-interpreter.ts of this PR?

I added two helper functions that consider local variables on get/set. However, I decided to make the changes throughout the whole file and am not sure if this is correct in all cases. Could you please verify that?

Note: I did not yet change the active state to false on implicit declarations in the do-type3 loop. I will wait for the host access to perform some tests. After that (and your additional review comments) the PR will ready to be merged.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -319,47 +361,241 @@ function runInstruction(
...
+  // Get current do-type3 state
+  const doType3Context = context.doType3.get(node);
+
+  if (!doType3Context) {
+    // Store all expressions that must be evaluated at the start of the loop
+    condition &&= runDoType3Initialization(doType3, context, node);
+  } else {
+    // Subsequent iterations
+    condition &&= runDoType3UntilCheck(doType3, context);
+    condition &&= runDoType3VariableUpdate(doType3, context, node);
```
msujew: Suggestion: Shouldn't we pass the `doType3Context` instead of the node here? Otherwise we use `doType3.get(node)` twice.
```suggestion
    condition &&= runDoType3VariableUpdate(doType3, context, doType3Context);
```

msujew (approved):
Thanks, solid work! Just one minor issue, then you can merge the PR yourself 👍

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -319,47 +361,241 @@ function runInstruction(
...
+  // Implicitly declare the loop variable if it doesn't exist, or update existing variable
+  if (!loopVar) {
+    loopVar = {
+      name: varName,
+      declarationNode: variable.reference?.owner,
+      value: {
+        value: start.value,
+        type: inst.DeclaredType.Fixed,
+      },
+      active: true,
```
msujew: As discussed earlier today, implicit variables are always inactive by default. But feel free to wait with that until you have access to the mainframe compiler.

packages/language/src/preprocessor/instruction-interpreter.ts
danieldietrich: Done 👍

packages/language/src/preprocessor/instruction-interpreter.ts
danieldietrich: Good catch, Sir! That makes absolutely sense.

PR #352, opened 2025-08-21
Support preprocessor builtins

msujew:
Depends on #350.

Adds support for all preprocessor builtins. Adds a bunch of tests that confirm that these work as expected.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -49,8 +49,8 @@ type Value = ScalarValue | ArrayValue;
 interface ScalarValue {
   // Scalar preprocessor variables are always strings or numbers.
   // We store them as strings for simplicity and convert them as needed based on the type.
-  value: string;
-  type: inst.DeclaredType;
+  readonly value: string;
+  readonly type: inst.DeclaredType;
 }
 
 function isScalarValue(value: Value): value is ScalarValue {
@@ -65,9 +65,9 @@ function copyScalarValue(value: ScalarValue): ScalarValue {
 }
 
 interface ArrayValue {
-  array: Value[];
-  lower: number;
-  upper: number;
+  readonly array: Value[];
+  readonly lower: number;
+  readonly upper: number;
 }
 
 function isArrayValue(value: Value): value is ArrayValue {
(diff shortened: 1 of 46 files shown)


packages/language/test/fourslash/preprocessor/procedures/builtins/uppercase.ts
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../../../framework.ts" />
+
+//// %DCL X CHAR;
+//// %X = QUOTE(UPPERCASE("hello world"));
```
ssmifi, line 15: _Minor_: Does this test need `QUOTE`? As quote is already tested separately, we could test more specifically here.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -568,17 +626,22 @@ function evaluateUnaryExpression(
...
 
 const defaultEmptyValue: ScalarValue = {
   type: inst.DeclaredType.Character,
   value: "",
 };
+const unsetVariable: ScalarValue = {
+  type: inst.DeclaredType.Character,
+  value: "",
+};
+const zero = numberToValue(0);
```
ssmifi, line 866: _Suggestion_: There are several other `return numberToValue("0")` statements. Could use `zero` for all of them.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1153,3 +1196,341 @@ function resolveIncludeFileUri(
...
+
+type PreprocessorBuiltin = (
+  context: InterpreterContext,
+  // Use the correct typing to ensure that we always handle the case in which an argument is not provided
+  args: (Value | undefined)[],
+) => Value;
+
+const builtinImplementations = new Map<string, PreprocessorBuiltin>();
+
+let collate = "";
```
ssmifi, line 1433: _Minor_: Perhaps we could separate the individual `builtinImplementations.set("XXX", (_, args) => {` by an empty line to improve the overview a little bit.

ssmifi (approved):
Looks good to me, just minor suggestions.

That said, the resolution of the new macro builtin definitions file isn’t working in the playground (and probably not in the web extension either?). If you’d prefer not to tackle this in this PR, we should at least create a backlog issue.

packages/language/test/fourslash/preprocessor/procedures/builtins/uppercase.ts
msujew: Unfortunately yes, since the lexer will automatically uppercase the code. Therefore, it prevents us from testing it, unless it is wrapped in quotes (same for the lowercase function).

PR #354, opened 2025-08-25
[Validation] Implement support for IBM1219I

Wagner-Laranjeiras:
### **Add validation for IBM1219I – LEAVE exits non-iterative DO-groups**

**Summary**

Warn when an unlabeled LEAVE sits inside a non-iterative DO that itself is nested within an iterative DO-loop. In this case the LEAVE exits only the group, not the loop, which is often not what the author intended. Now, in this scenario, we issue a warning.

**Implementation**

1. Skips LEAVE with an explicit label (intended behavior).
2. Walks up the AST to find the nearest enclosing DO.
3. Checks if the nearest DO is non-iterative (doType2/doType3 absent).
4. Searches further up to detect if there is an outer iterative DO.
5. If conditions match, emits warning IBM1219IW at the LEAVE token.

**Changes**

- Added validator IBM1219I_leave_exits_noniterative_do in validation/messages/...
- Parser/Ast: attach leaveToken to LeaveStatement for range/uri resolution
- Added tests: test/fourslash/validate/leave-in-noiterative-do-must-warn.ts and a passing test

**Notes**

- Should I provide more tests?
- I'm still understanding the doType4. Should I include it in the validation?

### DIFF
--- a/packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
+++ b/packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
@@ -1,0 +1,60 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import {
+  Severity,
+  tokenToRange,
+  tokenToUri,
+} from "../../language-server/types";
+import * as AST from "../../syntax-tree/ast";
+import { ValidationAcceptor } from "../validator";
+import * as PLICodes from "./pli-codes";
+
+/**
+ * IBM1219I – LEAVE will exit noniterative DO.
+ *
+ * Warn when an unlabeled LEAVE sits inside a non-iterative DO.
+ */
+export function IBM1219I_leave_exits_noniterative_do(
+  node: AST.LeaveStatement,
+  acceptor: ValidationAcceptor,
+): void {
+  const leaveToken = node.leaveToken;
+
+  // If LEAVE has a label, rule does not apply. If no token is found - return
+  if (node.label || !leaveToken) return;
+
+  // Find nearest enclosing DO
+  let ancestor: AST.SyntaxNode | null = node.container;
+  let nearestDo: AST.DoStatement | undefined;
+  while (ancestor) {
+    if (ancestor.kind === AST.SyntaxKind.DoStatement) {
+      nearestDo = ancestor;
+      break;
+    }
+    ancestor = ancestor.container;
+  }
+  if (!nearestDo) return;
+
+  // Nearest DO must be a non-iterative DO-group
+  if (nearestDo.doType2 || nearestDo.doType3 || nearestDo.doType4) return;
+
+  // Generate range and uri based on token
+  const warningRange = tokenToRange(leaveToken);
+  const warningUri = tokenToUri(leaveToken);
+  if (!warningRange || !warningUri) return;
+
+  acceptor(Severity.W, PLICodes.Warning.IBM1219I.message, {
+    code: PLICodes.Warning.IBM1219I.fullCode,
+    range: warningRange,
+    uri: warningUri,
+  });
+}
--- a/packages/language/src/parser/parser.ts
+++ b/packages/language/src/parser/parser.ts
@@ -4549,6 +4549,7 @@ export class PliParser extends AbstractParser {
       kind: ast.SyntaxKind.LeaveStatement,
       container: null,
       label: null,
+      leaveToken: null,
     };
   }
 
@@ -4557,6 +4558,7 @@ export class PliParser extends AbstractParser {
 
     this.CONSUME_ASSIGN1(tokens.LEAVE, (token) => {
       this.tokenPayload(token, element, CstNodeKind.LeaveStatement_LEAVE);
+      element.leaveToken = token;
     });
     this.OPTION1(() => {
       this.SUBRULE_ASSIGN1(this.LabelReference, {
--- a/packages/language/src/syntax-tree/ast.ts
+++ b/packages/language/src/syntax-tree/ast.ts
@@ -1635,12 +1635,14 @@ export function createLabelReference(): LabelReference {
 export interface LeaveStatement extends AstNode {
   kind: SyntaxKind.LeaveStatement;
   label: LabelReference | null;
+  leaveToken: Token | null;
 }
 export function createLeaveStatement(): LeaveStatement {
   return {
     kind: SyntaxKind.LeaveStatement,
     container: null,
     label: null,
+    leaveToken: null,
   };
 }
 export interface LFormatItem extends AstNode {
--- a/packages/language/src/validation/pli-validator.ts
+++ b/packages/language/src/validation/pli-validator.ts
@@ -24,6 +24,7 @@ import {
   ProcessGroup,
   ProgramConfig,
 } from "../workspace/plugin-configuration-provider";
+import { IBM1219I_leave_exits_noniterative_do } from "./messages/IBM1219I-leave-exits-noniterative-do";
 import { IBM1324IE_name_occurs_more_than_once_within_exports_clause } from "./messages/IBM1324IE-name-occurs-more-than-once-within-exports-clause.js";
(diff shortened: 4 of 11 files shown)


packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
```diff
@@ -0,0 +1,79 @@
...
+    while (ancestor) {
+        if (ancestor.kind === AST.SyntaxKind.DoStatement) {
+            if (ancestor.doType2 || ancestor.doType3) {
+                hasOuterIterative = true;
+                break;
+            }
+        }
+        ancestor = (ancestor as AST.AstNode).container;
+    }
+    if (!hasOuterIterative) return;
```
msujew: Suggestion: This is not necessary. The outer do group in the example is just to illustrate the problem: The user probably wanted to exit the outer do group, but wrote the leave statement inside of the inner do group, which does nothing, because the group is not iterative.

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
```diff
@@ -0,0 +1,79 @@
...
+): void {
+    // If LEAVE has a label, rule does not apply.
+    if (node.label) return;
+
+    // Find nearest enclosing DO
+    let ancestor: AST.SyntaxNode | null = node.container;
+    let nearestDo: AST.DoStatement | undefined;
+    while (ancestor) {
+        if (ancestor.kind === AST.SyntaxKind.DoStatement) {
+            nearestDo = ancestor as AST.DoStatement;
```
msujew: Suggestion: With the way our AST is structured via the discriminated unions (i.e. every AST type has a distinct `kind` field that is shared by no other AST type), you don't need to perform any casting here.
```suggestion
            nearestDo = ancestor;
```

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
```diff
@@ -0,0 +1,79 @@
...
+            break;
+        }
+        ancestor = (ancestor as AST.AstNode).container;
```
msujew: Suggestion: `container` exists as a property on every node. No need to do any casting.
```suggestion
        ancestor = ancestor.container;
```

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    // Retrieve token
+    const leaveToken = node.leaveToken;
+    if (!leaveToken) return;
+
+    // Generate range and uri based on token
+    const warningRange = tokenToRange(leaveToken);
+    if (!warningRange) return;
+    const warningUri = tokenToUri(leaveToken);
+    if (!warningUri) return;
```
msujew: Minor:
```suggestion
    const warningRange = tokenToRange(leaveToken);
    const warningUri = tokenToUri(leaveToken);
    if (!warningUri || !warningRange) return;
```

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Warn when an unlabeled LEAVE sits inside a non-iterative DO
+ * that itself is nested within an iterative DO-loop. In this case the
+ * LEAVE exits only the group, not the loop — often not what the author intended.
+*/
+export function IBM1219I_leave_exits_noniterative_do(
+    node: AST.LeaveStatement,
+    acceptor: ValidationAcceptor,
+): void {
+    // If LEAVE has a label, rule does not apply.
+    if (node.label) return;
```
msujew: Suggestion: We can return early in case no `leaveToken` exists. That also allows to remove the check for that in the code below as well.

```suggestion
    // If LEAVE has a label, rule does not apply.
    if (node.label || !node.leaveToken) return;
```

packages/language/test/fourslash/validate/leave-in-noiterative-do-must-warn.ts
```diff
@@ -0,0 +1,32 @@
...
+/// <reference path="../framework.ts" />
+
+/**
+ * LEAVE inside non-iterative DO-group nested in an iterative DO should warn (IBM1219I)
+ */
+
+// @wrap: main
+//// dcl n fixed bin(31);
+//// dcl i fixed bin(31);
+//// dcl a(*) fixed bin(31);
```
msujew: Minor: `a(*)` is not a valid array declaration (at least in this configuration), use `a(32)` or something similar.

msujew (commented):
Thanks, that already looks pretty good. A few additional things:

1. Please add a test case that only contains `DO; LEAVE; END;`. I've tried it out on the compiler, and that seems to be enough to trigger the warning.
2. Please run the `pnpm pretty` command in the root of the project to format the code correctly.

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
Wagner-Laranjeiras: Thanks Mark! I added a new test, ran pretty and removed the snippet you pointed out.

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
Wagner-Laranjeiras: I did almost exactly as you suggested. Please let me know what you think.

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
```diff
@@ -0,0 +1,63 @@
...
+} from "../../language-server/types";
+import * as AST from "../../syntax-tree/ast";
+import { ValidationAcceptor } from "../validator";
+
+/**
+ * IBM1219I – LEAVE will exit noniterative DO.
+ *
+ * Warn when an unlabeled LEAVE sits inside a non-iterative DO
+ * that itself is nested within an iterative DO-loop. In this case the
+ * LEAVE exits only the group, not the loop — often not what the author intended.
```
msujew: Minor: Please adjust the comment, it's no longer accurate.

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
Wagner-Laranjeiras: Question: That structure crossed my mind, but my decision revolved around performance. I wanted to make sure to just call the second method if completely necessary, in order to avoid unnecessary resources consumption. Would it be negligible in terms of performance then? (I've applied your suggestion anyway - it's just a matter of curiosity)

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (ancestor.kind === AST.SyntaxKind.DoStatement) {
+      nearestDo = ancestor;
+      break;
+    }
+    ancestor = ancestor.container;
+  }
+  if (!nearestDo) return;
+
+  // Nearest DO must be a non-iterative DO-group
+  if (nearestDo.doType2 || nearestDo.doType3) return;
```
msujew: Suggestion: DoType4 (i.e. `DO FOREVER; END;`) is also an iterative DO-group and should therefore also return here.

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { PLICodes } from ".";
```
msujew: Minor: Please don't import from the index (i.e. `"."`).

msujew (commented):
Thanks, just a few minor issues, see below. Could you also add negative tests to your changes? Ideally, the tests should also show for example that adding a label or using a `DO FOREVER` do not trigger the validation. You can use the `verify.noDiagnostics(label)` method to do that.

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
msujew: Yes, in these cases it's negligible :)

packages/language/src/validation/messages/IBM1219I-leave-exits-noniterative-do.ts
msujew: Works well, thank you 👍

msujew (approved):
Thanks, looks good to me!

Please squash your commits into one and add a commit detail `Signed-off-by` to pass the DCO. Afterwards we can merge the PR :)

Issue #347, opened 2025-08-19
[Validation] Implement support for IBM2615I


msujew:
Refer to:

https://github.com/zowe/zowe-pli-language-support/blob/cbd0f4f321a5db83fcadd028160164cafdcacab2/packages/language/src/validation/messages/pli-codes.ts#L3615-L3633

PR #355, opened 2025-08-25
Support member calls on DoType3

msujew:
Fixes an issue related to https://github.com/zowe/zowe-pli-language-support/issues/347.

Currently, we only support "flat" references on the DoType3 parser. This is not correct, as the compiler actually allows to use member calls (i.e. `A.B.C...`) there as well.


ssmifi (approved):
LGTM

PR #356, opened 2025-08-25
Improve support for includes within copybooks

msujew:
Fixes point 4 of https://github.com/zowe/zowe-pli-language-support/issues/335.

The issue can be explained as follows: Imagine you have opened your IDE, and are currently looking through the copybook (without having opened an entry file of the compilation unit). This means, the traditional way to lookup process configs doesn't work anymore, since we don't know what the actual entry point of a file is.

This PR attempts to circumvent this issue a bit, by adding a new method to the process config provider that uses the current file URI to see whether it's within one of the copybooks of the registered process configs. If it is, then it should actually return the corresponding process config. This allows the currently open file to perform includes, etc.


ssmifi (approved):
LGTM

PR #342, opened 2025-08-19
Extend compiler options and tests #3

ssmifi:
Part of https://github.com/zowe/zowe-pli-language-support/issues/296
Based on https://github.com/zowe/zowe-pli-language-support/pull/318

- Implemented compiler options `maxNest`, `include`, `maxRunOnIf`, `maxStatic`, `maxStmt`, `maxTemp`, `mDeck`, `msgSummary`, `name`, `names`, `natLang`, `nest`, `nullDate`, `object`, `offset`, `offsetSize`, `onSnap`, `optimize`, and `option`.
- Added `noDiagnosticsExceptAt` to also only test at specific labels.

PR #357, opened 2025-08-26
Extend compiler options and tests #4

ssmifi:
Based on https://github.com/zowe/zowe-pli-language-support/pull/342

- Implemented compiler options `precType`, `proceed`, `process`, `quote`, `reduce`, `rent`, `resExp`, `respect`, `rtCheck`, `semantic`, `service`, `source`, `stdsys`, `stmt`, `storage`, `stringOfGraphic`, and `syntax`.
- Fixed a bug that prevented the harness wrapper to wrap strings like `'$'`.
- Extended the harness API such that individual indices can be replaced by generated strings and tests can be executed manually, see `service.ts` test for an example.

### DIFF
--- a/packages/language/test/fourslash/preprocessor/compiler-options/service.ts
+++ b/packages/language/test/fourslash/preprocessor/compiler-options/service.ts
@@ -1,0 +1,38 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../../framework.ts" />
+
+// @wrap: process
+////*PROCESS <|1:NOSERVICE|>;
+////*PROCESS <|2:SERVICE|>;
+////*PROCESS <|4:SERVICE|>(<|5:)|>;
+////*PROCESS <|6:SERVICE|>(<|7:"xXx"|>);
+////*PROCESS <|8:SERVICE|>(<|9:'yYy'|>);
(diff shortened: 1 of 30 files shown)


packages/language/test/fourslash/preprocessor/compiler-options/service.ts
```diff
@@ -0,0 +1,42 @@
...
+////*PROCESS <|1:NOSERVICE|>;
+////*PROCESS <|2:SERVICE|>;
+////*PROCESS <|4:SERVICE|>(<|5:)|>;
+////*PROCESS <|6:SERVICE|>(<|7:"xXx"|>);
+////*PROCESS <|8:SERVICE|>(<|9:'yYy'|>);
+////*PROCESS <|10:SERVICE|>(<|11>);
+////*PROCESS <|12:SERVICE|>(xYz);
+
+testAPI.replaceIndex(11, "x".repeat(128));
+testAPI.execute();
```
msujew: I'm not sure this is a good idea - as in: the whole `replaceIndex`, `@deferred` and `execute` system. One of the advantages that the fourslash framework has, is that it's not possible to write a test that acts on generated code. Instead, all the code needs to be stated explicitly in the test document. This `replaceIndex` method allows to circumvent this completely.

If you need a test that generates code, please don't use the fourslash framework to do this. There are also a few other issues related to this (i.e. tokenization eventually becoming async, which will make `execute` async as well, etc.), which will make the testing framework much harder to maintain in the future.

msujew (commented):
Thanks, looks generally pretty good. I just have an issue with the `replaceIndex` feature from an architectural standpoint, see below.

packages/language/test/fourslash/preprocessor/compiler-options/service.ts
ssmifi: I added this for convenience, but I see your point. I restored the previous harness interface. 👍

msujew (approved):
Thanks, looks good to me 👍

PR #358, opened 2025-08-26
Convert preprocessor parser to functions

msujew:
This is a non-functional change that converts the old `PliPreprocessorParser` class into a list of functions.

This will be eventually helpful when we refactor our parser to be handwritten.


ssmifi (approved):
LGTM

Issue #359, opened 2025-08-26
[Bug] Playground: Go To Definition Jumps to Incorrectly Cased File


montymxb:
Observed on the latest dev playground as of Aug. 26th.

It looks like a recent change to resolve case-insensitivity issues w/ imports in the playground may have left a bug with Go to Definition requests. In all cases, it appears we only jump to the lower-cased file name. So having a file in your workspace called **cpy/ABC.pli**, and a program like so:
```c
 %INCLUDE ABC;
```

Will import successfully, but jumping to the imported file will take you to **cpy/abc.pli**, which doesn't exist in the vfs, and fails.

Same applies for `%INCLUDE "ABC.pli";` as well.

We are also jumping to files with different casing in the native extension as well. This isn't a problem for case-insensitive file-systems (on Windows/macOS), but will probably be an issue on Linux with ext4.

PR #360, opened 2025-08-26
Update Web Extension 

### LINKED ISSUE #317: Update Web-Extension
The web-extension has fallen a bit behind the regular extension & playground implementation. It's missing some key details that prevent it from working correctly for most new features, notably a file system integration (it's using an empty one at the moment). This blocks anything related to the pli config as well.

montymxb:
Closes #317 by bringing the web extension up to parity with regards to the existing playground state.

Points that I'm still thinking of:
- This more or less mimics the approach in the playground, where we're pre-emptively loading most of the workspace (ex. all proc groups are iterated over and loaded). This won't work for larger workspaces. I had a thought about adding a notification that the LS can send back to the LC to 'request' for documents to be looked for & added lazily on demand to avoid this, but nothing in that area has been put into place here.
- main-browser changes are a bit large, especially all the traversal logic for loading the workspace into the VFS, open for suggestions on improving that
- Ran into timing issues w/ regards to the LC starting while we were still sending files to the LS worker, that's been resolved here, but larger workspaces could trigger strange diagnostics for a currently opened file while we're still populating the workspace on the LS side. Especially before the plugin configuration is reloaded.
- This doesn't replace the playground implementation _yet_, but it mirrors several parts. Ideally either this PR is pushed up to subsume those duplicate parts, or a subsequent one should bring those together
    - could be either directly incorporating the web-extension, or directly replacing & reusing this LS & LC setup there

This looks to also resolves a recent bug in the playground where case-insensitivity of imports leads to Go To Def requests jumping to non-existent files whenever the actual file is not fully lower-cased. So `%INCLUDE abc` fails when the file is actually `ABC`. Although the fixed state is just in the web-extension here. I've opened up #359 to track that separately, especially since this doesn't modify the playground yet. Note that although the jumped to file may not be there, the import still resolves correctly as expected.

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -116,7 +116,8 @@
     "pli-language": "workspace:*",
     "util": "^0.12.5",
     "vscode-languageclient": "~9.0.1",
-    "vscode-languageserver": "~9.0.1"
+    "vscode-languageserver": "~9.0.1",
+    "vscode-uri": "~3.0.8"
   },
   "devDependencies": {
     "@types/vscode": "~1.67.0",
--- a/packages/vscode-extension/src/extension/main-browser.ts
+++ b/packages/vscode-extension/src/extension/main-browser.ts
@@ -9,33 +9,140 @@
  *
  */
 
-import type { LanguageClientOptions } from "vscode-languageclient/browser.js";
+import { WorkspaceDidChangePlipluginConfigNotification } from "pli-language";
 import * as vscode from "vscode";
+import type { LanguageClientOptions } from "vscode-languageclient/browser.js";
 import { LanguageClient } from "vscode-languageclient/browser.js";
 import { BuiltinFileSystemProvider } from "./builtin-files";
+import {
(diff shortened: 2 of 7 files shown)


ssmifi (commented):
Looks good to me. Imho it's fine to move the playground/web extension sync into an upcoming PR. Can you create an issue with your findings or a link to this PR?

montymxb:
Definitely, just sketched that up under #370 with a link back here.

montymxb:
@msujew do you think you take a look before we roll this one in? ~Still needs a rebase +~ those follow-up points mentioned in the description, but beyond that another pair of eyes would be appreciated.

msujew (commented):
I feel like this is now unncessarily complicated - now that we have an async file system. Why not simply request the files from the language server directly via the lsp-connection -> VS Code `fs` API? That was most of the motivation behind making the API async in the first place.

msujew:
Superseeded by https://github.com/zowe/zowe-pli-language-support/pull/406.

PR #361, opened 2025-08-27
Fix preprocessor variable lookup

msujew:
Fixes an issue introduced by https://github.com/zowe/zowe-pli-language-support/commit/7bb709cceb4ce10ca02a91cf4556bb950316b552 in https://github.com/zowe/zowe-pli-language-support/pull/351. Reverts some of the changes and adds tests to ensure that these issues don't crop up in the future.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -89,10 +89,30 @@ function copyValue(value: Value): Value {
   }
 }
 
-function variables(context: InterpreterContext): Map<string, Variable> {
-  // Returns the correct variable map for the current scope:
-  // local (in a procedure call), otherwise global
-  return context.localVariables ?? context.variables;
+function getVariable(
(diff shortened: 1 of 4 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -89,10 +89,19 @@ function copyValue(value: Value): Value {
...
+): Variable | undefined {
+  return context.localVariables?.get(name) ?? context.variables.get(name);
+}
+
+function setVariable(context: InterpreterContext, variable: Variable): void {
+  if (context.localVariables?.has(variable.name)) {
+    context.localVariables.set(variable.name, variable);
+  } else {
+    context.variables.set(variable.name, variable);
+  }
```
danieldietrich, line 115: This makes sense! I had this implementation too but somehow reverted it in #351.

danieldietrich (approved):
LGTM!

PR #362, opened 2025-08-27
Support multiple INCAFTER & FLAG Options

montymxb:
Closes point 1 from #335 by:

- Supporting multiple compiler options declarations for INCAFTER & FLAG
- Adds a 'multipleAllowed' prop to ID such options
- Avoids producing a duplicate option validation for these options
- INCAFTER accumulates includes for each case
- Flag just replaces it's value (I don't think it can accumulate after reading it over, seems like it just overrides like any other rule?)
- Tests to ensure multiple options of INCAFTER & FLAG don't trigger diagnostics + verifying multiple INCAFTERs work as expected.

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/translator.ts
+++ b/packages/language/src/preprocessor/compiler-options/translator.ts
@@ -36,6 +36,7 @@ interface TranslatorRule {
   negative?: string[];
   positiveTranslate?: Translate;
   negativeTranslate?: Translate;
+  multipleAllowed?: boolean;
 }
 
 type Translate = (option: CompilerOption, options: CompilerOptions) => void;
@@ -67,17 +68,27 @@ class Translator {
 
   private rules: TranslatorRule[] = [];
 
+  /**
+   * Adds a rule to the translator.
+   * @param positive Positive option names that trigger this rule.
+   * @param positiveTranslate Translation function for positive option.
+   * @param negative Negative option names that trigger this rule.
+   * @param negativeTranslate Negative translation function for negative option.
+   * @param multipleAllowed Whether this rule supports multiple occurrences. Duplicate validations are skipped in this case.
+   */
   rule(
     positive: string[],
     positiveTranslate: Translate,
     negative?: string[],
     negativeTranslate?: Translate,
+    multipleAllowed?: boolean,
   ) {
     this.rules.push({
       positive,
       negative,
       positiveTranslate,
       negativeTranslate,
+      multipleAllowed: multipleAllowed,
     });
   }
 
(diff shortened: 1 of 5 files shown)


packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -1368,23 +1381,29 @@ translator.rule(
...
+          1,
+        );
+      }
     }
-  }
-});
+  },
+  undefined,
+  undefined,
+  true,
```
ssmifi, line 1405: _Question_: Can we please reiterate why `FLAG` should not report a duplicate warning? 
Yes, the previous setting will be overwritten, but this is practically true for all compiler options with the exception of some that actually accumulate the settings. But that's the rationale of the warning, right?

packages/language/src/preprocessor/pli-lexer.ts
```diff
@@ -135,30 +135,41 @@ function generateIncAfterInstruction(
...
-  includeItem.token = incAfter.token || null;
-  if (incAfter.token) {
-    includeItem.token = incAfter.token;
-    incAfter.token.element = includeItem;
-    incAfter.token.kind = CstNodeKind.IncludeItem_FileID;
+
+  // generate chained include instructions for each entry
+  let currentNode = existingNode;
+
+  // process in reverse order so the first INCAFTER appears first
```
ssmifi, line 145: _Question_: Did you mean the `last` appears first?
Ist this behavior confirmed on the mainframe?

ssmifi (commented):
Code looks good, but I still have two questions about the rationale of the change, see below.

packages/language/src/preprocessor/pli-lexer.ts
ssmifi: I wanted to run a quick test on the mainframe, but 
```
*PROCESS INCAFTER(PROCESS(test3));
*PROCESS INCAFTER(PROCESS(test4));
```
actually only gets reported as `+ INCAFTER( PROCESS( "TEST4" ) )` in the options used listing. Could be that the listing is incomplete in that case, but can you please verify that `INCAFTER` behaves as described in #335?

packages/language/src/preprocessor/compiler-options/translator.ts
montymxb: It's a good question. The only reason it's here is that it was stated in #335, but you're absolutely right it's no different than the other options in this regard, it just replaces its value. We can probably bring this one up later today for clarification, especially with regards to all the other options.

packages/language/src/preprocessor/pli-lexer.ts
montymxb: It is supposed to be that the first appears first, since it's a linked instruction list via next, so we build it backwards to get that result.

As for the behavior, I'll double check. This behavior is not in the documentation, but it has been reported to us as being needed. I'm also not sure if our test env will reflect this, since we're aware of discrepancies. I'll follow up here.

montymxb: Can confirm the single entry in the listings, we should bring this up later today to clarify before proceeding.

montymxb:
Closing out. Existing behavior has been confirmed to be correct & desired after discussion.

PR #363, opened 2025-08-27
Implement support for IBM1068I

Wagner-Laranjeiras:
### Implement IBM1068I — require RETURNS when procedure contains RETURN

Issue: [#345 ](https://github.com/zowe/zowe-pli-language-support/issues/345)

**Summary**

This PR adds a validator that implements IBM1068I: when a PROCEDURE contains a RETURN statement, it should declare a RETURNS attribute on the PROCEDURE statement. If a RETURN is found and the PROCEDURE does not provide a RETURNS option, the validator emits an Information diagnostic at the RETURN keyword.

**Changes**

- IBM1068I_proc_with_return_hasnt_returns_att: validation code described above.
- pli-validator: register validation code for IBM1068I.
- Parser: update ReturnStatement rule to assign element.returnToken = token in the CONSUME(tokens.RETURN, ...) callback
- AST: add returnToken to ReturnStatement
- Tests: add/update tests under packages/language/test/fourslash/validate/IBM1068I/


packages/language/src/validation/messages/Infos/IBM1068I-proc-with-return-hasnt-returns-att.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ */
+export function IBM1068I_proc_with_return_hasnt_returns_att(
+  node: AST.ProcedureStatement,
+  acceptor: ValidationAcceptor,
+) {
+  // Check for RETURN statements
+  const returnStmt = node.statements?.find(
+    (stmt) => stmt.value?.kind === AST.SyntaxKind.ReturnStatement,
+  );
```
msujew: Suggestion: As soon as the procedure becomes more complicated, this simple search will no longer be able to find the return statement:

```
TEST: PROC RETURNS(FIXED);
  IF X > 5 THEN
    RETURN 1;
  ELSE
    RETURN 0;
END;
```

The `node` will only contain the `IfStatement`, whereas the return statements are nested inside of the `IfStatement`.

You can use the `forEachNode` function recursively to iterate over all children of any `AstNode`.

packages/language/src/validation/messages/Infos/IBM1068I-proc-with-return-hasnt-returns-att.ts
```diff
@@ -0,0 +1,61 @@
...
+  // Retrieve token of RETURN statement (NOT from the ProcedureStatement)
+  const token = typedReturn.returnToken;
+  if (!token) return;
+
+  // Build diagnostic
+  const infoRange = tokenToRange(token);
+  const infoUri = tokenToUri(token);
+  if (!infoRange || !infoUri) return;
+
+  acceptor(Severity.I, PLICodes.Info.IBM1068I.message, {
```
msujew: Minor: I just noticed that the compiler actually throws `IBM2412I` in this case. Weirdly enough, these error codes are identical, except for the fact that `IBM2412I` is an `E` (error) code, while `IBM1068I` is an I (info) code.

packages/language/src/validation/messages/Infos/IBM1068I-proc-with-return-hasnt-returns-att.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (hasReturnsAtt) return;
+
+  // Retrieve token of RETURN statement (NOT from the ProcedureStatement)
+  const token = typedReturn.returnToken;
+  if (!token) return;
+
+  // Build diagnostic
+  const infoRange = tokenToRange(token);
+  const infoUri = tokenToUri(token);
+  if (!infoRange || !infoUri) return;
```
msujew: Suggestion: The compiler actually shows the error on the `PROC` token of the containing procedure, not on the return token. This makes a lot more sense, as there might be multiple `RETURN` tokens, but only one `PROC` token per procedure (and we only want to show the error once).

packages/language/src/validation/messages/Infos/IBM1068I-proc-with-return-hasnt-returns-att.ts
```diff
@@ -0,0 +1,61 @@
...
+) {
+  // Check for RETURN statements
+  const returnStmt = node.statements?.find(
+    (stmt) => stmt.value?.kind === AST.SyntaxKind.ReturnStatement,
+  );
+  if (!returnStmt) return;
+
+  // Type-narrowing in order to access token
+  if (returnStmt.value?.kind !== AST.SyntaxKind.ReturnStatement) return;
+  const typedReturn = returnStmt.value as AST.ReturnStatement;
```
msujew: Question: Why do type narrowing when we already have done the check previously AND a cast afterwards? One of these seem redundant.

packages/language/test/fourslash/validate/IBM1068I/proc-without-return-with-returns.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Procedure that declares RETURNS but contains no RETURN must NOT trigger IBM1068I
+ */
+
+// @wrap: main
+//// d: proc returns(fixed bin(31));
+//// end;
+
+verify.noDiagnostics();
```
msujew: I'm not sure that it's a good idea to add this test. It will break once more validations are implemented (this is not a valid procedure), and therefore currently only tests the current behavior of the language server - not how the language server **should** behave.

packages/language/src/validation/messages/Infos/IBM1068I-proc-with-return-hasnt-returns-att.ts
Wagner-Laranjeiras: Got it! I was under the impression that we should show the error on the last `RETURN` statement. Changed that.

packages/language/src/validation/messages/Infos/IBM1068I-proc-with-return-hasnt-returns-att.ts
Wagner-Laranjeiras: It was indeed redundant. I ended up removing those lines entirely, but thanks for the heads up.

packages/language/test/fourslash/validate/IBM1068I/proc-without-return-with-returns.ts
Wagner-Laranjeiras: Got it! That makes sense. Test removed.

packages/language/src/validation/messages/Infos/IBM1068I-proc-with-return-hasnt-returns-att.ts
Wagner-Laranjeiras: That's really strange. I could prioritize the `IBM2412I` implementation, since an error should have priority over an info (I assume). Please let me know.

packages/language/src/validation/messages/Infos/IBM1068I-proc-with-return-hasnt-returns-att.ts
Wagner-Laranjeiras: Working on it! Thanks for the feedback. I've wrote some helpers to help us deal with these cases. They are under the pull request [#379](https://github.com/zowe/zowe-pli-language-support/pull/379).

Wagner-Laranjeiras:
Please check [PR #379](https://github.com/zowe/zowe-pli-language-support/pull/379) first. :)

packages/language/src/validation/messages/info-severity/IBM1068I-proc-with-return-hasnt-returns-att.ts
```diff
@@ -0,0 +1,56 @@
...
+  if (!token) return;
+
+  // Check if procedure has RETURNS attribute - return if present
+  const hasReturnsAtt = node.options?.some(
+    (att) => att.kind === AST.SyntaxKind.ReturnsOption,
+  );
+  if (hasReturnsAtt) return;
+  // Still has to pass proc-with-return-inside-if-with-returns -> option is being read first as 135
+
+  const returnStmt = findFirstNodeOfKind(node, AST.SyntaxKind.ReturnStatement);
```
msujew, line 43: Suggestion: This will result in false positives for nested procedures:

```
OUTER: PROC;
  INNER: PROC RETURNS(FIXED);
    RETURN (0);
  END INNER;
END OUTER;
```

Please also add a test for this.

PR #366, opened 2025-08-27
Playground Workspace Share Fix

### LINKED ISSUE #339: Share Workspace Missing Files in Playground
When sharing a workspace in the [playground](https://zowe.github.io/zowe-pli-language-support/dev/), it seems that oftentimes the folder containing lib files is completely omitted. This needs some investigation to pin down & resolve the issue.

montymxb:
Closes #339 by loading share files via the vscode fs api, instead of the textDocuments that are known to the editor. Previously this would result in sharing sometimes a sub-set of the files  that were actually present, especially new additions which were not in open tabs.

### DIFF
--- a/packages/playground/src/workspace.ts
+++ b/packages/playground/src/workspace.ts
@@ -45,8 +45,8 @@ export function registerButtons() {
   const shareWorkspaceButton = document.getElementById(
     "share-workspace-button",
   );
-  shareWorkspaceButton?.addEventListener("click", () => {
-    const workspaceFiles = createWorkspaceFiles(getWorkspaceDocuments());
+  shareWorkspaceButton?.addEventListener("click", async () => {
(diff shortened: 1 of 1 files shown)


ssmifi (approved):
LGTM. 

Probably unrelated to your change, but in my test I noticed that the LS only responses after a change in the first file (that is displayed by default.) Selecting a different file or triggering a change provokes a response. If you can confirm this on your system, we should probably add an issue for this.

montymxb:
Interesting, I'll double check that in a bit here.

montymxb:
Good catch, yeah I can replicate that quite easily. It's also the same on the live [dev](https://zowe.github.io/zowe-pli-language-support/dev/) playground as well, so it's not related to this. I've documented it under #371 for tracking going forward.

PR #369, opened 2025-08-27
Support inline procedure calls

msujew:
Enables users to call activated preprocessor procedures in normal PLI code. Takes special care of the `STATEMENT` attribute of procedures as per the documentation:

image

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -1182,7 +1182,7 @@ function performTokenScan(
   let advance: number = 1;
   let immediateFollow = token.immediateFollow;
   if (procedure && context.activeProcedures.has(image)) {
-    const label = procedure.labels.find((label) => label.name === image);
+    const label = procedure.labels.get(image);
     refNode = label;
     const procedureParseResult = parseAndEvaluateProcedure(
       tokens,
@@ -1190,11 +1190,9 @@ function performTokenScan(
       procedure,
       context,
     );
-    if (procedureParseResult) {
-      value = procedureParseResult.value;
-      advance = procedureParseResult.advance;
-      immediateFollow = procedureParseResult.immediateFollow;
-    }
+    value = procedureParseResult.value;
+    advance = procedureParseResult.advance;
+    immediateFollow = procedureParseResult.immediateFollow;
   } else if (variable?.active) {
     refNode = variable.declarationNode ?? undefined;
     value = variable.value;
@@ -1230,10 +1228,11 @@ function parseAndEvaluateProcedure(
   index: number,
   procedure: inst.ProcedureInstructionContainer,
   context: InterpreterContext,
-): InlineProcedureEvaluationResult | undefined {
+): InlineProcedureEvaluationResult {
   const evaluatedArgs: Value[] = [];
   let advance = 1;
   let immediateFollow = tokens[index].immediateFollow;
+  let suffix: string = "";
   if (!procedure.statement) {
     const parseResult = parseInlineProcedureInvocation(tokens, index);
     advance = parseResult.advance;
@@ -1250,11 +1249,12 @@ function parseAndEvaluateProcedure(
     const parseResult = parseInlineStatementProcedureInvocation(tokens, index);
     advance = parseResult.advance;
     immediateFollow = parseResult.immediateFollow;
+    suffix = parseResult.suffix;
     // Positional arguments
     for (const argTokens of parseResult.positionalArgs) {
       const replacedTokens = replaceTokensInText(argTokens, context);
       const text = stringifyTokens(replacedTokens);
-      evaluatedArgs.push(valueToString(text));
+      evaluatedArgs.push(stringToValue(text));
     }
     const usedParams = new Set<string>();
     // Named arguments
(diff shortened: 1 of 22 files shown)


packages/language/src/preprocessor/pli-lexer.ts
```diff
@@ -37,6 +37,7 @@ export interface LexingIssue {
   readonly severity: Severity;
   readonly range: Range | undefined;
   readonly uri: URI | undefined;
+  readonly code?: string;
```
Lotes, line 40: Minor idea: Instead of type `string`, we could use `keyof typeof Info`, where `Info` comes from the `pli-codes.ts`. We will not have any other codes or?

packages/language/test/fourslash/preprocessor/procedures/inline-procedure-with-arguments-merges-with-next-token.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ */
+
+/// <reference path="../../framework.ts" />
+
+//// %TEST: PROC (X) RETURNS (CHAR);
+////   RETURN (X);
+//// %END;
+//// %ACTIVATE TEST;
+//// TEST(VALUE)%;VAR
```
Lotes, line 18: "%;"? Concats with the following text? I was wondering about this syntax. Is there a real-world example or are there docs? O.o crazy

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1139,51 +1144,337 @@ function replaceTokensInText(
...
+  };
+}
+
+interface InlineProcedureEvaluationResult {
+  value: Value | undefined;
+  advance: number;
+  immediateFollow: boolean;
+}
+
+function parseAndEvaluateProcedure(
```
Lotes, line 1226: The functions here are very dense. Do you see any opportunities to make the code more readable? Like splitting the code into smaller functions.
The comments help a bit. Having a smaller context to think about would help understanding the core of your change much better.

packages/language/test/fourslash/preprocessor/procedures/inline-procedure-with-arguments-merges-with-next-token.ts
Lotes: * because semicolon normally finalizes a statement... it would make sense to me if there was no semicolon ...

packages/language/test/fourslash/linker/preprocessor-inline-procedure.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ Preprocessor variables can be linked from PL/I code.
+ */
+//// %<|1:TEST|>: PROC RETURNS (CHAR);
+////   RETURN ("VALUE");
+//// %END;
+//// %ACTIVATE TEST;
+//// <|1>TEST
+
+linker.expectLinks();
```
Lotes, line 23: Cool testing framework !

packages/language/test/fourslash/preprocessor/procedures/inline-procedure-with-arguments-merges-with-next-token.ts
msujew: Yes, it's the famous "null statement". The documentation with regards to `%` is actually incorrect, I.e. the `DCL A%B FIXED` stuff that is in the documentation actually does not compile. `DCL A%;B FIXED` does however compile correctly:

image

Lotes: OK, thanks

packages/language/src/preprocessor/pli-lexer.ts
msujew: We actually do have additional codes, see [here](https://github.com/zowe/zowe-pli-language-support/blob/development/packages/language/src/preprocessor/compiler-options/codes.ts). Note that the `fullCode` also differs from the name of the code (usually by a single-character suffix).

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1299,6 +1269,54 @@ function parseAndEvaluateProcedure(
...
+    const text = stringifyTokens(replacedTokens);
+    evaluatedArgs.push(stringToValue(text));
+  }
+  return evaluatedArgs;
+}
+
+function evaluateNamedArguments(
+  args: Value[],
+  namedArgs: Map<string, InlineProcedureNamedArgument>,
+  procArgs: string[],
```
Lotes: Rename to `procParams`. I like your distinction between parameters and arguments. We should keep this difference consistent :-).

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1229,67 +1229,37 @@ function parseAndEvaluateProcedure(
...
-          new PreprocessorError(
-            PLICodes.Error.IBM3581I.message(argTokens.nameToken.image),
-            argTokens.nameToken,
-            undefined,
-            PLICodes.Error.IBM3581I.fullCode,
-          ),
-        );
-      }
-    }
+    evaluatedArgs = evaluatePositionalArguments(
```
Lotes, line 1246: Thanks for moving out. It helps :D

Lotes (approved):
One little remark about the naming above. It is much more readable now (unbelievable! :D thx). LGTM.

Issue #374, opened 2025-08-28
Findings 28.8


asatklichov:
Compiler Options related 

- [x] a.    Currently, you can not add comments after `%PROCESS xyz;` compiler options
 <img width="1120" height="105" alt="Image" src="https://github.com/user-attachments/assets/484c4b20-374f-4763-8c1a-ce22551ec400" />

- [x] b.  `%PROCESS default(ALIGNED, aligneD); `  //validation missing 

- [ ] c. `%PROCESS BLANK('$');`   //will be provided more info after customer visit

- [x] d. Should provide validate - out of range 
 `%PROCESS margins(102,201); `
 
 
 - [x]  e.  `%PROCESS optimize(); ` //should should show error for empty brackets.
 
 
 - [x] f. `%PROCESS or('');`  //should show error for blank in between quotation

- [x] g.  LARGE file issues on `%INCLUDE ABC; ` still exist on includes and in nested includes as well

- [x] f.  Findings on 10.9  word file provided - please, go through items 17-29 

- [x] g.  Findings on 18.9  word file provided - please, go through items 30-38

ssmifi:
e) is accepted by the mainframe and equivalent to `%PROCESS optimize(0);` although it is not specifically mentioned in the spec.

ssmifi:
Results regarding f)

17.: Dependency interaction between different options are not yet implemented, but are tracked here: https://github.com/zowe/zowe-pli-language-support/issues/322

18 a: `*PROCESS MSGSUMMARY;` is valid on the mainframe and equivalent to `MSGSUMMARY( NOXREF )`. Is logged in https://github.com/zowe/zowe-pli-language-support/issues/321.
18 b: `*PROCESS MSGSUMMARY();` is not valid on the mainframe and results in IBM1161I, IBM1158I. Added explicit note to https://github.com/zowe/zowe-pli-language-support/issues/321.

19: Limits were tested on the mainframe. Logged in `translator.ts`, line 2116. I'll add it to the list of discrepancies in https://github.com/zowe/zowe-pli-language-support/issues/321.

20 a: `*PROCESS MAXMSG;` Is invalid on the mainframe. Added this case specifically to https://github.com/zowe/zowe-pli-language-support/issues/321.
20 b: `*PROCESS MAXMSG();` Is valid on the mainframe and equivalent to the defaults. Is logged in https://github.com/zowe/zowe-pli-language-support/issues/321.

[...]

PR #377, opened 2025-08-29
Fix comments following *process

ssmifi:
Suggestion for https://github.com/zowe/zowe-pli-language-support/issues/374 a

Fixes the obvious cases of ML and SL comments on the process line. 
However, as stated in https://github.com/zowe/zowe-pli-language-support/pull/338 everything behind the `;` is basically just ignored on the mainframe. So, in 
```
*PROCESS MARGINS(2, 72); /* xxx 
*PROCESS MARGINS(2, 74);  xxx */
```
the second directive is actually executed normally. Since our options currently parses the directives separately, tokens like `/*` become possible. Also, the syntax highlighting does not reflect this. So 
```
*PROCESS MARGINS(2, 72); 
*PROCESS MARGINS(2, 74);  /* xxx 
```
is valid, but the program afterwards is highlighted as a comment.

Conversely, 
```
*PROCESS MARGINS(2, 100); /* test 
 */
```
is not valid code on the mainframe.

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/parser.ts
+++ b/packages/language/src/preprocessor/compiler-options/parser.ts
@@ -26,7 +26,7 @@ import {
   CompilerOptionValue,
   SyntaxKind,
 } from "../../syntax-tree/ast";
-import { Token } from "../../parser/tokens";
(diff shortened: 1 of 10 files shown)


msujew (approved):
Makes sense to me, thanks 👍

PR #368, opened 2025-08-27
Extend compiler options and tests #5

ssmifi:
Part of https://github.com/zowe/zowe-pli-language-support/issues/296
Based on https://github.com/zowe/zowe-pli-language-support/pull/357

- Implemented compiler options `terminal`, `test`, `unroll`, `usage`, `widechar`, `window`, `writable`, `xinfo`, `xml`, `xref`, and `prefix`.
- `noDiagnosticsExcept` also accepts arrays now.
- Added `ensureArgument` to compiler options translator to add some convenience to argument selection. Should probably be applied to some old options to make the code more concise in a later PR.

PR #378, opened 2025-09-01
Add range validation for Margins compiler option

### LINKED ISSUE #374: Findings 28.8

ssmifi:
Closes https://github.com/zowe/zowe-pli-language-support/issues/374 d
Based on https://github.com/zowe/zowe-pli-language-support/pull/368

- Added range checks for the margins process directive (and refactored to latest changes). 
- `expectDiagnosticsExcept` now also accepts plain strings.
- Moved margins compiler options tests to the fourslash harness.

### DIFF
--- a/packages/language/test/test-builder.ts
+++ b/packages/language/test/test-builder.ts
@@ -39,6 +39,7 @@ import {
 import { InternalCodes } from "../src/validation/messages";
 import { CompilerOptions } from "../src/preprocessor/compiler-options/options";
 import { tokenize } from "../src/parser/tokenizer";
+import { escapeRegExp } from "../src/parser/tokens";
 
 export type Label = string | number | string[] | number[];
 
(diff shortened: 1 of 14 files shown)


packages/language/test/test-builder.ts
```diff
@@ -480,13 +480,29 @@ export class TestBuilder {
...
+    }
+    if (typeof exceptions[0] === "string") {
+      this.noDiagnosticsExcept(
+        (exceptions as string[]).map(
+          (s) => new RegExp(s.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")),
+        ),
+        label,
+      );
+      return this;
+    }
```
msujew: Suggestion: I think it makes more sense to convert all of them to regex and continue.
```suggestion
    if (typeof exceptions[0] === "string") {
      exceptions = (exceptions as string[]).map(escapeRegExp);
    }
```

msujew (approved):
Looks good to me, just one issue, see below.

packages/language/test/test-builder.ts
ssmifi: 👍 makes sense. thanks!

PR #379, opened 2025-09-01
Add utility functions for syntax node traversal and filtering.

Wagner-Laranjeiras:
### Add utility functions for syntax node traversal and filtering

Issue: N/A

**Summary**

This PR introduces three general-purpose AST utilities to support syntax tree analysis in a consistent and reusable way. These functions are especially useful when implementing validators that need to locate or examine nodes of a specific kind.

**Changes**

_traverseAllNodes:_ Recursively visits all nodes in the syntax tree using pre-order traversal. Used internally by the other two functions.

_containsNodeOfKind:_ Checks whether a subtree contains any node of a given SyntaxKind.

_collectNodesOfKind:_ Collects all nodes of a given SyntaxKind under a syntax tree root.

These utilities reduce repetitive traversal logic and enable more declarative validation rules.

### DIFF
--- a/packages/language/src/syntax-tree/ast-iterator.ts
+++ b/packages/language/src/syntax-tree/ast-iterator.ts
@@ -11,6 +11,20 @@
 
 import { SyntaxKind, SyntaxNode } from "./ast";
 
+export enum TraversalState {
+  Continue,
+  Stop,
+}
+
+/**
+ * Applies the given `action` to relevant child nodes of the provided `SyntaxNode`,
+ * based on its `SyntaxKind`.
+ *
+ * Used for one-level traversal of meaningful AST children.
+ *
+ * @param node - The parent `SyntaxNode`.
+ * @param action - Function to apply to each relevant child node.
+ */
 export function forEachNode(
   node: SyntaxNode,
   action: (node: SyntaxNode) => void,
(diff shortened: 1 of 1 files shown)


packages/language/src/syntax-tree/ast-iterator.ts
```diff
@@ -931,3 +931,53 @@ export function forEachNode(
       break;
   }
 }
+
+export function traverseAllNodes(node: SyntaxNode, traverse: (n: SyntaxNode) => void) {
+  traverse(node);
+  forEachNode(node, (child) => {
+    traverseAllNodes(child as SyntaxNode, traverse);
+  });
+}
```
msujew: Suggestion: We should be able to return from this traversal early in case it is no longer required, like is the case in `containsNodeOfKind`. For example via something like this:
```suggestion
export enum TraversalState {
  Continue,
  Stop,
}

export function traverseAllNodes(node: SyntaxNode, traverse: (n: SyntaxNode) => TraversalState | undefined | void) {
  const state = traverse(node);
  if (state !== TraversalState.Stop) {
    forEachNode(node, (child) => {
      traverseAllNodes(child as SyntaxNode, traverse);
    });
  }
}
```

packages/language/src/syntax-tree/ast-iterator.ts
```diff
@@ -931,3 +931,53 @@ export function forEachNode(
...
+ * Checks whether the given syntax tree contains at least one node of the specified kind.
+ *
+ * @param root - The root node to start searching from.
+ * @param kind - The syntax kind to search for.
+ * @returns True if a matching node is found, false otherwise.
+ */
+export function containsNodeOfKind(
+  root: SyntaxNode,
+  kind: SyntaxKind,
+): boolean {
```
msujew: Suggestion: Can we make this into a `find` instead? That is much more useful than the binary info whether something exists, even though it requires the same amount of work.
```suggestion
export function findFirstNode<T>(
  root: SyntaxNode,
  kind: SyntaxKind,
): T | undefined {
```

packages/language/src/syntax-tree/ast-iterator.ts
```diff
@@ -931,3 +931,53 @@ export function forEachNode(
...
+  kind: SyntaxKind,
+): T[] {
+  const matches: T[] = [];
+  traverseAllNodes(root, (n) => {
+    if (n.kind === kind) {
+      matches.push(n as T);
+    }
+  });
+  return matches;
+}
```
msujew, line 1010: This is good 👍

msujew (commented):
Thanks, makes a lot of sense to add this traversal function!

I have a few comments for performance reasons, see below.

packages/language/src/syntax-tree/ast-iterator.ts
Wagner-Laranjeiras: Sure thing!

msujew (approved):
Thanks, looks good to me 👍

PR #380, opened 2025-09-02
Use async mechanism to load files

msujew:
This PR features a lot of changes, but mostly to function signatures (due to moving from sync to async). Main changes are concerned with:

* All `FileSystemProvider` methods are now async
* Adds a `FileStore` API that stores the text documents used for `INCLUDE` statements.

This change is in preparation for loading files through the VS Code API, which will be eventually required for the web extension to work correctly.


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -349,19 +349,56 @@ export function doRunInstructions(
...
+  } catch (err) {
+    handleInstructionError(err, context);
+  }
+  if (node.instruction.kind === inst.InstructionKind.Halt) {
+    return undefined; // Stop execution
+  }
+  return result;
+}
+
+function doRunInstructionsSync(
```
ssmifi, line 379: _Question_: Is it worth to duplicate the code for the `sync` functions and manage two "modes"? Couldn't we make all functions async?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -404,11 +459,9 @@ function runInstruction(
     case inst.InstructionKind.Do:
       return runDoInstruction(instruction, context, node);
     case inst.InstructionKind.Include:
-      runIncludeInstruction(instruction, context);
-      break;
+      throw new Error("INCLUDE instruction cannot be run in synchronous mode.");
     case inst.InstructionKind.Inscan:
-      runInscanInstruction(instruction, context);
-      break;
+      throw new Error("INSCAN instruction cannot be run in synchronous mode.");
```
ssmifi, line 471: _Comment_: It is getting harder to keep track of what is allowed in what context. For example, is an include allowed in an `Compound`? Apparently not (anymore), but that it not obvious from the naming.

packages/language/src/preprocessor/pli-lexer.ts
```diff
@@ -148,12 +152,7 @@ function generateIncAfterInstruction(
   // It allows to include ONE SINGLE file. Afterwards the preprocessor runs as normal.
   const instruction: InstructionNode = {
     labels: [],
-    instruction: createIncludeInstruction(
-      includeItem,
-      incAfter.process,
-      false,
-      incAfter.token,
```
ssmifi, line 156: _Question_: `The incAfter.token` was provided for the diagnostics, right? What happened to this?

packages/language/test/fourslash-harness/harness.test.ts
```diff
@@ -1,238 +0,0 @@
...
-    expect(mainFile?.content).toBe(`WRAP_BEGIN MY CONTENT WRAP_END`);
-  });
-
-  test("should parse a wrapper file", () => {
-    const wrapFileContent = `////WRAP_BEGIN <...> WRAP_END`;
-    const wrapper = parseWrapperFile(wrapFileContent);
-
-    expect(wrapper.wrap("MY CONTENT")).toBe("WRAP_BEGIN MY CONTENT WRAP_END");
-  });
-});
```
ssmifi: _Question_: Why was this removed?

packages/language/test/fourslash-harness/implementation/test-builder.ts
```diff
@@ -62,19 +62,19 @@ export function createTestBuilderHarnessImplementation(
         testBuilder.expectCompletions(label.toString(), content),
     },
     hover: {
-      expectIncludeAt(label, filePath, markdown) {
-        testBuilder.expectHover(label.toString(), {
+      async expectIncludeAt(label, filePath, markdown) {
```
ssmifi: _Minor_: Should be defined like the other ones (`=>`).

packages/language/test/fourslash/hover/copybook.mts
```diff
@@ -20,7 +20,7 @@
 
 //// %include <|1>"lib.pli";
 
-hover.expectIncludeAt(
+await hover.expectIncludeAt(
```
ssmifi: _Question_: Now, one has to know explicitly which ones are async and which not. Is there a way to let the harness handle this? Or should we at least indicate this by naming them differently?

packages/language/test/lsp/definition-requests.test.ts
```diff
@@ -1,83 +0,0 @@
...
-    const main = ` %INCLUDE "include.pli";
- <|1>Y = 42;
- <|1>Y = 45;`;
-
-    createFsTestBuilder([
-      { uri: "file:///main.pli", content: main },
-      { uri: "file:///include.pli", content: include },
-    ]).expectLinks();
-  });
-});
```
ssmifi, line 83: _Question_: Why was this removed?

packages/language/test/parser/pli-preprocessor-include.test.ts
```diff
@@ -1,217 +0,0 @@
...
-    PluginConfigurationProviderInstance.setProgramConfigs("", []);
-    PluginConfigurationProviderInstance.setProcessGroupConfigs([]);
-  });
-
-  // Tests have been ported to the fourslash framework
-  // But in the future, we might need more involved tests here
-  test("Stub", () => {
-    tokenize.toString();
-  });
-});
```
ssmifi, line 217: _Question_: Why was this removed?

packages/language/test/test-builder.ts
```diff
@@ -165,10 +165,12 @@ export class TestBuilder {
         replaceNamedIndicesWithDocument(file),
       ]),
     );
+  }
 
+  private async init(options?: TestBuilderOptions) {
```
ssmifi: _Minor_: Having the options set here and in the constructor seems to be unnecessary.

ssmifi (commented):
Looks good. Perhaps you can help me and answer the questions so that I can fully understand all implications. Thank you!

packages/language/src/preprocessor/instruction-interpreter.ts
msujew: We could, but that adds a lot of noise and slows down the instruction interpreter considerably. It also makes debugging much more difficult, because we cannot go through the call stack in async code.

packages/language/src/preprocessor/instruction-interpreter.ts
msujew: I've added a comment explaining which are allowed/which aren't.

packages/language/src/preprocessor/pli-lexer.ts
msujew: The `incAfter` token is still included in the `includeItem`.

packages/language/test/fourslash-harness/harness.test.ts
msujew: Reinstated.

packages/language/test/lsp/definition-requests.test.ts
msujew: We are now doing all of our linking tests via the fourslash framework (and these tests are all already covered there). These tests are now obsolete.

packages/language/test/parser/pli-preprocessor-include.test.ts
msujew: We are now doing all of our linking tests via the fourslash framework (and these tests are all already covered there). These tests are now obsolete.

msujew:
@ssmifi FYI I completely refactored this PRs approach to async stuff in the LSP code - it's now sync again, because its using the new `FileStore`. This alleviates likely a bunch of issues (especially in the testing code).

packages/language/test/fourslash/hover/copybook.mts
msujew: I've reverted this to be sync again.

packages/language/src/preprocessor/instruction-interpreter.ts
ssmifi: Where? Or did you mean the `throw new Error`s? Then again, one can look it up here in `runInstructionSync` once familiar with the code.

ssmifi (approved):
LGTM. Thanks for the clarifications and the new `FileStore`. 👍

packages/language/src/preprocessor/instruction-interpreter.ts
msujew: Oops, seems like I didn't push that change, the comment should be up now.

PR #381, opened 2025-09-02
Extend compiler options and tests #Rules

ssmifi:
Based on https://github.com/zowe/zowe-pli-language-support/pull/378

- Implemented compiler options `rules`.
- `expectCompilerOptions` can test for partial objects.

Also improved some diagnostics that were reported in https://github.com/zowe/zowe-pli-language-support/issues/374 f: 
- Fixed `undefined` in `MAXGEN`, `MAXBRANCH`.
- Added abbreviations for `INSOURCE`.
- Improved reporting in `INCAFTER` and `GONUMBER`.
- Added test for `INCAFTER`.

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/codes.ts
+++ b/packages/language/src/preprocessor/compiler-options/codes.ts
@@ -84,13 +84,19 @@ export const CompilerOptionsCodes = {
   ExpectedNumberRange: {
     code: "COOP09",
     severity: Severity.W,
-    message: (number: number, min: number, max: number) => {
+    message: (
+      number: number,
+      min: number | undefined,
+      max: number | undefined,
+    ) => {
       if (min !== undefined && max !== undefined) {
         return `Expected a number between ${min} and ${max}, but received ${number}.`;
-      } else if (min) {
+      } else if (min !== undefined) {
         return `Expected a number greater than or equal to ${min}, but received ${number}.`;
-      } else {
+      } else if (max !== undefined) {
         return `Expected a number less than or equal to ${max}, but received ${number}.`;
+      } else {
+        throw new Error("At least one of min or max must be defined");
       }
     },
     fullCode: "COOP09W",
@@ -140,6 +146,16 @@ export const CompilerOptionsCodes = {
       message: (value: string) =>
         `Expected "ASSERT", "DISPLAY" or "PUT", but received '${value}'.`,
       fullCode: "COIG01W",
+    } as ParametricPLICode,
+  },
+
+  IncAfter: {
+    InvalidParameter: {
+      code: "COIA01",
+      severity: Severity.W,
+      message: (value: string) =>
+        `Expected "PROCESS" option with a process name, but received '${value}'.`,
+      fullCode: "COIA01W",
     } as ParametricPLICode,
   },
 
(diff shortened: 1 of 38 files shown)


packages/language/src/preprocessor/compiler-options/codes.ts
```diff
@@ -87,7 +87,7 @@ export const CompilerOptionsCodes = {
     message: (number: number, min: number, max: number) => {
       if (min !== undefined && max !== undefined) {
         return `Expected a number between ${min} and ${max}, but received ${number}.`;
-      } else if (min) {
+      } else if (min !== undefined) {
```
msujew, line 94: Suggestion: Something is wrong here - either `min` can be `undefined`, or the calling side has some sort of type checking error.

packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -2792,6 +2804,597 @@ translator.rule(["RTCHECK"], (option, options) => {
...
+    options.rules = getDefaultCompilerOptions().rules;
+  }
+  for (const value of option.values) {
+    if (value.kind === SyntaxKind.CompilerOptionText) {
+      if (value.value.length === 0) {
+        throw TranslationError.fromCode(
+          value.token,
+          CompilerOptionsCodes.ExpectedPlainNotEmpty,
+        );
+      }
```
msujew, line 2820: Question: With the way this currently works, when we encounter one invalid `RULES` sub-option, the whole parsing of the `RULES` compiler option stops with the error being thrown. Do we want to change this, so that we can continue parsing somehow? The error should obviously still be reported.

packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -2792,6 +2804,597 @@ translator.rule(["RTCHECK"], (option, options) => {
...
+      const name = value.value.toUpperCase();
+      switch (name) {
+        case "IBM":
+          options.rules!.ibm = "IBM";
```
msujew, line 2825: Suggestion: Can we somehow get rid of the non-null assertion? Perhaps the comment above should help with that as well.

msujew (commented):
Looks mostly, good. Just a few comments.

packages/language/src/preprocessor/compiler-options/translator.ts
ssmifi: True and I agree, especially with reporting of dupe/mutex suboptions. But I would postpone this and move it into the PR that is going to solve https://github.com/zowe/zowe-pli-language-support/issues/374 b, if you agree.

msujew: Sure, fine with me 👍

packages/language/src/preprocessor/compiler-options/translator.ts
ssmifi: I already implemented `ensureToBeDefined` in the follow-up PR https://github.com/zowe/zowe-pli-language-support/pull/388 to get rid of these assertions. This pops up at other places, too. I would refactor this after we pulled that one in, if that is ok with you. I'll leave a todo.

msujew: Fine with me 👍

msujew (approved):
LGTM, thanks 👍

Issue #383, opened 2025-09-02
Implement support for IBM2412I/IBM2410I/IBM2409I


Wagner-Laranjeiras:
- [x] IBM2410I
- [x] IBM2412I
- [x] IBM2409

**Refer to:**

https://github.com/zowe/zowe-pli-language-support/blob/cbd0f4f321a5db83fcadd028160164cafdcacab2/packages/language/src/validation/messages/pli-codes.ts#L9009-L9020

also

https://github.com/zowe/zowe-pli-language-support/blob/cbd0f4f321a5db83fcadd028160164cafdcacab2/packages/language/src/validation/messages/pli-codes.ts#L9035-L9051

and

https://github.com/zowe/zowe-pli-language-support/blob/cbd0f4f321a5db83fcadd028160164cafdcacab2/packages/language/src/validation/messages/pli-codes.ts#L8993-L9007

PR #384, opened 2025-09-02
Implement support for IBM2412I

Wagner-Laranjeiras:
**Implement IBM2412I: Procedure with RETURN must specify RETURNS attribute**
Issue: #383 

**Summary**
This PR adds a validator that implements IBM2412I: Inform when a procedure contains a RETURN statement but does not declare a RETURNS attribute on its PROCEDURE statement. The check inspects the procedure’s options for a ReturnsOption and searches for the presence of any ReturnStatement in the body. If a RETURN is found without a corresponding RETURNS attribute, an error is reported.

**Changes**
IBM2412I_proc_with_return_stmt_needs_returns_att: validation code implementing the rule described above.
pli-validator: register the new validation code.
AST and parser: add "procToken" property to ProcedureStatement.
Tests: add fourslash tests under packages/language/test/fourslash/validate/IBM2412I/

### DIFF
--- a/packages/language/src/validation/messages/error-severity/IBM2412I-proc-with-return-stmt-needs-returns-att.ts
+++ b/packages/language/src/validation/messages/error-severity/IBM2412I-proc-with-return-stmt-needs-returns-att.ts
@@ -1,0 +1,54 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { ValidationAcceptor } from "../../validator";
+import * as AST from "../../../syntax-tree/ast";
+import {
+  Severity,
+  tokenToRange,
+  tokenToUri,
+} from "../../../language-server/types";
+import * as PLICodes from "./../pli-codes";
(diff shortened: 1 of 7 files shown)


packages/language/src/validation/messages/error-severity/IBM2412I-proc-with-return-stmt-needs-returns-att.ts
```diff
@@ -0,0 +1,54 @@
...
+export function IBM2412I_proc_with_return_stmt_needs_returns_att(
+  node: AST.ProcedureStatement,
+  acceptor: ValidationAcceptor,
+): void {
+  const hasReturnsAtt = node.options?.some(
+    (att) => att.kind === AST.SyntaxKind.ReturnsOption,
+  );
+  if (hasReturnsAtt) return;
+
+  const returnStmt = findFirstNodeOfKind(node, AST.SyntaxKind.ReturnStatement);
```
msujew, line 39: Suggestion: This will result in false positives for nested procedures:

```
OUTER: PROC;
  INNER: PROC RETURNS(FIXED);
    RETURN (0);
  END INNER;
END OUTER;
```

Please also add a test for this.

packages/language/test/fourslash/validate/IBM2412I/1.ts
msujew, line 1: Please give these files more descriptive names.

msujew (commented):
Do you plan to combine the new validation with `IBM2410I`. Most of the logic is the same (just inverted).

Issue #372, opened 2025-08-28
Implement missing preprocessor statements


msujew:
Preprocessor statements

<img width="134" height="511" alt="Image" src="https://github.com/user-attachments/assets/efa927b5-2ccc-42ec-94f5-abc1b9070772" />

There are a few last missing preprocessor statements that we should implement to fully support the PLI macro preprocessor:

- [x] [ANSWER](https://www.ibm.com/docs/en/epfz/6.1.0?topic=procedures-preprocessor-answer-statement) (procedure only)
- [x] [CALL](https://www.ibm.com/docs/en/epfz/6.1.0?topic=procedures-preprocessor-call-statement) (procedure only)
- [x] [NOTE](https://www.ibm.com/docs/en/epfz/6.1.0?topic=statements-note-statement)
- [x] [SELECT](https://www.ibm.com/docs/en/epfz/6.1.0?topic=statements-select-statement)
- [x] [REPLACE](https://www.ibm.com/docs/en/epfz/6.1.0?topic=statements-replace-statement) (non-procedure only)

Lotes:
I will continue on ANSWER.

In the first run, I wanted to start on CALL, but it would be nice to have ANSWER as kind of "printf" for CALL.

msujew:
@danieldietrich As for the special LSP semantics of `%REPLACE`:

> The %REPLACE statement allows for the immediate replacement of a name with a string constant or a numeric constant. The name does not need to be a declared variable to have a value assigned to it.

Let's look at an example:

```
%REPLACE X WITH 42;
DCL Y FIXED INIT(X);
```

Should result in the following output after running the preprocessor:

```
DCL Y FIXED INIT(42);
```

Therefore, `X` just acts as another `variables` value in the `InterpreterContext` and should follow the same semantics as those variables when it comes to replacement inside of `TokenStatement` elements. The scoping/linking will be taken care of by itself, if you pass the `declarationNode` to the variable produced by the `REPLACE` statement.

The runtime semantics, might differ slightly from the above, but right now we can't check against the compiler - once we can, we can always update the interpreter to align the behavior with the compiler.

Lotes:
ANSWER was added with PR #404 .

Will take care of CALL now.

msujew:
All preprocessor statements are now supported.

PR #385, opened 2025-09-03
Add preprocessor statement %NOTE

Lotes:
A small statement for the beginning.

Refers to #372 , but will not handle all missing statements. Only `%NOTE`.

Having small PRs to keep reading time low :D ...

Questions from my side:

1. I was not so sure about CODE / SEVERITY fields for lexer errors and diagnostics. Are they fine?
2. There is also a TODO about a wrong code, what shall happen then? Do you have an opinion?

### DIFF
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -218,10 +218,7 @@ function generateNoteInstruction(
     ? {
         kind: inst.InstructionKind.Note,
         noteToken: node.noteToken,
-        message: message || {
-          kind: inst.InstructionKind.String,
-          value: "",
-        },
+        message,
         code,
       }
     : undefined;
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -204,13 +204,21 @@ function generateTokenInstruction(
   };
 }
 
-function generateNoteInstruction(node: ast.NoteDirective): inst.NoteInstruction {
+function generateNoteInstruction(
+  node: ast.NoteDirective,
+): inst.NoteInstruction {
   return {
     kind: inst.InstructionKind.Note,
     noteToken: node.noteToken!,
-    message: generateExpressionInstruction(node.message) || { kind: inst.InstructionKind.String, value: "" },
-    code: generateExpressionInstruction(node.code) || { kind: inst.InstructionKind.Number, value: "0" },
-  }
+    message: generateExpressionInstruction(node.message) || {
+      kind: inst.InstructionKind.String,
+      value: "",
+    },
+    code: generateExpressionInstruction(node.code) || {
+      kind: inst.InstructionKind.Number,
+      value: "0",
+    },
+  };
 }
 
 function generateActivateInstruction(
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -214,10 +214,10 @@ function generateNoteInstruction(
 ): inst.NoteInstruction | undefined {
   const message = generateExpressionInstruction(node.message);
   const code = generateExpressionInstruction(node.code) || DefaultCode;
-  return message
+  return message && node.noteToken
     ? {
         kind: inst.InstructionKind.Note,
-        noteToken: node.noteToken || undefined,
+        noteToken: node.noteToken,
         message: message || {
           kind: inst.InstructionKind.String,
           value: "",
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -209,7 +209,7 @@ function generateNoteInstruction(
 ): inst.NoteInstruction {
   return {
     kind: inst.InstructionKind.Note,
-    noteToken: node.noteToken,
+    noteToken: node.noteToken || undefined,
     message: generateExpressionInstruction(node.message) || {
       kind: inst.InstructionKind.String,
       value: "",
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -204,21 +204,27 @@ function generateTokenInstruction(
   };
 }
 
+const DefaultCode: inst.ExpressionInstruction = {
+  kind: inst.InstructionKind.Number,
+  value: "0",
+};
+
 function generateNoteInstruction(
   node: ast.NoteDirective,
-): inst.NoteInstruction {
-  return {
-    kind: inst.InstructionKind.Note,
-    noteToken: node.noteToken || undefined,
-    message: generateExpressionInstruction(node.message) || {
-      kind: inst.InstructionKind.String,
-      value: "",
-    },
-    code: generateExpressionInstruction(node.code) || {
-      kind: inst.InstructionKind.Number,
-      value: "0",
-    },
-  };
+): inst.NoteInstruction | undefined {
+  const message = generateExpressionInstruction(node.message);
+  const code = generateExpressionInstruction(node.code) || DefaultCode;
+  return message
+    ? {
+        kind: inst.InstructionKind.Note,
+        noteToken: node.noteToken || undefined,
+        message: message || {
+          kind: inst.InstructionKind.String,
+          value: "",
+        },
+        code,
+      }
+    : undefined;
 }
 
 function generateActivateInstruction(
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -209,7 +209,7 @@ function generateNoteInstruction(
 ): inst.NoteInstruction {
   return {
     kind: inst.InstructionKind.Note,
-    noteToken: node.noteToken!,
+    noteToken: node.noteToken,
     message: generateExpressionInstruction(node.message) || {
       kind: inst.InstructionKind.String,
       value: "",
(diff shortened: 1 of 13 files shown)


packages/language/src/preprocessor/instruction-generator.ts
```diff
@@ -196,6 +199,23 @@ function generateTokenInstruction(
   };
 }
 
+function generateNoteInstruction(
+  node: ast.NoteDirective,
+): inst.NoteInstruction {
+  return {
+    kind: inst.InstructionKind.Note,
+    noteToken: node.noteToken!,
```
msujew: Suggestion: Do not use non-null assertions here.

msujew (commented):
Looks pretty good! Just one issue, see below. Note that we might want to merge https://github.com/zowe/zowe-pli-language-support/pull/386 first, which improves the dev experience when working with diagnostics (especially in the preprocessor).

packages/language/src/preprocessor/instruction-generator.ts
Lotes: The `noteToken` comes from a `consume` call inside the parser. Actually it should be set. It would be better to remove the NULL from the type instead, but I am not sure what ist good at this place. `undefined!` during initialization of the AST node is also bad???

msujew: Yeah, I would not do that. Please just make the `noteToken` on the instruction optional. In some cases, we want to generate synthetic AST nodes, and having mandatory properties on there makes this rather difficult.

Alternatively, you can also generate `undefined` in case it doesn't make sense to generate a `NoteInstruction` without the token.

See as an example: https://github.com/zowe/zowe-pli-language-support/blob/bc20bad078048ee1eef165a974e03a81d769f9b3/packages/language/src/preprocessor/pli-lexer.ts#L140-L165

packages/language/src/preprocessor/preprocessor-parser.ts
```diff
@@ -1194,6 +1201,42 @@ function memberCall(
...
 }
 
+function noteStatement(state: PreprocessorParserState): ast.NoteDirective {
+  const note = ast.createNoteDirective();
+  const noteToken = state.consume(
+    note,
+    CstNodeKind.NoteDirective_PercentNOTE,
+    PreprocessorTokens.Note,
+  );
+  note.noteToken = noteToken || null; // noteToken could be synthetic
```
msujew: Question: Why the `|| null`? Is the token returned by `state.consume` ever falsy?

packages/language/src/preprocessor/instruction-generator.ts
Lotes: I made it optional now. Thanks

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -478,13 +479,55 @@ function runInstructionSync(
...
+  const severity = codeToSeverity(code);
+  context.errors.push(
+    new PreprocessorError(
+      message ?? "",
+      instruction.noteToken,
+      undefined,
+      code?.toString(),
+      severity,
+    ),
+  );
```
msujew: Suggestion: I'm wondering whether we should actually push an empty message to the diagnostics. That would look pretty weird in the UI to be honest.

packages/language/src/preprocessor/instruction-generator.ts
```diff
@@ -201,6 +204,23 @@ function generateTokenInstruction(
...
+    kind: inst.InstructionKind.Note,
+    noteToken: node.noteToken || undefined,
+    message: generateExpressionInstruction(node.message) || {
+      kind: inst.InstructionKind.String,
+      value: "",
+    },
+    code: generateExpressionInstruction(node.code) || {
+      kind: inst.InstructionKind.Number,
+      value: "0",
+    },
```
msujew: Suggestion: With the point above, do we simply want to return `undefined` in case the `message` instruction is not valid (or the `noteToken` is not set)?

packages/language/src/preprocessor/preprocessor-parser.ts
Lotes: True, I mixed up the wrong things in my head.

packages/language/src/preprocessor/instruction-interpreter.ts
Lotes: Yes! I can check whether the message is empty

packages/language/src/preprocessor/instruction-generator.ts
Lotes: I made also a change here. Please have another look.

packages/language/src/preprocessor/instruction-generator.ts
```diff
@@ -201,6 +204,29 @@ function generateTokenInstruction(
...
+  const message = generateExpressionInstruction(node.message);
+  const code = generateExpressionInstruction(node.code) || DefaultCode;
+  return message && node.noteToken
+    ? {
+        kind: inst.InstructionKind.Note,
+        noteToken: node.noteToken,
+        message: message || {
+          kind: inst.InstructionKind.String,
+          value: "",
+        },
```
msujew: Minor: We already know that the message is truthy.
```suggestion
        message,
```

packages/language/src/validation/validator.ts
```diff
@@ -143,6 +143,7 @@ export function lexerIssuesToDiagnostics(
         severity: error.severity,
         range: error.range,
         message: error.message,
+        source: "lexer",
```
msujew: Suggestion: Please revert this, see https://github.com/zowe/zowe-pli-language-support/pull/386#discussion_r2321552842 for an explanation.

msujew (approved):
Looks good, however let me merge https://github.com/zowe/zowe-pli-language-support/pull/386 before this.

packages/language/src/validation/validator.ts
Lotes: Need to rebase...

Lotes:
@msujew After the rebase I removed the code field for the diagnostics which was coming from the code field of the note directive. It is just mapped to the severity and the code itself seems not to be very helpful as a code for a diagnostic (since it is some value between 0 and 16, whcih is effectively the severity that we already have).

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -473,13 +474,62 @@ function runInstructionSync(
...
+    case 8:
+      return Severity.E;
+    case 12:
+      return Severity.S;
+    case 16:
+      return Severity.U;
+    default:
+      return undefined;
+  }
+}
```
msujew: I just noticed, that the code not only emits a specific severity, but also a specific PLI message code. For example `0` emits `IBM3024I` and `4` emits `IBM3259I`. Can you return the whole code here instead?

Note that `IBM3999I` seems to have an incorrect `message` return value.

msujew (changes requested):
Just noticed something that we should probably address before merging 👍

packages/language/src/preprocessor/instruction-interpreter.ts
Lotes: I removed `codeToSeverity` to reduce complexity. For the case when the code is invalid, I could not find any PL/I code...

Lotes:
There is a [%NOTE statement](https://www.ibm.com/docs/en/epfz/6.2.0?topic=statements-note-statement) and a [%NOTE directive](https://www.ibm.com/docs/en/epfz/6.2.0?topic=directives-note-directive). I will use the "NOTE statement" documentation here, because it is more specific and has no gaps for invalid CODEs.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -473,13 +479,82 @@ function runInstructionSync(
...
+  context: InterpreterContext,
+): void {
+  const message = valueToString(
+    evaluateExpression(instruction.message, context),
+  );
+  const codeValue = evaluateExpression(instruction.code, context);
+  const code = valueToNumber(codeValue);
+  let error: Diagnostic = undefined!;
+  if (message) {
+    // Attention! This implementation follows the documentation of "%NOTE statement", not of "%NOTE directive"
```
msujew: Do you mind adding a link here? Without knowing the exact phrasing of the language reference, there's no way to know what the difference between those is supposed to be :D

packages/language/test/fourslash/preprocessor/simple-note-unknown-code.ts
```diff
@@ -0,0 +1,23 @@
...
+//// %<|1:NOTE|>("This is unknown!", 100);
+
+preprocessor.expectTokens("");
+verify.expectDiagnosticsAt(1, [
+  {
+    severity: 4,
+    code: "IBM1941IU",
+    message: "This is unknown!",
+  },
+]);
```
msujew: Suggestion: We can make this a single object. Same for the other tests.
```suggestion
verify.expectDiagnosticsAt(1, {
  severity: 4,
  code: "IBM1941IU",
  message: "This is unknown!",
});
```

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -473,13 +479,82 @@ function runInstructionSync(
...
+function runNoteInstruction(
+  instruction: inst.NoteInstruction,
+  context: InterpreterContext,
+): void {
+  const message = valueToString(
+    evaluateExpression(instruction.message, context),
+  );
+  const codeValue = evaluateExpression(instruction.code, context);
+  const code = valueToNumber(codeValue);
+  let error: Diagnostic = undefined!;
```
msujew: I don' think this is necessary.

```suggestion
  let error: Diagnostic;
```

msujew (approved):
Thanks, LGTM. Please see below for a few suggestions.

PR #386, opened 2025-09-03
Streamline diagnostics in the language server

msujew:
Streamlines the use of diagnostics (especially throughout the tokenization infrastructure) and makes it easier to generate "invalid" diagnostics which will be filtered out later before being sent to the language client.


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1357,21 +1346,6 @@ function resolveIncludeFileUri(
...
-      !pgroup["include-extensions"])
-  ) {
-    const msg = pgroup["include-extensions"]?.length
-      ? `expected one of: ${pgroup["include-extensions"]?.join(", ")}`
-      : `expected no extension`;
-    throw new PreprocessorError(
-      `Unsupported include extension for included file, '${item.fileName}', ${msg}`,
-      item.token,
-    );
-  }
```
msujew, line 1764: Note that the compiler will simply emit a normal "file cannot be opened" error in this case. We do the same now.

packages/language/src/language-server/types.ts
```diff
@@ -109,14 +113,90 @@ export function severityToLsp(severity: Severity): lsp.DiagnosticSeverity {
...
-  range: Range;
+  uri?: string;
+  range?: Range;
   message: string;
   code?: string;
   data?: any;
   source?: string;
 }
 
+export function diagnosticFromCode(
```
Lotes, line 124: Nice function!

Lotes: But I guess the parameters are without names for autocompletion, right?
Is there a chance to get those? Like: `diagnosticFromCode(PLICodes.Error.IBM3581I, argTokens.nameToken)(argTokens.nameToken.image)`? Might look ugly, and maybe it goes also without second parentheses. I think you could infer the message type with a given IBM code. Have you tried it out?

Lotes: Or are those incomplete messages the INVALID ones you were talking about?

Lotes (commented):
Simple and understandable from my point of view. I was just wondering about the `args` and the IntelliSense/hint for parameter types. Look at the comment :)

packages/language/src/language-server/types.ts
msujew: > I think you could infer the message type with a given IBM code. Have you tried it out?

We could, but unfortunately, all the type information about the exact message is erased right now, since it is always casted to `as ParametricPLICode`.

Lotes: Ah I see it was already gone before :-(, ok nevermind.

packages/language/src/language-server/types.ts
```diff
@@ -109,14 +113,90 @@ export function severityToLsp(severity: Severity): lsp.DiagnosticSeverity {
...
+
+export function diagnostic(
+  severity: Severity,
+  message: string,
+  token: Token | null | undefined,
+): Diagnostic;
+export function diagnostic(
+  severity: Severity,
+  message: string,
+  range: Range | null | undefined,
```
Lotes, line 174: Why NULL and UNDEFINED? Also for URI and TOKEN?

packages/language/src/language-server/types.ts
```diff
@@ -138,28 +221,31 @@ export function diagnosticsToLSP(
...
-          },
+      start: offsetToPosition(doc, diagnostic.range.start),
+      end: offsetToPosition(doc, diagnostic.range.end),
     },
     message: diagnostic.message,
     code: diagnostic.code,
     data: diagnostic.data,
-    source: diagnostic.source ?? "pli",
+    source: diagnostic.source ?? "PL/I",
   };
```
Lotes, line 245: I was wondering here, what task this string has to fulfill? Is it meant as a language ID? Where is it used or displayed? In the context of PL/I we also have "lexer" as source I think. So, "PL/I" or "pli" are just fallback string to be displayed later? Or do they need to point to something like an ID would do.

packages/language/src/language-server/types.ts
msujew: Because in some cases, the ranges/URIs are simply unavailable (mainly due to the preprocessor or invalid AST). Instead of checking for them **everywhere**, we simply accept those as valid values and filter them out at the end before sending them to the user.

packages/language/src/language-server/types.ts
msujew: The `source` is shown in the UI and presented to the user:

image

I've removed `lexer` from the `source` a while ago, because the `source` is supposed to be used to inform the user about which extension is sourcing the diagnostic. We use the `code` for more specific information.

Lotes (approved):
LGTM.

Issue #375, opened 2025-08-28
`INSCAN` Hover Support


asatklichov:
Currently, we are supporting hover-over on %INCLUDE ABC;

And also, we need a hover-over support for below preprocessors.

%INSCAN ALI;
%XINSCAN VELI;
%XINCLUDE EDA;

PR #387, opened 2025-09-03
Support hover and definition request for INSCAN

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/375

Note that this change supports the hover and definition request on the `INSCAN` token, not the token after the keyword. The token after the keyword is always a variable, which has its own hover/definition result.

### DIFF
--- a/packages/language/src/language-server/hover-request.ts
+++ b/packages/language/src/language-server/hover-request.ts
@@ -242,12 +242,14 @@ interface IncludeItemNode {
 function getIncludeItemRepresentation(
   unit: CompilationUnit,
   node: IncludeItemNode,
-  include: boolean,
+  type: string,
 ): string | null {
   if (!node.filePath || !node.relativeFilePath) {
     return null;
   }
 
+  // TODO: Don't store the file content in the node itself
+  // Instead, implement a proper caching mechanism
   let partialContent = node.sourceText;
   const fileUri = URI.parse(node.filePath);
 
@@ -270,7 +272,15 @@ function getIncludeItemRepresentation(
     // cache for later requests
     node.sourceText = partialContent;
   }
-  return `${formatPliCodeBlock(`%${include ? "INCLUDE" : "INSCAN"} "${node.relativeFilePath}"`)}\n\n---\n${formatPliCodeBlock(partialContent)}`;
+  return generateIncludeItemMarkup(type, node.relativeFilePath, partialContent);
+}
+
+export function generateIncludeItemMarkup(
+  type: string,
+  relativePath: string,
+  sourceText: string,
+): string {
+  return `${formatPliCodeBlock(`${type} "${relativePath}"`)}\n\n---\n${formatPliCodeBlock(sourceText)}`;
 }
 
 /**
@@ -286,9 +296,17 @@ function getNodeRepresentation(
     case SyntaxKind.LabelPrefix:
       return getLabelPrefixRepresentation(node);
     case SyntaxKind.IncludeItem:
-      return getIncludeItemRepresentation(unit, node, true);
+      let type = "%INCLUDE";
+      const ppInclude = unit.compilerOptions?.pp?.ppInclude?.value;
+      if (
+        node.container?.kind === SyntaxKind.IncludeAltDirective &&
+        ppInclude
+      ) {
+        type = ppInclude;
+      }
+      return getIncludeItemRepresentation(unit, node, type);
     case SyntaxKind.InscanDirective:
-      return getIncludeItemRepresentation(unit, node, false);
+      return getIncludeItemRepresentation(unit, node, "%INSCAN");
     default:
       return null;
   }
--- a/packages/language/src/language-server/hover-request.ts
+++ b/packages/language/src/language-server/hover-request.ts
@@ -25,7 +25,6 @@ import {
   DimensionBound,
   Dimensions,
   Expression,
-  IncludeItem,
   LabelPrefix,
   ProcedureStatement,
   SyntaxKind,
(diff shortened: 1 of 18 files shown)


packages/language/src/language-server/hover-request.ts
```diff
@@ -264,7 +270,7 @@ function getIncludeItemRepresentation(
     // cache for later requests
     node.sourceText = partialContent;
```
ssmifi, line 273: _Question_: Does this actually change the content of the AST node? Couldn't this influence other parts of the LS (later on)? Should we maybe cache this by other means?

_Edit_: Since this is actually set here and is basically the cache, maybe we should rename the property to reflect this.

packages/language/src/syntax-tree/ast.ts
```diff
@@ -1518,8 +1518,10 @@ export interface IncludeItem extends AstNode {
   fileName: string | null;
   string: boolean;
   ddname: string | null;
-  filePath: string | null;
   token: Token | null;
+  // Properties filled by the preprocessor
```
ssmifi, line 1522: _Minor_: Actually, it is set in the `hover-request.ts`.

packages/language/test/fourslash-harness/implementation/test-builder.ts
```diff
@@ -65,7 +64,13 @@ export function createTestBuilderHarnessImplementation(
...
         testBuilder.expectHover(label.toString(), {
           kind: MarkupKind.Markdown,
-          value: `%INCLUDE "${URI.parse(filePath).fsPath}"\n\n---\n${markdown}`,
+          value: `${formatPliCodeBlock(`%INCLUDE "${filePath}"`)}\n\n---\n${markdown}`,
+        });
+      },
+      expectInscanAt(label, filePath, markdown) {
+        testBuilder.expectHover(label.toString(), {
+          kind: MarkupKind.Markdown,
+          value: `${formatPliCodeBlock(`%INSCAN "${filePath}"`)}\n\n---\n${markdown}`,
```
ssmifi: _Comment_: Not a high priority right now, but if we intent to add more markdowns in the future, we should probably add some kind of store similar to the error codes to change the messages in one place.

packages/language/test/fourslash-harness/implementation/test-builder.ts
```diff
@@ -65,7 +64,13 @@ export function createTestBuilderHarnessImplementation(
...
         });
```
ssmifi: _Question_: Actually, personally, I would prefer to to not encode specific messages into the harness interface, but use the same pattern as for the validations like validation + specific message in the test itself. So, for the hover it would be, e.g. something like `hover.expectHover(1, lsMessages.HoverRequest.message("INCLUDE", filePath, markdown);`. Wdyt?

packages/language/test/test-builder.ts
```diff
@@ -746,11 +746,11 @@ export class TestBuilder {
     return this.createPositionMessage(start, this.unit.uri.toString());
   }
 
-  async expectHover(label: string, content: MarkupContent) {
+  expectHover(label: string, content: MarkupContent) {
     const indices = this.getLabelPositions(label);
```
ssmifi, line 750: _Minor_: Maybe `indicesOffsets` or `indicesPositions` is a more fitting name. At first, I thought, you're actually providing the fourslash label, because the test-builder separates labels in indices and ranges. :)

packages/language/test/test-builder.ts
```diff
@@ -746,11 +746,11 @@ export class TestBuilder {
...
 
     for (const index of indices) {
-      const hoverResult = await hoverRequest(this.unit, this.unit.uri, index);
+      const hoverResult = hoverRequest(this.unit, this.unit.uri, index);
```
ssmifi, line 753: _Question_: While this checks that the correct hover message for that token is returned by the LS, it does not check for the correct offsets. Should we also check for the correct span of the hover-over?

ssmifi (commented):
Looks good and works fine. I have a more general architectural question about the testing though (see below). Please let me know what you think.

packages/language/src/language-server/hover-request.ts
msujew: Likely, but I would do that as a follow up.

packages/language/test/fourslash-harness/implementation/test-builder.ts
msujew: I've adjusted this and combined it into one method for all `Include`, `IncludeAlt` and `Inscan` directives.

packages/language/test/test-builder.ts
msujew: I'll do that in a separate PR - the changes are unrelated to this PR.

packages/language/test/test-builder.ts
msujew: I'll do that in a separate PR - the changes are unrelated to this PR.

packages/language/test/fourslash-harness/implementation/test-builder.ts
ssmifi: What do you think about this in general?

ssmifi:
Could you please add todo notes for the postponed, unrelated changes? Or do you want to follow up on that immediately?

packages/language/test/fourslash-harness/implementation/test-builder.ts
msujew: Sure, we can do that as well. I'll adjust this on monday.

msujew:
@ssmifi I've updated the code to fix https://github.com/zowe/zowe-pli-language-support/pull/387#discussion_r2329720426 and include the `TODO`s.

ssmifi (approved):
LGTM. Thank you!

PR #389, opened 2025-09-08
Implement IBM2412I, IBM2409I, IBM2410I

Wagner-Laranjeiras:
### Implement IBM2412I, IBM2409I, IBM2410I: RETURN statement and RETURNS attribute validations

Issue: #383

**Summary:**

This PR introduces validation logic for RETURN statements in procedures and their interaction with the RETURNS attribute. The following IBM PL/I rules are enforced:

_IBM2412I:_ If a procedure contains a RETURN ( ... ) statement, it should declare the RETURNS attribute.
_IBM2409I_: When a procedure specifies the RETURNS attribute, all RETURN statements must provide an expression.
_IBM2410I:_ Procedures with the RETURNS attribute must contain at least one RETURN statement.

**Changes**

IBM2412I_IBM2410I_IBM2409I_handle_return_stmt_and_returns_att: new validation function implementing the rules above.
pli-validator: register the new validation logic.
AST: ensure ProcedureStatement and ReturnStatement expose the necessary tokens (procToken, returnToken).
Tests: add cases covering each of the three validation scenarios under packages/language/test/fourslash/validate/<IBM2412I or IBM2409I or IBM2410I>

### DIFF
--- a/packages/language/src/validation/messages/error-severity/IBM2412I-IBM2410I-IBM2409I-handle-return-stmt-and-returns-att.ts
+++ b/packages/language/src/validation/messages/error-severity/IBM2412I-IBM2410I-IBM2409I-handle-return-stmt-and-returns-att.ts
@@ -1,0 +1,96 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { ValidationAcceptor } from "../../validator";
+import * as AST from "../../../syntax-tree/ast";
+import { diagnosticFromCode } from "../../../language-server/types";
+import * as PLICodes from "../pli-codes";
+import {
+  TraversalState,
+  traverseAllNodes,
+} from "../../../syntax-tree/ast-iterator";
+
+/**
+ * IBM2412I: If a procedure contains a RETURN statement, it should have the RETURNS attribute
+ * specified on its PROCEDURE statement.
+ *
+ * IBM2409I: RETURN statement without an expression is invalid inside a nested PROCEDURE that
+ * specifies the RETURNS attribute. All RETURN statements inside functions must specify a value
+ * to be returned.
+ *
+ * IBM2410I: Functions must contain at least one RETURN statement.
+ *
+ * @param node The AST node being analyzed.
+ * @param acceptor The mechanism used to collect validation issues.
+ * @returns Validation results or diagnostics, as appropriate.
+ */
+export function IBM2412I_IBM2410I_IBM2409I_handle_return_stmt_and_returns_att(
+  node: AST.ProcedureStatement,
+  acceptor: ValidationAcceptor,
+): void {
+  const returnStmts: AST.ReturnStatement[] = [];
+
+  traverseAllNodes(node, (n) => {
+    if (n !== node && n.kind === AST.SyntaxKind.ProcedureStatement)
+      return TraversalState.Stop;
+    if (n.kind === AST.SyntaxKind.ReturnStatement)
+      returnStmts.push(n as AST.ReturnStatement);
+
+    return TraversalState.Continue;
+  });
+
+  const hasReturnsAtt = node.options?.some(
+    (att) => att.kind === AST.SyntaxKind.ReturnsOption,
+  );
+
+  if (returnStmts.length === 0 && !hasReturnsAtt) return;
+
+  const returnSomething: AST.ReturnStatement[] = [];
+  const returnNothing: AST.ReturnStatement[] = [];
+
+  for (const ret of returnStmts) {
+    if (ret.expression) returnSomething.push(ret);
+    else returnNothing.push(ret);
+  }
+
+  if (hasReturnsAtt && returnSomething.length > 0 && returnNothing.length === 0)
+    return;
+
+  // IBM2409I: All RETURN statements inside functions that specified the RETURNS attribute must specify a value to be returned.
+  if (hasReturnsAtt && returnNothing.length > 0) {
+    for (const ret of returnNothing) {
+      const returnToken = ret.returnToken;
+      if (!returnToken) return;
+
+      acceptor(diagnosticFromCode(PLICodes.Error.IBM2409I, returnToken));
+    }
+    return;
+  }
+
+  const procToken = node.procToken;
+  if (!procToken) return;
+
+  //IBM2410I: Procedures with RETURNS attribute must contain at least one RETURN statement.
+  if (hasReturnsAtt && returnStmts.length === 0) {
+    const procName =
+      node.container?.kind === AST.SyntaxKind.Statement &&
+      node.container.labels[0]?.nameToken?.originalImage
+        ? node.container.labels[0].nameToken.originalImage
+        : "<unnamed>";
+
+    acceptor(diagnosticFromCode(PLICodes.Error.IBM2410I, procToken, procName));
+  }
+
+  // IBM2412I: If a procedure contains a RETURN (...) statement, it should have the RETURNS attribute.
+  if (returnSomething.length > 0 && !hasReturnsAtt) {
+    acceptor(diagnosticFromCode(PLICodes.Error.IBM2412I, procToken));
+  }
+}
--- a/packages/language/test/fourslash/hover/procedure.ts
+++ b/packages/language/test/fourslash/hover/procedure.ts
@@ -30,7 +30,6 @@
 //// CALL <|2>MyProc;
 //// CALL <|3>MyProc2;
 
-verify.noDiagnostics();
 const expectedMarkdown = hover.codeBlock(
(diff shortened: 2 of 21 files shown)


packages/language/src/validation/messages/error-severity/IBM2412I-IBM2410I-IBM2409I-handle-return-stmt-and-returns-att.ts
```diff
@@ -0,0 +1,101 @@
...
+      if (n.kind === AST.SyntaxKind.ProcedureStatement)
+        return TraversalState.Skip;
+      if (n.kind === AST.SyntaxKind.ReturnStatement)
+        returnStmts.push(n as AST.ReturnStatement);
+      return TraversalState.Continue;
+    });
+  };
+  forEachNode(node, (child) => {
+    traverseChildren(child);
+  });
```
msujew: Minor: We can rewrite this to make this a bit easier to read.
```suggestion
  traverseAllNodes(node, (n) => {
    if (n !== node && n.kind === AST.SyntaxKind.ProcedureStatement)
      return TraversalState.Skip;
    if (n.kind === AST.SyntaxKind.ReturnStatement)
      returnStmts.push(n as AST.ReturnStatement);
    return TraversalState.Continue;
  });
```

packages/language/src/validation/messages/error-severity/IBM2412I-IBM2410I-IBM2409I-handle-return-stmt-and-returns-att.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  if (returnStmts.length === 0 && !hasReturnsAtt) return;
+
+  const returnSomething: AST.ReturnStatement[] = [];
+  const returnNothing: AST.ReturnStatement[] = [];
+
+  returnStmts?.forEach((ret: any) => {
+    if (ret.expression) returnSomething.push(ret);
+    if (!ret.expression) returnNothing.push(ret);
+  });
```
msujew: Minor: No need to check twice. Also, we usually prefer to use the `for each` loop of JS for this
```suggestion
  for (const ret of returnStmts) {
    if (ret.expression) 
      returnSomething.push(ret); 
    else 
      returnNothing.push(ret);
  };
```

packages/language/src/validation/messages/error-severity/IBM2412I-IBM2410I-IBM2409I-handle-return-stmt-and-returns-att.ts
```diff
@@ -0,0 +1,101 @@
...
+  if (!procToken) return;
+
+  //IBM2410I: Procedures with RETURNS attribute must contain at least one RETURN statement.
+  if (hasReturnsAtt && returnStmts.length === 0) {
+    acceptor(
+      diagnosticFromCode(
+        PLICodes.Error.IBM2410I,
+        procToken,
+        PLICodes.Error.IBM2410I.message(procToken.image),
+      ),
```
msujew: Suggestion: You only need to pass the actual argument in there, the call to `message` already happens internally.
```suggestion
      diagnosticFromCode(
        PLICodes.Error.IBM2410I,
        procToken,
        procToken.image,
      ),
```

packages/language/test/fourslash/hover/procedure.ts
```diff
@@ -30,7 +30,7 @@
 //// CALL <|2>MyProc;
 //// CALL <|3>MyProc2;
 
-verify.noDiagnostics();
```
msujew, line 33: Minor: Feel free to remove the whole `verify` call here, it's prone to breakage.

packages/language/src/validation/messages/error-severity/IBM2412I-IBM2410I-IBM2409I-handle-return-stmt-and-returns-att.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (ret.expression) returnSomething.push(ret);
+    if (!ret.expression) returnNothing.push(ret);
+  });
+
+  if (hasReturnsAtt && returnSomething.length > 0 && returnNothing.length === 0)
+    return;
+
+  // IBM2409I: All RETURN statements inside functions that specified the RETURNS attribute must specify a value to be returned.
+  if (hasReturnsAtt && returnNothing.length > 0) {
+    returnNothing.forEach((ret) => {
```
msujew:
```suggestion
    for (const ret of returnNothing) {
```

msujew (commented):
Looks pretty code, and very nice tests. The comments below are mostly for stylistic and code clarity issues 👍

packages/language/test/fourslash/validate/IBM2409I/proc-with-multi-tests.ts
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../../framework.ts" />
+
+// @wrap: main
+//// MAINPR: <|1:proc|> options( main );
```
Lotes: You are testing a lot of cases here. It is readable but also overwhelming when opening the first time.
The cases are independent as far as I can see. They do not influence each other. Can we split them into single files, if not already covered?

Lotes: Or what was your intent :-)...

msujew: I agree, these should be split into multiple tests

packages/language/src/validation/messages/error-severity/IBM2412I-IBM2410I-IBM2409I-handle-return-stmt-and-returns-att.ts
Wagner-Laranjeiras: Thanks, it looks cleaner indeed!

packages/language/src/validation/messages/error-severity/IBM2412I-IBM2410I-IBM2409I-handle-return-stmt-and-returns-att.ts
Wagner-Laranjeiras: Got it!

packages/language/test/fourslash/hover/procedure.ts
Wagner-Laranjeiras: Ok, removed.

packages/language/test/fourslash/validate/IBM2409I/proc-with-multi-tests.ts
Wagner-Laranjeiras: On it! Thanks guys :)

Wagner-Laranjeiras: I maintained the tests because even though we had similar ones, those were nested, so I think it was worth checking anyway. They are now split.

packages/language/src/syntax-tree/ast-iterator.ts
```diff
@@ -957,14 +958,20 @@ export function forEachNode(
...
-  if (state !== TraversalState.Stop) {
-    forEachNode(node, (child) => {
-      traverseAllNodes(child as SyntaxNode, traverse);
+  if (state === TraversalState.Stop) return TraversalState.Stop;
+  if (state === TraversalState.Skip) return TraversalState.Continue;
+
+  if (state === TraversalState.Continue) {
+    forEachNode(node, (child): TraversalState | void => {
+      const childState = traverseAllNodes(child as SyntaxNode, traverse);
+      if (childState === TraversalState.Stop) return TraversalState.Stop;
```
msujew: I don't really understand this change. Returning `TraversalState.Stop` inside of the callback for `forEachNode` doesn't do anything. Care to elaborate?

packages/language/test/fourslash/validate/IBM2409I/proc-with-multi-tests.ts
msujew: This test still looks the same. Forgot to commit something?

packages/language/src/syntax-tree/ast-iterator.ts
Wagner-Laranjeiras: The idea is to have three different states: `Continue` would keep the iteration going, `Skip` would jump to the next sibling node (ignoring the current nodes children) and `Stop` would stop the iteration entirely. 

The code you are referring doesn't do anything though, thanks for pointing that out. I'm still working on it, but I think it could be a good idea to do so in a different branch/issue. I also realized that for the IBM2410I I'm not really passing the right name to the acceptor, so that's what I'm working on now.

*I'll also remove the test with multiple labels on the next commit.

msujew: Yeah, I think it should be fine to do that in a separate PR. The functions works well enough as it is for now.

msujew: @Wagner-Laranjeiras Are you planning to revert the change? I think the function was perfectly fine as it was before the change.

Wagner-Laranjeiras: Ah ok, I will revert it and start from scratch on another branch.

msujew (approved):
Thanks, looks good to me 👍

PR #390, opened 2025-09-08
Improve ID tokenization rule

msujew:
Improves the perfomance of the ID tokenizer rule. It produced a lot of garbage during its runtime, which led to more than half of the tokenization time being spent in the garbage collector. This change reduces the tokenization time by roughly 20%.


ssmifi (approved):
LGTM. For consistency reasons you could also use consts in `isIdChar` similar to `isWhitespace` (or the other way around), but generally I'm also comfortable with these specific codes.

Issue #395, opened 2025-09-09
[Builtins] Update parameters and return types of builtin functions


msujew:
All builtin functions, their arguments and their return types can be found [here](https://www.ibm.com/docs/en/epfz/6.2.0?topic=subroutines-descriptions-individual-built-in-functions-pseudovariables).

VitGottwald:
Adding a list of the builtin functions [builtin-functions.txt](https://github.com/user-attachments/files/25003076/builtin-functions.txt).

When used in a call syntax (e.g. `DATE()`) they are considered implicitly declared by the compiler.

If used without a call syntax (e.g. `A = SYSNULL`) they have to be explicitly declared (`DCL SYSNULL  BUILTIN;`).

msujew:
@VitGottwald Thanks for the clarification, that wasn't clear so far!

VitGottwald:
It would also be great if we checked at the very least for the number of arguments on the builtin calls. It makes a big difference.

Ideally we would suggest the arguments by name and provide description as well. But that is more nice to have.

VitGottwald:
> [@VitGottwald](https://github.com/VitGottwald) Thanks for the clarification, that wasn't clear so far!

For the record this is documented in the Language Reference manual under `Null and optional arguments` section where it says:

> Some built-ins do not require arguments. You must either explicitly declare these with the BUILTIN
attribute or contextually declare them by including a null argument list in the reference—for example,
ONCHAR(). Otherwise, the name is not recognized as a built-in.

There is additional info under `Implicit declaration` section.

msujew:
@VitGottwald https://github.com/zowe/zowe-pli-language-support/pull/575 will fix the issue you brought up 👍

asatklichov:
done

PR #396, opened 2025-09-09
Improve LSP services for builtins

msujew:
Fixes an issue where the `definition` request did not work for variables/procedures from builtin files.

Additionally does no longer send diagnostics for builtin files - if they are invalid, that's by design. See also https://github.com/zowe/zowe-pli-language-support/issues/395.


ssmifi (approved):
LGTM.

PR #400, opened 2025-09-10
add tel events

asatklichov:


packages/vscode-extension/src/extension/main.ts
```diff
@@ -180,6 +216,7 @@ function watchPlipluginFolder(
   client: LanguageClient,
   workspaceFolder: string,
   context: vscode.ExtensionContext,
+  telemetryReporter: TelemetryReporter | undefined,
```
msujew, line 219: I'm not entirely sure it makes sense to add the report in here - it only listens to changes in the `.pliplugin` folder - nothing else. Does that really make sense?

msujew (approved):
Looks good to me. Just a question, see below.

packages/vscode-extension/src/extension/main.ts
asatklichov: at least I can see how many users  use config files (sure, after creation may not use it, but at least we can differentiate the users)

Issue #391, opened 2025-09-08
Declaration of SQLCA and SQLDA


asatklichov:
Declare SQLCA and SQLDA implicitly, so all variables will be resolved. 
See Programming Spec - page 127
---- 
`exec sql include sqlca;`
https://www.ibm.com/docs/en/db2-for-zos/12.0.0?topic=sqlca-included
`exec sql include sqlda;`
https://www.ibm.com/docs/en/db2-for-zos/12.0.0?topic=sqlda-included

asatklichov:
This we need to solve as an implicit copybook, which means hover-over support will show whole structure of SQLCA/SQLDA  and go-to definitions of variables will go to SQLCA/SQLDA where defined.

asatklichov:
done @msujew thanks

PR #401, opened 2025-09-10
Add SQLCA and SQLDA as normal builtins

msujew:
Resolves https://github.com/zowe/zowe-pli-language-support/issues/391

Adds builtins for the `SQLCA` and `SQLDA` variables to our language server. Does not add any actual support for `EXEC SQL` statements so far.

Most of the changes are unrelated to the actual change - I was unhappy with how we pull the `pli-options`, etc. variable names throughout the whole language server without any validation, etc.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -1840,7 +1840,7 @@ async function resolveIncludeFileUri(
       );
       const files: string[] = [libFileUri.path];
       // Generate a glob pattern for each include extension
-      for (const ext of pgroup["include-extensions"] ?? []) {
+      for (const ext of pgroup.includeExtensions) {
         const extFileUri = libFileUri.with({
           path: `${libFileUri.path}${ext}`,
         });
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -1832,7 +1832,7 @@ async function resolveIncludeFileUri(
 
   if (pgroup) {
     // lib file as either a string or a member from a known process group
-    for (const lib of pgroup.libs ?? []) {
+    for (const lib of pgroup.libs) {
       const libFileUri = UriUtils.joinPath(
         URI.parse(PluginConfigurationProviderInstance.getWorkspacePath()),
         lib,
--- a/packages/language/src/workspace/plugin-configuration-provider.ts
+++ b/packages/language/src/workspace/plugin-configuration-provider.ts
@@ -18,6 +18,7 @@ import {
 import { translateCompilerOptions } from "../preprocessor/compiler-options/translator";
 import { minimatch } from "minimatch";
 import { CompilerOptionResult } from "../preprocessor/compiler-options/options";
+import { isBoolean, isRecordOf, isString, isStringArray } from "../utils/types";
 
 /**
  * Pli options are effectively macros to set w/ the given values
@@ -31,7 +32,7 @@ export type PliOptions = Record<string, string>;
 export interface ProgramConfig {
   program: string;
   pgroup: string;
-  "pli-options"?: PliOptions;
+  pliOptions: PliOptions;
 
   /**
    * Prebuilt abstract options for this program config.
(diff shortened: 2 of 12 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1840,7 +1840,7 @@ async function resolveIncludeFileUri(
       );
       const files: string[] = [libFileUri.path];
       // Generate a glob pattern for each include extension
-      for (const ext of pgroup["include-extensions"] ?? []) {
+      for (const ext of pgroup.includeExtensions) {
```
ssmifi, line 1843: _Minor_: Same change is also valid on `lib` (line 1835) now.

packages/language/test/fourslash/linker/link-sqlca.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+/// <reference path="../framework.ts" />
+
+// @wrap: main
+//// PUT(<|1:SQLCA|>.SQLCODE);
+
+// Should link to the SQLCA builtin copybook definition
+// Therefore, no diagnostics expected
+verify.noDiagnostics(1);
```
ssmifi: _Minor_: Could also test for SQLDA, e.g., with tag 2.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -127,7 +180,7 @@ export class PluginConfigurationProvider {
     }
     for (const processGroup of this.processGroupConfigs.values()) {
       const libs = processGroup.libs || [];
-      const extensions = processGroup["include-extensions"] || [];
+      const extensions = processGroup.includeExtensions || [];
```
ssmifi: _Minor_: Is always set now. Same is true for `libs`.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -429,8 +502,8 @@ export class PluginConfigurationProvider {
     programConfig: ProgramConfig,
   ): string {
     const group = this.getProcessGroupConfig(programConfig.pgroup);
-    const groupOpts = group && group["pli-options"] ? group["pli-options"] : {};
-    const progOpts = programConfig["pli-options"] || {};
+    const groupOpts = group?.pliOptions ?? {};
+    const progOpts = programConfig.pliOptions || {};
```
ssmifi: _Minor_: Same.

ssmifi (approved):
LGTM, thank you. Just minor checks that became obsolete and one _suggestion_:
Now, we should probably use the `SerializedProcessGroup` interface also in the `main.ts` of the extension where the plugin folder with its defaults is created.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -310,6 +357,19 @@ export class PluginConfigurationProvider {
     return [abstractOptions, translatedOptions, collectedIssues];
   }
 
+  parseProgramConfigs(workspacePath: string, text: string): boolean {
+    try {
+      const programConfigs: SerializedProgramConfig[] = JSON.parse(text).pgms;
+      const programConfigsDeserialized = programConfigs.map(
```
montymxb: Small nitpick, can we change these names to `programConfigsSerialized` and `programConfigs`? Effectively flipping them around a bit. The first isn't a list of ProgramConfigs, but the 2nd is.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -328,6 +388,19 @@ export class PluginConfigurationProvider {
     }
   }
 
+  parseProcessGroupConfigs(text: string): boolean {
+    try {
+      const processGroupConfigs: SerializedProcessGroup[] =
+        JSON.parse(text).pgroups;
+      const groupConfigs = processGroupConfigs.map(deserializeProcessGroup);
```
montymxb: Same here as above. The 2nd can be `configs`, 1st can be something like `serializedConfigs` to align the names.

montymxb (commented):
Looks good, noted a pair of nitpicks.

Although not part of the original PR intent, I'm also on board with moving those config names away from strings plus pre-validating the values, that's a nice addition.

PR #404, opened 2025-09-15
Add ANSWER preprocessor statement

Lotes:
[ANSWER](https://www.ibm.com/docs/en/epfz/6.1.0?topic=procedures-preprocessor-answer-statement) has a lot of useless options like

* SKIP n
* PAGE
* COLUMN c
* MARGINS left right

They all get evaluated, since we could expect side-effects that can influence the answer.

And then there is also an optional "scan mode (`SCAN`|`RESCAN`|`NOSCAN`)". Here I have two tests that cover the cases for `RESCAN` and `NOSCAN`. Unfortunately, I have found no solution yet. I know there is code that does this task, but I could not find out yet how to feed it with the correct input.

All my tests also run on the mainframe I added a `ppp: PROC; ... END;` around each example to make it compilable.


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -152,6 +152,32 @@ function valueToString(value?: Value, defaultStr?: string): string | undefined {
   return value.value;
 }
 
+function concatValues(returnValue: Value, expression: Value): ScalarValue {
+  // TODO verify on mainframe if this is correct
```
Lotes: Yup, TODO

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -482,13 +508,39 @@ function runInstructionSync(
...
   }
   return undefined;
 }
 
+function runAnswerInstruction(
+  instruction: inst.AnswerInstruction,
+  context: InterpreterContext,
+): void {
+  if (instruction.expression) {
+    const expression = evaluateExpression(instruction.expression, context);
```
Lotes, line 506: TODO make NOSCAN/SCAN work here

code_samples/preprocessor/answer.pli
Lotes, line 1: DELETE.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -152,6 +152,32 @@ function valueToString(value?: Value, defaultStr?: string): string | undefined {
...
+  }
+  pushValue(returnValue);
+  pushValue(expression);
+  if (values.length === 0) {
+    return defaultEmptyValue;
+  }
+  return {
+    type: inst.DeclaredType.Character,
+    value: values.map((v) => v.value).join(""),
+  };
```
msujew: Suggestion: Joining these variables without any spaces in between will merge them in the output:

```
ANS("X") SKIP; // SKIP moves the cursor to the next line
ANS("Y") SKIP;
```
Should result in the output:

```
X
Y
```

Which are two distinct tokens. But in our case, it will merge the output to just a single `XY` token.

packages/language/src/syntax-tree/ast.ts
```diff
@@ -784,6 +787,39 @@ export function createTokenStatement(): TokenStatement {
...
 }
 
+export type SkipMode =
+  | {
+      type: "page";
+    }
+  | {
+      type: "skip";
+      count: Expression | null;
+    };
```
msujew: Suggestion: Please use an `enum` here. Number comparison is faster than string comparisons, which is why the rest of the code is now also `enum`-only.

msujew (commented):
Looks quite good. Please run `pnpm pretty` to fix the CI. Also there's a failing test.

packages/language/src/preprocessor/instruction-interpreter.ts
Lotes: Done, thx.

FYI, tested on the mainframe:

```
ANS("X");
ANS("Y") SKIP; //prepends a newline BEFORE the own output
```

Result:
```
X
Y
```

packages/language/src/preprocessor/instruction-interpreter.ts
Lotes: Concatenation is now done differently

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -518,29 +492,117 @@ function runInstructionSync(
   return undefined;
 }
 
+function tokenOfType(token: Token, type: TokenType): boolean {
+  return (
+    token.tokenType === type ||
+    (token.tokenType.CATEGORIES?.includes(type) ?? false)
+  );
+}
```
msujew: Suggestion: We already have a `tokenMatcher` function exposed by `chevrotain` that we use throughout the codebase for this.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -518,29 +492,117 @@ function runInstructionSync(
...
+    endOffset: rightId.endOffset,
+    tokenType: PreprocessorTokens.Id,
+    element: undefined,
+    kind: undefined,
+    immediateFollow: false,
+    tokenTypeIdx: leftId.tokenTypeIdx,
+    uri: undefined,
+    isInsertedInRecovery: false,
+  };
+}
```
msujew: Suggestion: I don't think this logic is correct. Which is why `runTokenInstruction` (now `mergePush`) is using `lex` to generate the merged token. Something like `DE` and `CLARE` are individually just IDs, but when they're merged they become `DECLARE`, which is a different token type.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -482,13 +482,127 @@ function runInstructionSync(
...
+        tokens.length > 0 &&
+        tokenOfType(tokens[0], PreprocessorTokens.Id)
+      ) {
+        const lastToken = context.tokens.pop()!;
+        const firstToken = tokens.shift()!;
+        const newToken: Token = mergeIdTokens(lastToken, firstToken);
+        context.tokens.push(newToken);
+      }
+      context.tokens.push(...tokens);
+    }
```
msujew: Suggestion: Why not use `mergePush` instead?

packages/language/src/preprocessor/instruction-interpreter.ts
Lotes: OK, thanks for this hint.

Lotes: Done.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -482,13 +482,94 @@ function runInstructionSync(
...
+      breakCount = skipCount;
+    } else {
+      context.diagnostics.push(
+        diagnosticFromCode(
+          Severe.IBM3948I,
+          instruction.skipToken,
+          "CONVERSION",
+          "612",
+        ),
+      );
```
msujew: Minor: Can you add a todo here, that this should be eventually moved into the type system? The type system should be able to statically identify the conversion error here.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -482,13 +482,94 @@ function runInstructionSync(
...
+    const columnCount = evaluateExpression(instruction.column, context);
+    if (typeof valueToNumber(columnCount) !== "number") {
+      context.diagnostics.push(
+        diagnosticFromCode(
+          Severe.IBM3948I,
+          instruction.columnToken,
+          "CONVERSION",
+          "612",
+        ),
+      );
```
msujew: Same here.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -482,13 +482,94 @@ function runInstructionSync(
...
+      }
+    }
+    if (hasError) {
+      context.diagnostics.push(
+        diagnosticFromCode(
+          Severe.IBM3948I,
+          instruction.marginsToken,
+          "CONVERSION",
+          "612",
+        ),
```
msujew: Same here.

package.json
```diff
@@ -10,7 +10,7 @@
     "bundle": "pnpm --dir packages/vscode-extension esbuild:bundle",
     "merge": "tsx ./scripts/generate-tmlanguage.mts",
     "build": "pnpm clean && pnpm compile && pnpm bundle && pnpm merge",
-    "package": "pnpm build && pnpm --dir packages/vscode-extension package",
+    "package": "pnpm build && pnpm --dir packages/vscode-extension package --allow-package-secrets sendgrid",
```
msujew: Suggestion: Why not add this to the `package` script of the `vscode-extension`? This will lead to errors in case someone uses the script directly instead of the one defined in root.

msujew (approved):
Looks good, just a few minor issues, see below.

package.json
Lotes: Ah! I see, I thought this was already the final call, but it is just a call to the workspace folder.

Lotes: Done.

PR #405, opened 2025-09-15
Remove `langium` dependency remnants

msujew:
I believe I wasn't really throughout enough, when removing the references to the `langium` package. This removes all package references.


ssmifi (approved):
LGTM

PR #406, opened 2025-09-15
Support includes in web extension

msujew:
Superseeds https://github.com/zowe/zowe-pli-language-support/pull/360.

This change adjusts the include support to use a `search` function that performs a system dependent, case insensitive file search for includes.

Contains three search implementations
1. Virtual file system (very simplistic, map based).
2. Node.js file search.
3. VS Code file search transmitting the data through the LSP connection. Used in the web extension implementation.

Implementation 2 and 3 currently need to be manually tested.


packages/vscode-extension/src/common/search.ts
```diff
@@ -0,0 +1,56 @@
+/**
```
ssmifi: _Minor_: Should we rename the file to indicate that this has something to do with the filesystem?

ssmifi (approved):
LGTM

Issue #392, opened 2025-09-08
Define structure support (TYPE, LIKE etc.)


asatklichov:
- [x] Currently   `DEFINE STRUCTURE` has limited support, only definition is resolved. 
It should work with  HANDLE(<field>)

```
DEFINE STRUCTURE                                                            
     01 SRV_CATALOG_EXE_ENTRY                                                  
       ,05 NXT_EXE_ENY_HL      HANDLE(SRV_CATALOG_EXE_ENTRY)                    
       ,05 DOM_ID              TYPE(SRV_DOMAIN)                                
       ,05 SRV_ID              TYPE(SRV_SERVICE)                                
       ,05 OPN_ID              TYPE(SRV_OPERATION)                              
       ,05 MAJ_VRS_ID          TYPE(SRV_MAJOR_VERSION)                          
       ,05 MNO_VRS_ID          TYPE(SRV_MAJOR_VERSION)                          
       ,05 ORI_ID              TYPE(SRV_ORIGIN)                                
       ,05 DLL_ID              CHAR(8)                                          
       ,05 FCT_ID              CHAR(64)                                        
       ,05 AVB_CHE_FL          CHAR(1)                                          
       ,05 FCT_PR              POINTER                                          
       ,05 DLL_PR              POINTER;  

```

DEFINE STRUCTURE, is just a definition, or type definition. Not declared yet. So, you can not use it unless declared as below. 

1-way of declaration (via parathesis) using defined type: 
DCL my_catalog_entry type(SRV_CATALOG_EXE_ENTRY); 
2-way of declaration using defined type: 
DCL my_catalog_entry type SRV_CATALOG_EXE_ENTRY;
Usage in code: 
my_catalog_entry.dom_id = 'xyz'; 

- [x]  Declaration with` 'LIKE'`
e.g.  
DCL    WRK_SAVED_EIB  LIKE DFHEIBLK; 
DCL a char(10);
DCL  b  like a;
 

or another example

<img width="815" height="287" alt="Image" src="https://github.com/user-attachments/assets/bbee7061-8c96-400b-959a-b95783bbdd44" />

IN TWO DIFFERENT SPACES
T - TYPE NAMESPACE
T - IN VARIABLE NAMESPAVE 
 
COMPILER WILL NOT COMPLAIN .. 

- [x] Also declaration with `SQL TYPE` and its all alternatives should be resolved 

[...]

PR #408, opened 2025-09-17
Support `SQL TYPE` attribute

msujew:
Resolves the last point of https://github.com/zowe/zowe-pli-language-support/issues/392.

Implements two major changes:
1. Implements support for all the `SQL TYPE` attributes on variables, see [here](https://www.ibm.com/docs/en/db2-for-zos/12.0.0?topic=pli-host-variable-arrays-in).
2. Refactors our token creation logic to make it easier to add new keywords to the list (this is 99% of this PR).

Also removes a few outdated tests that were already translated to the fourslash framework.


VitGottwald:
How is the SQL TYPE support enabled/disabled in compiler options?

msujew:
@VitGottwald Good question. I haven't found anything in the documentation about this feature being controlled by compiler options. Do you know more about this, or can you link some documentation?

VitGottwald:
Try removing the PP(SQL) option from the compile script ;-)

VitGottwald:
It is documented in the [Programming Guide](https://www.ibm.com/docs/en/SSY2V3_6.1/pdf/pg.pdf).

VitGottwald:
One more note - SQL related support is documented in [here](https://www.ibm.com/docs/en/SSEPEK_13.0.0/pdf/db2z_13_apsgbook.pdf) in section _PL/I applications that issue SQL statements_ which has about 40 pages. It is a newer version of the doc for DB2 13. The link in the description points to DB2 12 doc. I also find the PDF easier to navigate than the html variant.

msujew:
@VitGottwald Thanks for the hint, I thought the SQL preprocessor only worked on the `EXEC SQL` stuff, but it seems like it also takes care of transforming the `SQL TYPE` attribute into "normal" PLI attribute. I'll close the PR, since this requires the SQL preprocessor step to be in place, which is probably something I'll take care of next.

PR #409, opened 2025-09-18
Add CALL preprocessor statement

Lotes:
TODO:

- [x] add test cases with nested PROCEDUREs
  - [x] there are no nested procedure in preprocessor 
- [x] add test cases with diagnostics

### DIFF
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -104,6 +104,7 @@ function generateProcedureInstructionContainer(
     statement: statementType,
     parameters: params,
     node: instructionList.entryNode,
+    options: procedure.options
   };
 }
 
@@ -221,6 +222,7 @@ function generateCallInstruction(
   return {
     kind: inst.InstructionKind.Call,
     procedureName: node.call.procedure.text,
+    procedureNameToken: node.call.procedure.token,
     args,
     node,
   };
--- a/packages/language/src/preprocessor/instruction-generator.ts
+++ b/packages/language/src/preprocessor/instruction-generator.ts
@@ -11,7 +11,7 @@
 
 import * as inst from "./instructions";
 import * as ast from "../syntax-tree/ast";
-import { getAttributes } from "./util";
+import { assertType, getAttributes } from "./util";
 
 interface GenerateInstructionContext {
   labels: Map<string, inst.InstructionNode>;
@@ -172,6 +172,9 @@ function generateInstructionForStatement(
     case ast.SyntaxKind.AnswerStatement:
       instruction = generateAnswerInstruction(value);
       break;
+    case ast.SyntaxKind.CallStatement:
+      instruction = generateCallInstruction(value, context);
+      break;
     default:
       return undefined;
   }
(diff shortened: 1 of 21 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -485,13 +493,80 @@ function runInstructionSync(
...
+      ),
+    );
+    for (
+      let count = procedure.parameters.length;
+      count < instruction.args.length;
+      count++
+    ) {
+      instruction.args.pop();
+    }
+  }
```
msujew: Suggestion: This for loop is not necessary, the `evaluateProcedure` invocation is already taking care of superfluous arguments to the procedure call. The diagnostic should be shown in there instead of here.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -485,13 +493,80 @@ function runInstructionSync(
...
+      let count = instruction.args.length;
+      count < procedure.parameters.length;
+      count++
+    ) {
+      instruction.args.push({
+        kind: inst.InstructionKind.Number,
+        value: "0",
+      });
+    }
+  }
```
msujew: Suggestion: This for loop is not necessary, the `evaluateProcedure` invocation is already taking care of missing arguments to the procedure call. The diagnostic should be shown in there instead of here.

packages/language/src/preprocessor/instruction-interpreter.ts
Lotes: Now I did so, but I do not like the passing of the token... I need the token of the invocation...

packages/language/src/preprocessor/instruction-generator.ts
```diff
@@ -198,6 +202,32 @@ function generateReturnInstruction(
...
+function generateCallInstruction(
+  node: ast.CallStatement,
+  context: GenerateInstructionContext,
+): inst.CallInstruction | undefined {
+  if (!node.call?.procedure?.text) {
+    return undefined; // No procedure to call
+  }
+  if (!node.call.procedure.text) {
+    return undefined;
+  }
```
msujew: Duplicate?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1188,12 +1263,34 @@ function runProcedure(
...
+    } else if (procedure.parameters.length < args.length) {
+      context.diagnostics.push(
+        diagnosticFromCode(
+          Warning.IBM3324I,
+          callToken,
+          callToken.image,
+        ),
+      );
+    }
+  }
```
msujew: I'm wondering whether it's better to move these diagnostics to the validator (and leave a note/todo here) - a loop such as this, will result in multiple diagnostics on the same position:

```
DO I = 1 TO 5;
  CALL TEST_PROC(/* too few args */);
END;
```

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -485,13 +493,45 @@ function runInstructionSync(
...
+  if (
+    procedure.options.some(
+      (option) => option.kind === ast.SyntaxKind.ReturnsOption,
+    )
+  ) {
+    context.diagnostics.push(
+      diagnosticFromCode(Severe.IBM3970I, instruction.procedureNameToken),
+    );
+  }
```
msujew: Suggestion: I believe that all of these should be handled by the type system. See reason above.

packages/language/src/preprocessor/instruction-interpreter.ts
Lotes: Removed

packages/language/src/preprocessor/instructions.ts
```diff
@@ -106,14 +107,16 @@ export type Instruction =
...
+  | NoteInstruction
+  | CallInstruction;
 
 export interface ProcedureInstructionContainer {
   names: string[];
   labels: Map<string, ast.LabelPrefix>;
   parameters: string[];
   statement: boolean;
   node: InstructionNode;
+  options: ast.ProcedureOption[];
```
Lotes: Remove

msujew (approved):
Looks good to me, thanks 👍

Issue #397, opened 2025-09-09
Macro Preprocessor Support


asatklichov:
Currently we have support for macro preprocessor statements. 
Now, we want to support  for 'Macro preprocessor options'. 

See spec.  Programming Guide Chapter 2 (page 116).  

 
Macro preprocessor options TABLE

<img width="428" height="406" alt="Image" src="https://github.com/user-attachments/assets/3b089799-38c3-495d-9549-0bca5ad2140d" />

If you want to specify more than one option, you must separate them with a comma or one or
more blanks. For example, to specify the CASE(ASIS) and RESCAN(UPPER) options, you can specify
PP(MACRO(’CASE(ASIS) RESCAN(UPPER)’)) or PP(MACRO("CASE(ASIS),RESCAN(UPPER)")). You can
specify the options in any order
 

<img width="794" height="370" alt="Image" src="https://github.com/user-attachments/assets/f06d385d-b048-49b6-8bb8-640d4342e4d5" />

PP

<img width="746" height="401" alt="Image" src="https://github.com/user-attachments/assets/aaa89b4e-5bc4-4776-a2ad-110b8023aec0" />

ssmifi:
Follow up on selected effects in https://github.com/zowe/zowe-pli-language-support/issues/441

asatklichov:
Not fully implemented yet

PR #417, opened 2025-09-19
Add macro compiler options

ssmifi:
First part of https://github.com/zowe/zowe-pli-language-support/issues/397

Mostly refactoring to allow for multiple compiler options categories and staged parsing.


packages/language/src/preprocessor/compiler-options/options-macro.ts
```diff
@@ -0,0 +1,57 @@
...
+  namePrefix?:
+    | {
+        character: string;
+      }
+    | false;
+  rescan?: CompilerOptions.Rescan;
+}
+
+export namespace CompilerOptions {
+  export type Case = "UPPER" | "ASIS";
```
Lotes, line 35: I am wondering if these literals should be enums. In the AST we use enums because strings are expensive for comparison.

packages/language/src/preprocessor/compiler-options/translator-pli.ts
Lotes, line 1: This file is very big. Was it generated?

packages/language/src/preprocessor/compiler-options/options-macro.ts
ssmifi: I kept most of the options as they were originally and they usually represent the compiler options literally. We could change that of course in the future. I would refrain from this in this PR though.

packages/language/src/preprocessor/compiler-options/translator-pli.ts
ssmifi: The actual rules grew over time. The options/stubs might have been generated. I thought about splitting it up, but decided against it for now, because the structure is relatively simple, ordered alphabetically and specific options can be searched easily. I put the options of the other (new) preprocessors and common functions in separate files in this PR though. Please let me know if you have further suggestions/concerns.

packages/language/src/preprocessor/compiler-options/options.ts
```diff
@@ -11,7 +11,15 @@
...
 import { Token } from "../../parser/tokens";
+import * as Pli from "./options-pli";
+import * as Macro from "./options-macro";
 
+export type CompilerOptionsPP = Pli.CompilerOptions | Macro.CompilerOptions;
+
+export interface CompilerOptions {
+  pli: Pli.CompilerOptions;
+  macro: Macro.CompilerOptions;
+}
```
msujew: Question: Is there a specific reason why the `macro` options are not embedded in the `pli` options? This makes it IMO more difficult to access the normal compiler options.

msujew (commented):
Looks mostly good to me, I just have one question.

packages/language/src/preprocessor/compiler-options/options.ts
ssmifi: This was merely a choice to separate the huge file into separate concerns as I didn't want the macro rules to interfere with the normal rules. The macro (and other pp rules) can occur at several places and I wanted to reuse our rules system, but they should of course not be parsable with the normal options.
I can merge the macro options back into the pli options (and separate the rules via `TranslatorRule` f.e. to avoid interference), if you prefer it that way. (More options for the other preprocessors are still incoming though.)
I would still prefer to separate the rules from the utilities, though, if you agree. Please let me know what you think.

msujew: I'm fine with splitting the processing of the compiler options - my worry is only concerned with the readability of code that uses the compiler options :)

ssmifi: 👍 Let me think about that. :)

packages/language/src/preprocessor/compiler-options/translator-pli.ts
Lotes: No suggestions right now.

packages/language/src/preprocessor/compiler-options/options.ts
ssmifi: The standard compiler options are now accessible as before with the others pp options being nested inside.

msujew (approved):
Alright, looks good to me 👍

Issue #332, opened 2025-08-14
Quick fix to resolve copybook


asatklichov:
Currently we asking (on every opened file) repeatedly to create **.plugin** folder with two config files if user has not created it yet. 

Let's change the logic as below.  

1. Ask just only-once to create folder when user opens file for the first time. User can accept YES or NO, that is fine.
2. Then once user opens a file (or working on same file) and starts using INCLUDE, we need to help user with "quick fix" 
a - if folder is not there create it with respective config files, and open "proc_grps.json" file directly, 
and notify user "Please define your include libraries where your copybooks reside" 
b. if user has already that folder, then just open "proc_grps.json" file directly, 
and notify user "Please define your include libraries where your copybooks reside"

montymxb:
Un-assigning myself and dropping in a general workflow that I sketched for both issues, so the next assignee can pick up nicely on this. This isn't an exact solution, but should give a sense of what needs to be done, how it can be done, and what creative liberties you can take along the way too.

## Workflow: Initial Configuration Prompt

**Trigger:**  
When a user opens any pli file and either:
- The `.pliplugin` folder does not exist, or
- The required configuration files within `.pliplugin` are missing.

**Steps:**

1. **Check Configuration Status**
     - On file open, check for the existence of the `.pliplugin` folder.
     - If the folder exists, verify that both required configuration files (`proc_grps.json`, `pgm_conf.json`) are present:
     - If the contents of either file are invalid, don't proceed, we can't guarantee a clean state in such a case.
     - The current logic already prompts when we're missing the entire plugin folder, so we can build on this.

2. **Prompt User (Only Once Per Session)**
     - If configuration is missing or invalid, display a prompt:
       - "PL/I configuration is incomplete. Would you like to create the `.pliplugin` folder and configuration files?"
       - Options: YES / NO / Don't ask again
     - Track the user's response using the extension context to prevent re-prompting during the session.

3. **User Response Handling**
     - If YES:
       - Create the `.pliplugin` folder if missing.
       - Create or overwrite `proc_grps.json` and `pgm_conf.json` with valid default structure.
       - Use the opened file as the entry point in `pgm_conf.json`
           - In some cases, the given file may just be a copybook, then trace back to the owning compile unit & its associated process group when possible.
     - If NO:
       - Dismiss prompt.
       - Set session flag to prevent re-prompting.
     - If "Don't ask again":
       - Store persistent preference to avoid prompting again in this workspace.

4. **Session Management**
[...]

PR #418, opened 2025-09-24
Implement Quick Fix for unresolved INCLUDE files

Wagner-Laranjeiras:
### Implement Quick Fix for unresolved INCLUDE files (update proc_grps.json libs)

Issue: #332

**Summary:**
This PR introduces a Quick Fix action that helps resolve unresolved INCLUDE files by automatically updating the proc_grps.json configuration. When an INCLUDE file is detected but not present in the libs of the default process group, this fix suggests adding the parent folder of the file to the configuration.

**Changes**
quickFixResolveInclude:
Retrieves unresolved INCLUDE file information from diagnostics.
Locates the file in the workspace using FileSystemProviderInstance.findFilesByGlob.
Computes the parent folder relative to the workspace.
Updates proc_grps.json by adding the folder path to the libs array of the default process group (if not already present).
Returns a CodeAction with a Quick Fix that applies the change through the pli.applyIncludeFix command.
Implemented `computeWorkspaceRelativeParentFolder` helper.

**Notes**
1. Had to reintroduce `findFilesByGlob` helper. Please let me know how we should proceed in order to avoid technical debt by having this function. I have another way to solve this in a local branch, but it seemed way over complicated/complex.
2. I couldn't think of a way to retrieve the unresolved file name other than: 1. adding it to the diagnostic or 2. slicing it from the diagnostic message. I thought it was cleaner to change the diagnostic, but any suggestion is welcome.
3. All tests were manual. I realized I would be taking more time in order to modify the `verify` function, so I opted for creating this PR already. I would like to tackle this later though, with you guys help.

### DIFF
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -1,0 +1,102 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { URI } from "../../utils/uri";
+import { FileSystemProviderInstance } from "../../workspace/file-system-provider";
+import { UriUtils } from "../../utils/uri";
+import {
+  CodeAction,
+  CodeActionKind,
+  Diagnostic,
+} from "vscode-languageserver-types";
+import {
+  PluginConfigurationProvider,
+  PluginConfigurationProviderInstance,
+  serializeProcessGroup,
+} from "../../workspace/plugin-configuration-provider";
+
+export async function quickFixResolveInclude(
+  diagnostic: Diagnostic,
+): Promise<CodeAction | undefined> {
+  const unresolvedFile = diagnostic.data?.unresolvedFile;
+  const procGrpsConfig =
+    PluginConfigurationProviderInstance.getProcessGroupConfig("default");
+  if (!procGrpsConfig || !unresolvedFile) return undefined;
+
+  const configExtensions = procGrpsConfig.includeExtensions;
+  const unresolvedFilePath = await FileSystemProviderInstance.search({
+    path: URI.parse(unresolvedFile),
+    extensions: configExtensions,
+    global: true,
+  });
+  if (!unresolvedFilePath) return undefined;
+
+  const workspaceFolderUri = URI.parse(
+    PluginConfigurationProviderInstance.getWorkspacePath(),
+  );
+
+  let parentFolder = UriUtils.computeWorkspaceRelativeParentFolder(
+    unresolvedFilePath,
+    workspaceFolderUri,
+  );
+  if (!parentFolder) return undefined;
+  if (procGrpsConfig.libs.includes(parentFolder)) return undefined;
+
+  const serializedProcGrpsConfig = serializeProcessGroup(procGrpsConfig);
+  const newContent = JSON.stringify(
+    {
+      pgroups: [
+        {
+          ...serializedProcGrpsConfig,
+          libs: [...(serializedProcGrpsConfig.libs ?? []), parentFolder],
+        },
+      ],
+    },
+    undefined,
+    2,
+  );
+
+  const procGrpsFileUri = UriUtils.joinPath(
+    workspaceFolderUri,
+    PluginConfigurationProvider.PROCESS_GROUP_CONFIG_FILE,
+  );
+
+  const action: CodeAction = {
+    title: `Add '${parentFolder}' to INCLUDE libs`,
+    kind: CodeActionKind.QuickFix,
+    diagnostics: [diagnostic],
+    command: {
+      title: "Apply INCLUDE fix",
+      command: "pli.applyIncludeFix",
+      arguments: [procGrpsFileUri.toString(), newContent],
+    },
+  };
+
+  return action;
+}
+
+export async function applyQuickFixes(
+  diagnostics: Diagnostic[],
+): Promise<CodeAction[] | undefined> {
+  const actions: CodeAction[] = [];
+  // PLI CODES LIST
+  const CODE_UNRESOLVED_INCLUDE = "IBM3841IS"; // The INCLUDE file could not be found, or if found, it could not be opened.
+
+  for (const diagnostic of diagnostics) {
+    if (diagnostic.code === CODE_UNRESOLVED_INCLUDE) {
+      const action = await quickFixResolveInclude(diagnostic);
+      if (action) actions.push(action);
+    }
+  }
+
+  if (!actions.length) return undefined;
+  return actions;
+}
--- a/packages/language/src/language-server/connection-handler.ts
+++ b/packages/language/src/language-server/connection-handler.ts
@@ -20,7 +20,12 @@ import { URI, UriUtils } from "../utils/uri";
 import { definitionRequest } from "./definition-request";
 import { referencesRequest } from "./references-request";
 import { semanticTokenLegend, semanticTokens } from "./semantic-tokens";
-import { Location, TextEdit } from "vscode-languageserver-types";
+import {
+  CodeAction,
+  Diagnostic,
+  Location,
+  TextEdit,
+} from "vscode-languageserver-types";
 import {
   completionItemToLSP,
   documentSymbolToLSP,
@@ -35,6 +40,8 @@ import { PluginConfigurationProviderInstance } from "../workspace/plugin-configu
 import { completionRequest } from "./completion/completion-request";
 import { hoverRequest } from "./hover-request";
 import { Mutex } from "../workspace/mutex";
+import { applyQuickFixes } from "./code-actions/apply-quick-fixes";
+import { FileSystemProviderInstance } from "../workspace/file-system-provider";
 
 /**
  * Notification sent to the LS when the workspace's plugin configuration changes.
@@ -45,6 +52,8 @@ export const WorkspaceDidChangePlipluginConfigNotification =
 export function startLanguageServer(connection: Connection): void {
   const compilationUnitHandler = new CompilationUnitHandler();
   compilationUnitHandler.listen(connection);
+  const APPLY_QUICK_FIXES = "pli.applyIncludeFix";
+
   connection.onInitialize(async (params) => {
     // init the plugin config provider in reverse folder order, last plugin config encountered will take precedence
     // TODO @montymxb Apr 23rd, 2025: Consider addressing multiple workspaces w/ multiple plugin configs
@@ -71,6 +80,10 @@ export function startLanguageServer(connection: Connection): void {
         renameProvider: true,
         definitionProvider: true,
         referencesProvider: true,
+        codeActionProvider: true,
+        executeCommandProvider: {
+          commands: [APPLY_QUICK_FIXES],
+        },
         documentHighlightProvider: true,
         semanticTokensProvider: {
           legend: semanticTokenLegend,
@@ -291,5 +304,31 @@ export function startLanguageServer(connection: Connection): void {
       });
     },
   );
+
+  connection.onCodeAction(async (params) => {
+    return Mutex.read(async () => {
+      const diagnostics = params.context.diagnostics as Diagnostic[];
+      if (!diagnostics.length) return;
+      const actions: CodeAction[] | undefined =
+        await applyQuickFixes(diagnostics);
+      if (!actions) return;
+
+      return actions;
+    });
+  });
+
+  connection.onExecuteCommand(async (params) => {
+    if (params.command === APPLY_QUICK_FIXES) {
+      return Mutex.run(async () => {
+        const [uri, content] = params.arguments as string[];
+        try {
+          await FileSystemProviderInstance.writeFile(URI.parse(uri), content);
+        } catch (err) {
+          console.error("Failed to write proc_grps.json:", err);
+        }
+      });
+    }
+  });
+
   connection.listen();
 }
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -1886,9 +1886,13 @@ async function runInclude(
     if (error) {
       console.log("Failed to resolve include file:", error);
     }
-    context.diagnostics.push(
-      diagnosticFromCode(PLICodes.Severe.IBM3841I, item.token, item.fileName),
+    const diagnostic = diagnosticFromCode(
+      PLICodes.Severe.IBM3841I,
+      item.token,
+      item.fileName,
     );
+    if (item.fileName) diagnostic.data = { unresolvedFile: item.fileName };
+    context.diagnostics.push(diagnostic);
   }
 
   if (!uri) {
@@ -2025,7 +2029,7 @@ async function resolveIncludeFileUri(
       const match = await FileSystemProviderInstance.search({
         path: libFileUri,
         extensions: pgroup.includeExtensions,
-      });
+      })) as URI | undefined;
       if (match) {
         return match;
       }
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2026,10 +2026,10 @@ async function resolveIncludeFileUri(
         libFileUri = UriUtils.joinPath(libUri, item.fileName);
       }
 
-      const match = (await FileSystemProviderInstance.search({
+      const match = await FileSystemProviderInstance.search({
         path: libFileUri,
         extensions: pgroup.includeExtensions,
-      }))
+      });
       if (match) {
         return match;
       }
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2029,7 +2029,7 @@ async function resolveIncludeFileUri(
       const match = await FileSystemProviderInstance.search({
         path: libFileUri,
         extensions: pgroup.includeExtensions,
-      })) as URI | undefined;
+      }) as URI | undefined;
       if (match) {
         return match;
       }
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2029,7 +2029,7 @@ async function resolveIncludeFileUri(
       const match = (await FileSystemProviderInstance.search({
         path: libFileUri,
         extensions: pgroup.includeExtensions,
-      })) as URI | undefined;
+      }))
       if (match) {
         return match;
       }
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2026,10 +2026,10 @@ async function resolveIncludeFileUri(
         libFileUri = UriUtils.joinPath(libUri, item.fileName);
       }
 
-      const match = await FileSystemProviderInstance.search({
+      const match = (await FileSystemProviderInstance.search({
         path: libFileUri,
         extensions: pgroup.includeExtensions,
-      }) as URI | undefined;
+      })) as URI | undefined;
       if (match) {
         return match;
       }
--- a/packages/language/src/utils/uri.ts
+++ b/packages/language/src/utils/uri.ts
@@ -61,4 +61,35 @@ export namespace UriUtils {
   export function normalize(uri: URI | string): string {
     return URI.parse(uri.toString()).toString();
   }
+
+  /**
+   * Compute workspace-relative parent folder for a found file candidate.
+   *
+   * Examples:
+   *  - workspace root: /repo/plugin-example
+   *  - candidate: /repo/plugin-example/cpy/test-nested-included/nested-not-here.pli
+   *  -> returns "cpy/test-nested-included"
+   *
+   * If workspace-relative cannot be computed, returns the parent folder name (e.g. "test-nested-included").
+   * Returns undefined if candidate looks invalid.
+   */
+  export function computeWorkspaceRelativeParentFolder(
+    candidateRaw: URI,
+    workspaceFolderUri: URI,
+  ): string | undefined {
+    if (!candidateRaw) return undefined;
+
+    const candidate = normalize(candidateRaw);
+    const workspaceFolder = normalize(workspaceFolderUri);
+
+    if (!candidate.startsWith(workspaceFolder)) return undefined;
+
+    const relativePath = candidate.slice(workspaceFolder.length);
+    const parentDir = relativePath.substring(0, relativePath.lastIndexOf("/"));
+    const parentRelativePath = parentDir.startsWith("/")
+      ? parentDir.slice(1)
+      : parentDir;
+
+    return parentRelativePath;
+  }
 }
--- a/packages/language/src/workspace/file-system-provider.ts
+++ b/packages/language/src/workspace/file-system-provider.ts
@@ -14,6 +14,7 @@ import { URI } from "../utils/uri";
 export interface SearchOptions {
   path: URI;
   extensions: string[];
+  global?: boolean;
 }
 
 export interface FileSystemProvider {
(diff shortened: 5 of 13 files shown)


packages/language/src/language-server/connection-handler.ts
```diff
@@ -299,5 +309,36 @@ export function startLanguageServer(connection: Connection): void {
...
+        await connection.sendRequest("workspace/fs/writeFile", {
+          uri,
+          value: content,
+          message,
+        });
+      } catch (err) {
+        console.error("Failed to write proc_grps.json:", err);
+      }
+    }
+  });
```
msujew, line 331: Suggestion: Instead of sending the `workspace/fs/writeFile` request to the language client, we should probably instead add a `writeFile` function to the `FileSystemProvider`. The respective implementation can then take care of whether to send this to the client or use the `fs` API directly.

packages/language/src/language-server/connection-handler.ts
```diff
@@ -299,5 +309,36 @@ export function startLanguageServer(connection: Connection): void {
...
+    const CODE_UNRESOLVED_INCLUDE = "IBM3841IS";
+
+    for (const diagnostic of diagnostics) {
+      if (diagnostic.code === CODE_UNRESOLVED_INCLUDE) {
+        const action = await quickFixResolveInclude(diagnostic, connection);
+        if (action) actions.push(action);
+      }
+    }
+
+    return actions;
```
msujew: Suggestion: Can we move this into a dedicated function like the other LSP requests, that then internally does the iteration over the `diagnostics`?

packages/language/src/language-server/code-actions/quick-fix-resolve-include.ts
```diff
@@ -0,0 +1,87 @@
...
+export async function quickFixResolveInclude(
+  diagnostic: Diagnostic,
+  connection: Connection,
+): Promise<CodeAction | undefined> {
+  const unresolvedFile = diagnostic.data?.unresolvedFile;
+  if (!unresolvedFile) return undefined;
+
+  const results = await FileSystemProviderInstance.findFilesByGlob(
+    `**/${unresolvedFile}`,
+  );
```
msujew: Suggestion: If the unresolved file is something like `test_file`, but the file exists as `test_file.pli` on the file system, it won't be picked up by the glob pattern. You need to specify the list of possible file extensions in the glob as well.

packages/language/src/language-server/code-actions/quick-fix-resolve-include.ts
```diff
@@ -0,0 +1,87 @@
...
+  const procGrpsFileUri = UriUtils.joinPath(
+    pliPluginFolderUri,
+    PROC_GRPS_CONFIG_FILE,
+  );
+  const previousConfigContent =
+    await FileSystemProviderInstance.readFile(procGrpsFileUri);
+  if (!previousConfigContent) return undefined;
+
+  const procGrpsContent = JSON.parse(previousConfigContent);
+  if (!procGrpsContent) return undefined;
```
msujew: Question: Why not use the `PluginConfigurationProviderInstance`?

packages/language/src/language-server/code-actions/quick-fix-resolve-include.ts
```diff
@@ -0,0 +1,87 @@
...
+} from "vscode-languageserver-types";
+import { Connection } from "vscode-languageserver";
+
+const CONFIG_FOLDER = ".pliplugin";
+const PROC_GRPS_CONFIG_FILE = "proc_grps.json";
+
+export async function quickFixResolveInclude(
+  diagnostic: Diagnostic,
+  connection: Connection,
+): Promise<CodeAction | undefined> {
```
msujew: Suggestion: Try to isolate the generation of the quick fix from the `Connection` object. Currently, it is essentially impossible to test this function using automated tests.

packages/language/src/utils/uri.ts
```diff
@@ -61,4 +61,34 @@ export namespace UriUtils {
...
+   *
+   * If workspace-relative cannot be computed, returns the parent folder name (e.g. "test-nested-included").
+   * Returns undefined if candidate looks invalid.
+   */
+  export function computeWorkspaceRelativeParentFolder(
+    candidateRaw: string,
+    workspaceFolderUri: URI,
+  ): string | undefined {
+    if (!candidateRaw) return undefined;
+    const candidate = normalize(candidateRaw);
```
msujew, line 82: Suggestion: Calling normalize on the `candidateRaw` string as is doesn't work on Windows, it will result in garbled output. Use the `URI` as an input type instead.

packages/vscode-extension/src/language/file-system.ts
```diff
@@ -55,4 +55,11 @@ export class VSCodeFileSystemProvider implements FileSystemProvider {
   deleteFile(_uri: URI): Promise<void> {
     throw new Error("Not supported.");
   }
+  async findFilesByGlob(pattern: string): Promise<string[]> {
+    const result = (await this._connection.sendRequest(
+      Messages.FindFilesByGlob,
+      pattern,
+    )) as string[];
+    return (result || []).map((p) => p);
+  }
```
msujew: Suggestion: The corresponding `onRequest(Messages.FindFilesByGlob)` is not implemented. You can use `vscode.workspace.findFiles` for that.

packages/language/src/language-server/connection-handler.ts
Wagner-Laranjeiras: Working on that. I'm returning early if there are no diagnostics. I then iterate inside a dedicated function `applyQuickFixes` (since we might have more Quick Fixes in the future, I created a more generic  function that then calls a different function for each error code).

packages/language/src/language-server/code-actions/quick-fix-resolve-include.ts
Wagner-Laranjeiras: Yep, thanks for pointing that out. I didn't catch this. Solved already, gonna be on the next PR.

packages/language/src/language-server/code-actions/quick-fix-resolve-include.ts
Wagner-Laranjeiras: Wasn't aware of it. But using it already. Will be on the next PR too;

packages/language/src/language-server/code-actions/quick-fix-resolve-include.ts
Wagner-Laranjeiras: I'm calling my àpplyQuickFix`function and passing only the diagnostics now.

packages/language/src/utils/uri.ts
Wagner-Laranjeiras: Got it! Gonna work on that.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -0,0 +1,114 @@
...
+    searchResult = await FileSystemProviderInstance.findFilesByGlob(
+      `**/${unresolvedFile}`,
+    );
+  } else {
+    for (const extension of configExtensions) {
+      const globArg = `**/${unresolvedFile}${extension}`;
+      searchResult = await FileSystemProviderInstance.findFilesByGlob(globArg);
+      if (searchResult.length) break;
+    }
+  }
```
msujew: Suggestion: No need to send one glob pattern per file extension or make this depended on the `fileHasExtension` var. You can instead construct a single glob for this:

```ts
const globArg = `**/${unresolvedFile}{,${configExtensions.join(","}}`;
```

I.e. this will generate the following glob for an include like `%INCLUDE test`, which matches the file without an extension, the `.inc` and `.pli` extension:

```glob
**/test{,.inc,.pli}
```

Feel free to ignore this comment, see https://github.com/zowe/zowe-pli-language-support/pull/418/files#r2414227339 instead.

packages/language/src/language-server/connection-handler.ts
```diff
@@ -299,5 +310,30 @@ export function startLanguageServer(connection: Connection): void {
...
   );
+
+  connection.onCodeAction(async (params) => {
+    const diagnostics = params.context.diagnostics as Diagnostic[];
+    if (!diagnostics.length) return;
+    const actions: CodeAction[] | undefined =
+      await applyQuickFixes(diagnostics);
+    if (!actions || !actions.length) return;
+
+    return actions;
```
msujew: Suggestion: Please wrap this code in the mutex - otherwise concurrent modifications of the workspace could lead to unexpected consequences.

packages/language/src/language-server/connection-handler.ts
```diff
@@ -79,6 +86,10 @@ export function startLanguageServer(connection: Connection): void {
         renameProvider: true,
         definitionProvider: true,
         referencesProvider: true,
+        codeActionProvider: true,
+        executeCommandProvider: {
+          commands: ["pli.applyIncludeFix"],
+        },
```
msujew, line 86: Minor: Please move this into a constant.

packages/language/src/utils/uri.ts
```diff
@@ -61,4 +61,35 @@ export namespace UriUtils {
...
+   * Examples:
+   *  - workspace root: /repo/plugin-example
+   *  - candidate: /repo/plugin-example/cpy/test-nested-included/nested-not-here.pli
+   *  -> returns "cpy/test-nested-included"
+   *
+   * If workspace-relative cannot be computed, returns the parent folder name (e.g. "test-nested-included").
+   * Returns undefined if candidate looks invalid.
+   */
+  export function computeWorkspaceRelativeParentFolder(
+    candidateRaw: string,
```
msujew: Suggestion: Use a URI here instead. It's way to easy to accidentally break the `URI.parse` calculation by passing a file path in here.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -54,6 +54,7 @@ interface SerializedProgramConfig {
   "pli-options"?: PliOptions;
 }
 
+// Implement a serialize program config
```
msujew: Leftover?

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -85,6 +86,7 @@ export interface ProcessGroup {
   issueCount?: number;
 }
 
+// implement a serialize process group
```
msujew: Leftover?

packages/vscode-extension/src/language/file-system.ts
```diff
@@ -49,10 +49,22 @@ export class VSCodeFileSystemProvider implements FileSystemProvider {
...
     }
   }
-  writeFile(_uri: URI, _value: string): Promise<void> {
-    throw new Error("Not supported.");
+  async writeFile(_uri: URI, _value: string): Promise<void> {
+    try {
+      await this._connection.sendRequest(Messages.WriteFile, { _uri, _value });
+    } catch {
+      throw new Error("Couldn't write file");
+    }
```
msujew: Suggestion: There is no `Messages.WriteFile` implementation on the language client. Also, passing the `URI` object through the connection will result in a serialized `URI` object on the other end - which is rather difficult to work with.
```suggestion
  async writeFile(uri: URI, value: string): Promise<void> {
    try {
      await this._connection.sendRequest(Messages.WriteFile, uri.toString(), value);
    } catch {
      throw new Error("Couldn't write file");
    }
```

packages/vscode-extension/src/language/main.ts
```diff
@@ -37,7 +38,11 @@ class NodeFileSystemProvider implements FileSystemProvider {
     }
   }
   async writeFile(uri: URI, value: string): Promise<void> {
-    throw new Error("Not supported.");
+    try {
+      await fs.promises.writeFile(uri.fsPath, value, "utf8");
+    } catch (err) {
+      throw new Error("Couldn't write file.");
+    }
```
msujew: Suggestion:

```suggestion
    return fs.promises.writeFile(uri.fsPath, value, "utf8");
```

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -25,6 +25,15 @@ export interface FileSystemProvider {
    * Performs a file search. Implementation depends on the provider.
    */
   search(options: SearchOptions): Promise<URI | undefined>;
+
+  /**
+   * Synchronously find files matching a glob pattern
```
msujew:
```suggestion
   * Asynchronously find files matching a glob pattern
```

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -109,6 +122,14 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
...
     return undefined;
   }
+
+  /**
+   * Simulates a glob lookup in the virtual file system
+   */
+  async findFilesByGlob(pattern: string): Promise<string[]> {
+    const regex = new RegExp(pattern.replace(/\*/g, ".*") + "$", "i");
+    return Array.from(this.files.keys()).filter((fp) => regex.test(fp));
+  }
```
msujew: Suggestion: I'm still on the fence about having `glob` in the API of the `FileSystemProvider`. The issue is that this simulation is highly limited and will not work with the change I have in mind (see https://github.com/zowe/zowe-pli-language-support/pull/418/files#r2413904300). Can't we change the `search` method options to use a `global` flag? This can then internally perform a glob search on node or invoke `vscode.workspace.findFiles` in vscode.

msujew (commented):
I took a bit of time to thoroughly think about this. The most important thing to change is https://github.com/zowe/zowe-pli-language-support/pull/418/files#r2414227339.

packages/language/src/language-server/connection-handler.ts
Wagner-Laranjeiras: Got it!

packages/language/src/utils/uri.ts
Wagner-Laranjeiras: Ok, got it.

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: Was there already, will delete though.

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: Same as above.

packages/vscode-extension/src/language/file-system.ts
Wagner-Laranjeiras: Hm... ok, gonna work on this. Thanks!

packages/language/src/workspace/file-system-provider.ts
Wagner-Laranjeiras: Thanks for noticing!

packages/language/src/workspace/file-system-provider.ts
Wagner-Laranjeiras: Gonna work on it, thanks!

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -95,19 +96,21 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
...
-      if (this.files.has(fullPath)) {
-        return URI.parse(fullPath).with({ scheme: pattern.path.scheme });
-      }
+      if (this.files.has(fullPath))
+        matches.push(URI.parse(fullPath).with({ scheme: options.path.scheme }));
     }
-    return undefined;
+    if (!matches.length) return undefined;
+    else if (options.global) return matches;
+    else return matches[0];
```
Wagner-Laranjeiras: Not sure if this is what you looking for. I can return all values independently on the `options.global` value.

packages/vscode-extension/src/extension/main-browser.ts
```diff
@@ -80,4 +80,8 @@ function registerFileSystemProvider(client: LanguageClient): void {
       return result.map((r) => r[0]);
     });
   });
+  client.onRequest(Messages.WriteFile, (uriString, value) => {
+    const uri = vscode.Uri.parse(uriString);
+    vscode.workspace.fs.writeFile(uri, value);
```
msujew: Minor: Await this function call
```suggestion
    await vscode.workspace.fs.writeFile(uri, value);
```

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -24,7 +25,7 @@ export interface FileSystemProvider {
   /**
    * Performs a file search. Implementation depends on the provider.
    */
-  search(options: SearchOptions): Promise<URI | undefined>;
+  search(options: SearchOptions): Promise<URI | undefined | URI[]>;
```
msujew: Question: Do we ever use more than one `URI` returned here? If not, I would keep the return type as it is.

packages/language/src/workspace/file-system-provider.ts
msujew: Suggestion: This currently seems to disregard the `global` flag. If we search for `file.pli` in `/a/b/c/file.pli`, it won't be matched in here.

packages/vscode-extension/src/language/file-system.ts
```diff
@@ -49,9 +49,18 @@ export class VSCodeFileSystemProvider implements FileSystemProvider {
...
   }
-  writeFile(_uri: URI, _value: string): Promise<void> {
-    throw new Error("Not supported.");
+  async writeFile(uri: URI, value: string): Promise<void> {
+    try {
+      const uriAsJsonString = JSON.stringify(uri);
+      await this._connection.sendRequest(Messages.WriteFile, {
+        uriAsJsonString,
+        value,
+      });
```
msujew: Suggestion: JSON Stringifying the URI will result in a string like `{ "path": "...", "scheme": "..." ... }`, which is not parseable by `vscode.Uri.parse` just use `uri.toString()` as recommended in https://github.com/zowe/zowe-pli-language-support/pull/418#discussion_r2413943037.

You can manually test whether this is working as expected by using the `Run Web Extension` launch config.

packages/vscode-extension/src/language/main.ts
```diff
@@ -37,21 +36,25 @@ class NodeFileSystemProvider implements FileSystemProvider {
...
   }
-  async search(options: SearchOptions): Promise<URI | undefined> {
-    const path = await searchFiles(options, async (path) => {
-      const result = await fs.promises.readdir(path);
-      return result.map((file) => nPath.basename(file));
+  async search(options: SearchOptions): Promise<URI | undefined | URI[]> {
+    const GLOBAL_PATTERN = "**";
+    const pattern = options.global
+      ? `${GLOBAL_PATTERN}${options.path.fsPath}{,${options.extensions.join(",")}}`
+      : `${options.path.fsPath}{,${options.extensions.join(",")}}`;
```
msujew: Suggestion: This doesn't work on Windows, since `fsPath` returns path separators with `\`. `glob` expects the path separators to be `/`. You can use `replace(/\\/g, "/")` for that.

packages/vscode-extension/src/language/file-system.ts
```diff
@@ -49,9 +49,18 @@ export class VSCodeFileSystemProvider implements FileSystemProvider {
...
+    } catch {
+      throw new Error("Couldn't write file");
+    }
```
msujew: Suggestion: This removes all information about the underlying file system error in case something goes wrong. Just let the original error bubble up, by removing the try-catch.

packages/vscode-extension/src/extension/main-browser.ts
```diff
@@ -80,4 +80,8 @@ function registerFileSystemProvider(client: LanguageClient): void {
...
+  });
```
msujew:
```suggestion
  client.onRequest(Messages.WriteFile, async (uriString, value) => {
    const uri = vscode.Uri.parse(uriString);
    await vscode.workspace.fs.writeFile(uri, value);
  });
```

msujew (commented):
Works mostly well after a modification (on Windows). See below. I believe afterwards this is good to go 👍

msujew:
Also, please write an integration test for the quick fix. This will also validate that the virtual file system provider is capable of dealing with the `global` option.

packages/language/src/workspace/file-system-provider.ts
Wagner-Laranjeiras: Not right now. I though maybe in the future, but it would make sense to implement it just when the necessity arises

packages/language/src/language-server/connection-handler.ts
```diff
@@ -299,5 +312,34 @@ export function startLanguageServer(connection: Connection): void {
...
+    });
+  });
+
+  connection.onExecuteCommand(async (params) => {
+    if (params.command === APPLY_QUICK_FIXES) {
+      return Mutex.run(async () => {
+        const [uri, content] = params.arguments as string[];
+        try {
+          FileSystemProviderInstance.writeFile(URI.parse(uri), content);
+          connection.sendNotification(
```
montymxb: Curious if this is needed, I think the client should pick up on the change via its own listener?

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -95,18 +96,31 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
...
+      if (this.files.has(searchPath)) return options.path;
+      for (const ext of extensions) {
+        const fullPath = `${searchPath}${ext}`;
+        if (this.files.has(fullPath))
+          return URI.parse(fullPath).with({ scheme: options.path.scheme });
+      }
+    } else {
+      for (const [filePath] of this.files) {
+        if (filePath.endsWith(searchPath)) return URI.parse(filePath);
+        for (const ext of extensions) {
```
montymxb, line 122: We can optimize this as a single regex check w/ a capture to get the matching extension, but again non-blocking in this case. In practice this should work pretty quick, but I would leave a mark that this is a candidate for optimization when we have time.

packages/vscode-extension/src/language/main.ts
```diff
@@ -37,21 +36,25 @@ class NodeFileSystemProvider implements FileSystemProvider {
...
+      nodir: true,
+      absolute: true,
+      nocase: true,
     });
-    if (path) {
-      return options.path.with({ path });
-    } else {
-      return undefined;
-    }
+    const result = URI.file(files[0]);
```
montymxb, line 56: Need to do a sanity check here on the files array, particularly when it's empty.

montymxb (commented):
Alrighty this looks nice! There are a few notes here, mainly I'm not seeing a diagnostic for a missing include file, which is related to trying to access a file that doesn't exist from in the return from glob, which is noted. Beyond that the rest is looking good on mac, but I didn't get to check on windows yet.

packages/language/src/language-server/connection-handler.ts
Wagner-Laranjeiras: I was having some erratic behavior when adding the new folder to the libs, so it was more of a double measure. I'm removing it though (after double checking)

Wagner-Laranjeiras:
Worked on the review, but gonna only push it Monday morning after revisiting it with a fresh set of eyes. Thanks for the support!

Wagner-Laranjeiras:
Got another bug during testing. Please don't merge it yet, but feel free to review if you have the time.

montymxb (approved):
:octocat: Approved! Did another check over things, but this looks pretty good to me. You'll need to rebase on development as well as probably lint, beyond that should be good to go.

packages/language/test/lsp/apply-quick-fixes.test.ts
```diff
@@ -0,0 +1,133 @@
...
+import {
+  deserializeProcessGroup,
+  PluginConfigurationProviderInstance,
+} from "../../src/workspace/plugin-configuration-provider";
+import { URI, UriUtils } from "../../src/utils/uri";
+
+// Mock dependencies
+vi.mock("../../workspace/file-system-provider");
+vi.mock("../../workspace/plugin-configuration-provider");
+vi.mock("../../utils/uri");
```
msujew, line 24: Suggestion: Mocking is usually an anti-pattern that hints at badly designed interfaces or too tight coupling between systems. I don't think mocking is even required at all for the file system provider - simply try to feed the data you need into the virtual file system provider, and let its `search` function handle the rest.

If you place the program config files in the file system provider, the `plugin-configuration-provider` mocks also become redundant.

msujew (commented):
I'm fine with the changes, and feel free to merge this. However, I would appreciate a follow-up that cleans up the tests, see https://github.com/zowe/zowe-pli-language-support/pull/418#discussion_r2426686812. (also remove the draft status from the PR)

packages/language/test/lsp/apply-quick-fixes.test.ts
Wagner-Laranjeiras: Got it! Gonna follow this up and create a new PR once it's refactored. Thanks for the feedback.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2022,10 +2026,10 @@ async function resolveIncludeFileUri(
         libFileUri = UriUtils.joinPath(libUri, item.fileName);
       }
 
-      const match = await FileSystemProviderInstance.search({
+      const match = (await FileSystemProviderInstance.search({
         path: libFileUri,
         extensions: pgroup.includeExtensions,
-      });
+      })) as URI | undefined;
```
msujew: Is the casting even needed? `search` always returns `URI | undefined`.

Wagner-Laranjeiras: Removed and pushed

PR #420, opened 2025-10-01
Support `REPLACE` statement

msujew:
Related to #372. Implements support for the `%REPLACE` statement. Does this by effectively simulating a DECLARE + assignment statement.


packages/language/test/fourslash/preprocessor/replace-with-string.ts
Lotes, line 1: I was wondering if you could replace inside the replaced text. But that does not work it seems, since the name is replaced by something with quotes or as some number with a gapp in front of it.

```
 %REPLACE X BY "A";
 %REPLACE A1 BY 123;
 DCL xyz CHAR INIT(X%;1);
```

```
 %REPLACE B BY 11111;
 %REPLACE A11111 BY 123;
 DCL xyz CHAR INIT(A%;B);
```

Maybe you have some idea to break it.

The implementation is smart and pragmatic (by using declaration and assignment).

msujew: Right, AFAIK `%REPLACE` really only generates string/number constants, not IDs. Our implementation might still not be 100% correct, but I think we should only worry about that once we actually get bug reports :)

Lotes (approved):

PR #421, opened 2025-10-01
Simplify token type creation

msujew:
Does as the PR says. Massively simplifies the token type creation process, eliminating ~1.000 LoC in the process.

### DIFF
--- a/packages/language/src/parser/tokens.ts
+++ b/packages/language/src/parser/tokens.ts
@@ -9,16 +9,7 @@
  *
  */
 
-import {
-  createToken,
-  CustomPatternMatcherFunc,
-  Lexer,
-  TokenType,
-} from "chevrotain";
-import {
-  CompilerOptions,
-  getDefaultCompilerOptions,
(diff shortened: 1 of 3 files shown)


packages/language/src/parser/tokens.ts
Lotes, line 1: Nice, how have you done it?

msujew: Using the original `createToken` function from Chevrotain created a lot of overhead. Adding an abstraction layer on top of it eliminated most of that.

ssmifi (approved):
LGTM. Thanks. 
By the way: I noticed that there is a `lexing.test.ts` (with "CompilerOptions parser" as description) and a `parser/pli-lexer.test.ts`. Should we merge them?

msujew:
@ssmifi I've renamed the test in `lexing.test.ts`. The test cases specifically test that the lexer is able to take the compiler options into account, but is unrelated to the parsing of compiler options.

PR #422, opened 2025-10-01
Primitive type system integration

Lotes:
This PR adds a type system by:

* describing types using the `descriptions.ts`
  * there are a lot of attributes ordered by "attribute kind"
* handling type construction with the `type-builder.ts`
* handling arithmetic operations according to the documentation in `arithmetic-operations.ts`
* creating a type inference service in `infer.ts` with two entry points: one for expressions and one for declarations
* adding a type cache that gets reset on lifecycle beginning in `type-cache.ts`

I also added:

* a `services` field in the `CompilationUnit`. A place where we can put and find unit-specific services like the file store or the type inferer
* added a `README` for the progress tracking of the expressions that need to be covered later
* added builtin constants for `TRUE` and `FALSE`
* several tests that cover arithmetic operations like given in the documentation
* some tests for building types (we have > 100 attributes, covering only some combinations). Testing every combination seems impossible. But I tested all the diagnostics I have introduced.

One of the latest changes was a cleanup of the code by introducing sorted attributes and groups switch cases in the type builder. I also added documentation links in redundant positions, so we have a change to quickly look it up.


packages/language/src/typesystem/descriptions.ts
```diff
@@ -0,0 +1,1061 @@
...
+  StringKind,
+  StringFormat,
+  StringLength,
+  FloatFormat,
+  AccessMode,
+  BufferMode,
+  FileUsage,
+}
+
+export const AttributeKinds: AttributeKind[] = [
```
Lotes, line 110: Arrays, Structs and Unions (?) will be integrated in a follow up PR

packages/language/src/typesystem/descriptions.ts
```diff
@@ -0,0 +1,1061 @@
...
+    format: StringFormat.NonVarying,
+    length: 1,
+  });
+  export const isBoolean = (
+    type: TypeDescriptions.Any,
+  ): type is StringTypeDescription =>
+    isString(type) && type.kind === StringKind.Bit && type.length === 1;
+
+  //TODO check default values
+  const DefaultValues: AttributeTypes = {
```
Lotes, line 966: There are several places in this file with default values. A opportunity to collect them here.

scripts/generate-typesystem-table.mts
```diff
@@ -0,0 +1,19 @@
+import { readFile } from "fs/promises"
+
+const text = await readFile("packages/language/src/syntax-tree/ast.ts", "utf8");
```
Lotes: Can be removed at the end of this PR, I think

Lotes:
Made ready. I do not want this PR to become bigger.

packages/language/test/fourslash/validate/typesystem/if-true.ts
```diff
@@ -0,0 +1,21 @@
...
+/// <reference path="../../framework.ts" />
+
+// @wrap: main
+//// DCL ANYTHING FIXED;
+//// IF TRUE THEN
+////   ANYTHING = 0;
+//// IF FALSE THEN
+////   ANYTHING = 1;
+
+verify.noDiagnostics();
```
ssmifi: _Suggestion_: We try to avoid general `noDiagnostics` tests, because they easily fail for unrelated reasons and aren't very specific. I would at least add a range marker. If you agree, we can also extend the harness, so that we can exclude specific diagnostics more easily. Some thing like `verify.noDiagnostics(codes.typesystem);` would be nice. Wdyt?

Lotes: Sounds like a good idea, I think. I understand the problem.

packages/language/src/typesystem/type-builder.ts
```diff
@@ -0,0 +1,734 @@
...
+    const token = attribute.typeToken!;
+    assertType<ast.DefaultAttribute>(attribute.type);
+    switch (attribute.type) {
+      /**
+       * Data type attributes
+       */
+      case "TASK":
+      case "FILE":
+      case "FORMAT":
+      case "AREA": {
```
msujew: Suggestion: Before continuing on that, can we refactor this to use an enum? The strings are pretty error-prone and difficult to maintain (they are an artifact of migrating the project from Langium). Ideally, we would have a mapping from token type -> `DefaultAttribute` enum value in the parsing phase.

packages/language/src/workspace/compilation-unit.ts
```diff
@@ -132,10 +143,20 @@ const getRootPreprocessorScope = createBuiltinScopeGetter(
...
 ): Promise<CompilationUnit> {
+  const compilerOptions = getDefaultCompilerOptions();
+  const services: CompilationServices = {
+    files: new FileStore(),
+    typeCache: new DefaultTypeCache(),
+    inferer: new DefaultTypeInferer(
+      (compilerOptions.rules?.ibm ?? "ANS") === "ANS"
+        ? CompilerOptionRules.ANS
+        : CompilerOptionRules.IBM,
+    ),
```
msujew, line 154: Suggestion: I don't think it's a good idea to provide the value as a constructor argument to the service - AFAIK the `IBM` compiler option can be changed during the lifetime of the `CompilationUnit`.

msujew (commented):
Generally good, see below for a few improvements.

packages/language/test/fourslash/validate/typesystem/if-true.ts
msujew: I agree. Once we start adding validations like "unused variable", quite a few of these tests will start failing.

packages/language/src/typesystem/type-builder.ts
Lotes: Sounds like a good idea. Never thought of that.

packages/language/src/parser/token-mappings.ts
msujew, line 1: That's perfect, thank you :)

packages/language/src/validation/pli-validator.ts
```diff
@@ -375,6 +378,10 @@ export class PliValidator implements Validator {
   }
 
   private knownBuiltins = new Set([
+    //TODO is this the correct place???
+    "TRUE",
+    "FALSE",
```
msujew: This is the correct place, yes :)

packages/language/src/workspace/compilation-unit.ts
Lotes: I think we can improve it in an upcoming PR. The change is not so difficult, but makes no sense right now, because it is not really used.

packages/language/src/typesystem/descriptions.ts
Lotes: This is actually difficult: some attributes imply other default values and some compiler options or architecture can also influence the default values. Not sure if we should solve this problem now.

msujew (approved):
Thanks, that looks really good to me 👍

packages/language/src/workspace/compilation-unit.ts
msujew: Sure, fine with me 👍

packages/language/src/parser/token-mappings.ts
```diff
@@ -0,0 +1,327 @@
+import { TokenType } from "chevrotain";
+import type { DefaultAttribute as DefaultAttributeType } from "../syntax-tree/ast";
+import * as tokens from "./tokens";
+
+export enum DefaultAttributeEnum {
```
msujew, line 16: Suggestion: Can we actually move this to the `ast.ts` and completely remove the string-typed stuff? Or should we do this in a separate PR?

Lotes: The string-typed stuff has the following intent: If the parser "grammar" is not finished, we will add missing attributes later but may forget to add them in the mapping. The strings are used to ensure that the entire grammar is covered by using the string type "DefaultAttribute" as key for the Record type. Every time the grammar gets a missing attribute more, the Typescript compiler will complain about a missing mapping.

I would like to keep the string-typed stuff. The merging of the attribute mapping happens only once at import time.

The file is also quite big, I would like to have it separated from the AST, because it has a different responsibility.

Lotes:
TODO Need to add the license headers.

packages/language/src/parser/token-mappings.ts
Lotes: I tend to do this in a separate PR.

PR #423, opened 2025-10-01
Support preprocessor SELECT statement

msujew:
Related to https://github.com/zowe/zowe-pli-language-support/issues/372.

Implements the `SELECT` preprocessor statement:

* Refactors the `IfInstruction` to a `SelectInstruction` that is used by both `IfStatement` and `SelectStatement`.
* Implements skipped code support for the select statements
* Adds a bunch of tests


packages/language/src/preprocessor/instructions.ts
```diff
@@ -156,12 +156,29 @@ export interface NoteInstruction {
...
+  return {
+    kind: InstructionKind.Select,
+    element,
+    compare,
+    cases,
+  };
+}
+
+export interface Cases {
+  conditions: ExpressionInstruction[];
```
Lotes, line 188: Empty conditions === OTHERWISE?
May add a JS Doc comment with this hint to this field.

packages/language/src/preprocessor/instruction-generator.ts
```diff
@@ -133,7 +133,10 @@ function generateInstructionForStatement(
       instruction = generateReturnInstruction(value);
       break;
     case ast.SyntaxKind.IfStatement:
-      instruction = generateIfInstruction(value, context);
+      instruction = generateSelectInstructionFromIf(value, context);
```
Lotes, line 136: Removing the IF from the generator/instructions/interpreter...
I have no real counter argument against it, it just feels a bit weird.

I think this would be a bad idea if we would have FLOW-specific differences between if and select statements. But I do not know any.

packages/language/src/language-server/skipped-code.ts
```diff
@@ -83,24 +82,51 @@ export function skippedCodeRanges(
...
+            start: textDocument.positionAt(unitRange.start),
+            end: textDocument.positionAt(unitRange.end),
+          });
+        }
+      }
+    } else if (
+      token.kind === CstNodeKind.SelectStatement_SELECT &&
+      token.element?.kind === SyntaxKind.SelectStatement
+    ) {
+      const element = token.element;
```
Lotes, line 114: Crazy feature :D

Lotes (approved):
LGTM. Good idea with merging if and select. One little remark about documenting `conditions`.

packages/language/src/preprocessor/instructions.ts
msujew: Yes, I've added some documentation 👍

packages/language/src/preprocessor/instruction-generator.ts
msujew: AFAIK I don't think there's any meaningful difference when it comes to the interpreter between those two statements.

Issue #394, opened 2025-09-08
Includes local resolution


asatklichov:
Currently local resolution of includes not fully working for "absolute path" (on Windows, and Linux file systems)
Includes should be resolved automatically, recursively (should look child folders) n 'proc_grps.json' 
 

See example below: 

<img width="715" height="152" alt="Image" src="https://github.com/user-attachments/assets/71b5e78e-a88f-4c9d-ac19-187409902c4d" />

montymxb:
Hey @asatklichov , hope all is well. Could you clarify with a quick PLI code example here as well? From what I'm reading it sounds like this is about resolving an include via a literal absolute path, ex. `%INCLUDE "/a/b/c`, but I'm also not too sure based on what you've written up. If you can give some additional context that would definitely help.

asatklichov:
Hi @montymxb   path are defined in config file,   

 "libs": [
        "cpy",
        "inc",
        "C:/__che4z-project/zelda/_dev_BAK_ilki/ca11",
        "C:/__che4z-project/zelda/_dev_BAK_ilki/ca11/AS892333.TEST4ZDB.APP.DBRMLIB",
      ],

or Linux path 'a/b/c/AS892333.TEST4ZDB.APP.DBRMLIB' 


then we can use it as normally `%INCLUDE RCUSDB;`

montymxb:
I see that you wrote those up under `include-extensions`, did you mean `libs`? If so then I can see that you're talking about wanting to have those absolute & relative paths incorporated as part of the lookup for libs, if I'm following correctly there?

I believe the relative path lookup is already supported as is via the libs property in the program configuration, but we'll check both up front anyways to see what the state currently is before working on this one.

Separately if we have something like `C:/a/b/c/AS892333.TEST4ZDB.APP.DBRMLIB` under libs we'll need to split it up as `C:/a/b/c/` and `.DBRMLIB` (first for the libs, second for the include extensions) to get the desired effect. That's assuming `DBRMLIB` is a known good extension for includes.

montymxb:
Following up after chatting about this. There's 3 points that need addressing:

- [x] `libs` should support absolute paths (need to double check & add if needed)
    - Ex. `"libs": ["C:/__che4z-project/zelda/_dev_BAK_ilki/ca11"]`, same as above.

- [x] `libs` entries should implicitly include all child folders as well
[...]

PR #424, opened 2025-10-02
Resolve Absolute Lib Paths

montymxb:
Resolves the first point of #394 by modifying the `setFilePath` logic to compute the `libFileUri` based on whether or not the current lib path is absolute or not (unix & windows in this case). Testing is setup for the `/` case but not for windows, given the behavior for handling drive letters only kicks in when `win32` is present. I did do a manual check for that case since that was the issue was originally reported for windows drive letters in the first place. But please feel free to do a secondary check on windows for that case.

There's a small existing issue in the hover logic on windows based on the hover logic computation when a lib is absolute but _not_ a windows drive letter. This leads to a mismatch between the from/to uris when the lib entry is something like `/A/...` instead of `C:/A/...`. The path still resolves in the context of the current drive letter, but on hover displays as `./A/...`. I'm leaving this alone for now, since it involves tweaking the relative path computation logic in a way that feels problematic, and doesn't impact functionality at the time.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -1938,10 +1938,7 @@ async function resolveIncludeFileUri(
         );
       } else {
         // use lib path over workspace
-        const libUri = URI.from({
-          scheme: "file",
-          path: lib,
-        });
+        const libUri = URI.file(lib);
         libFileUri = UriUtils.joinPath(libUri, item.fileName);
       }
 
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -1926,7 +1926,7 @@ async function resolveIncludeFileUri(
 
   if (pgroup) {
     // lib file as either a string or a member from a known process group
-    const absPathRegex = /^\/|\\|[A-Z]:/i;
+    const absPathRegex = /^(?:\/|\\|[A-Z]:)/i;
     for (const lib of pgroup.libs) {
       let libFileUri: URI;
       if (!absPathRegex.test(lib)) {
@@ -1938,7 +1938,9 @@ async function resolveIncludeFileUri(
         );
       } else {
         // use lib path over workspace
-        const libUri = URI.file(lib);
+        const libUri = URI.file(lib).with({
+          scheme: context.entryUri.scheme,
+        });
         libFileUri = UriUtils.joinPath(libUri, item.fileName);
       }
 
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -1759,6 +1759,14 @@ async function runIncludeInstruction(
   }
 }
 
+/**
+ * Sets the filePath & relativeFilePath of the given item, if a filePath is provided
+ * The relativeFilePath is calculated based on the currentUri in the context
+ *
+ * @param item Item to set the file paths for
+ * @param filePath File path to set, if null, no changes are made
+ * @param context Used for currentUri to calculate relative paths
+ */
 function setFilePath(
   item: { filePath: string | null; relativeFilePath: string | null },
   filePath: string | null,
@@ -1918,12 +1926,25 @@ async function resolveIncludeFileUri(
 
   if (pgroup) {
     // lib file as either a string or a member from a known process group
+    const absPathRegex = /^\/|\\|[A-Z]:/i;
     for (const lib of pgroup.libs) {
-      const libFileUri = UriUtils.joinPath(
-        URI.parse(PluginConfigurationProviderInstance.getWorkspacePath()),
-        lib,
-        item.fileName,
-      );
+      let libFileUri: URI;
+      if (!absPathRegex.test(lib)) {
+        // relative lib path, combine w/ workspace
+        libFileUri = UriUtils.joinPath(
+          URI.parse(PluginConfigurationProviderInstance.getWorkspacePath()),
+          lib,
+          item.fileName,
+        );
+      } else {
+        // use lib path over workspace
+        const libUri = URI.from({
+          scheme: "file",
+          path: lib
+        });
+        libFileUri = UriUtils.joinPath(libUri, item.fileName);
+      }
+
       const match = await FileSystemProviderInstance.search({
         path: libFileUri,
         extensions: pgroup.includeExtensions,
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -244,8 +244,8 @@ export interface EvaluationResults {
 
 interface InterpreterContext {
   unit: CompilationUnit;
-  currentUri?: URI;
-  entryUri?: URI;
+  currentUri: URI;
+  entryUri: URI;
   /**
    * These are global variables that are defined on the root level of the preprocessor
    */
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -1940,7 +1940,7 @@ async function resolveIncludeFileUri(
         // use lib path over workspace
         const libUri = URI.from({
           scheme: "file",
-          path: lib
+          path: lib,
         });
         libFileUri = UriUtils.joinPath(libUri, item.fileName);
       }
--- a/packages/language/test/fourslash/preprocessor/include/from-absolute-lib-path.ts
+++ b/packages/language/test/fourslash/preprocessor/include/from-absolute-lib-path.ts
@@ -21,7 +21,6 @@
 ////   ]
 //// }
 
-
 // @filename: .pliplugin/proc_grps.json
 //// {
 ////     "pgroups": [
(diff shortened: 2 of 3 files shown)


packages/language/test/fourslash/preprocessor/include/from-absolute-lib-path.ts
```diff
@@ -0,0 +1,47 @@
...
+////             "libs": [
+////                 "/tmp/cpy"
+////             ],
+////             "include-extensions": [
+////                 ".pli"
+////             ]
+////         }
+////     ]
+//// }
+
```
Lotes, line 40: Can you provide some comments what the actual subject to test is and what expectations you have. It is not clear to me.
Plus, more test cases might be good to cover the other if-branches as well:

* something with Windows drives and \ (when I interpret the RegExp correctly)
* something with non-absolute paths

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1918,12 +1926,25 @@ async function resolveIncludeFileUri(
...
     for (const lib of pgroup.libs) {
-      const libFileUri = UriUtils.joinPath(
-        URI.parse(PluginConfigurationProviderInstance.getWorkspacePath()),
-        lib,
-        item.fileName,
-      );
+      let libFileUri: URI;
+      if (!absPathRegex.test(lib)) {
+        // relative lib path, combine w/ workspace
+        libFileUri = UriUtils.joinPath(
```
Lotes, line 1934: Is the scheme her also "file"? Why do you need to specify it in line 1942 and not also here?

packages/language/test/fourslash/preprocessor/include/from-absolute-lib-path.ts
montymxb: I can add some more details here comment-wise, can double check the `\` case too.

The windows drives are hard to test here. I covered it at first, but saw we alternate handling based on the current OS, so as far I can tell we can't really test that part in an OS independent fashion, at least not without tweaking things further.

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: Yeah I believe it's loaded with the file scheme implicitly. Actually the one on line 1942 can be changed to `URI.file` to avoid needing to provide it explicitly actually. It's possible it's using the same call internally as well.

packages/language/test/fourslash/preprocessor/include/from-absolute-lib-path.ts
Lotes: We have a Windows build stage, there we could test it.
I think the vitest test can be added with a condition. But not sure if we should do this...

msujew: I agree, we can do something like `test.runIf(process.platform === 'win32')` or simply simulate the Windows case using the virtual file system provider.

montymxb: That's my general hesitation as well @Lotes , cross-platform testing is fickle. But if you're both on board with adding that in I'm happy to put that in. I did have a test for it before via 4slash (before I backed it out), but as you've both pointed out it's easier to achieve with vitest. I'll go for that over simulation I think.

montymxb:
Still need to come back to this one, just got caught up with a few other things at the moment after being out of office. I'll either get back to it later on Friday or Monday next week.

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: Resolved in upcoming commit.

packages/language/test/fourslash/preprocessor/include/from-absolute-lib-path.ts
montymxb: Ongoing resolution by testing for each OS independently. Will resolve once I'm happy with the results.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1976,12 +1926,22 @@ async function resolveIncludeFileUri(
 
   if (pgroup) {
     // lib file as either a string or a member from a known process group
+    const absPathRegex = /^\/|\\|[A-Z]:/i;
```
msujew: Suggestion: The `^` is only matched in the first alternative

```suggestion
    const absPathRegex = /^(\/|\\|[A-Z]:)/i;
```

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1976,12 +1926,22 @@ async function resolveIncludeFileUri(
...
+      if (!absPathRegex.test(lib)) {
+        // relative lib path, combine w/ workspace
+        libFileUri = UriUtils.joinPath(
+          URI.parse(PluginConfigurationProviderInstance.getWorkspacePath()),
+          lib,
+          item.fileName,
+        );
+      } else {
+        // use lib path over workspace
+        const libUri = URI.file(lib);
```
msujew: Minor: We should support non-file schemes here as well. This might happen in the web extension for example.

```suggestion
        const libUri = URI.file(lib).with({
          scheme: context.entryUri.scheme
        });
```

packages/language/test/preprocessor/pli-preprocessor.test.ts
```diff
@@ -0,0 +1,143 @@
...
+ * @param expectedTokens Expected array of tokens (by image) to be found in the resulting compile unit once done
+ */
+async function expectTokensForWorkspace(
+  mainFileContent: string,
+  libFiles: Array<[string, string]>,
+  expectedTokens: string[],
+): Promise<void> {
+  for (const [path, content] of libFiles) {
+    FileSystemProviderInstance.writeFile(URI.file(path), content);
+  }
```
msujew, line 72:
```suggestion
  for (const [path, content] of libFiles) {
    await FileSystemProviderInstance.writeFile(URI.file(path), content);
  }
```

packages/language/test/preprocessor/pli-preprocessor.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      // also test absolute path resolution w/ windows too
+      await init("/test/libs/");
+
+      await expectTokensForWorkspace(
+        ` %INCLUDE "lib.pli";`,
+        [["/test/libs/lib.pli", ` DECLARE LIB_VAR FIXED;`]],
+        ["DECLARE", "LIB_VAR", "FIXED", ";"],
+      );
+    },
+  );
```
msujew: Suggestion: Can't this test be combined with the other one? We essentially want to test: unix-style absolute paths are always supported, even on Windows (perhaps add this as a comment).

msujew (approved):
Looks good to me, just a few comments, see below.

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: Good catch

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: I'm assuming the entryUri will always be set when we're in a compile unit? I can see it's optional and I'm curious about that.

packages/language/test/preprocessor/pli-preprocessor.test.ts
montymxb: Yeah it can now, previously I had a version of this that was testing something else on windows, I'll consolidate that.

packages/language/test/preprocessor/pli-preprocessor.test.ts
montymxb: Added in commit

packages/language/src/preprocessor/instruction-interpreter.ts
msujew: Feel free to remove the `?` from the `entryUri` and `currentUri` fields. I probably made those optional under some old assumptions that no longer hold up.

Note that the function will already have returned at this point if the `entryUri` is not set.

montymxb: Gotcha, I'll tweak those declarations as well before pulling this in.

packages/language/test/preprocessor/pli-preprocessor.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  afterEach(() => {
+    setFileSystemProvider(undefined);
+    setPluginConfigurationProvider(undefined);
+    PluginConfigurationProviderInstance.setProgramConfigs("", []);
+    PluginConfigurationProviderInstance.setProcessGroupConfigs([]);
+  });
+
+  // macOS + linux absolute path resolution
+  test.runIf(["darwin", "linux", "win32"].includes(process.platform))(
```
Lotes, line 102: The more platform-specific tests we get, we should think about a helper functions or at least OS-enums, I think. On the other hand it is just `process.platform`, there is not much a helper function can help with. Only with getting the OS string right/typed.

montymxb: I agree. Let's keep an eye on it going forward.

PR #426, opened 2025-10-06
Fix issues in `%SELECT` statement

msujew:
Fixes a few issues related to the `%SELECT` statement that I've noticed in manual testing:

1. While writing an unfinished `%SELECT` statement, the language server crashed on me due to usage of `!` in the preprocessor code (no test, couldn't find a good way to reproduce).
2. Empty `%WHEN` statement bodies were incorrectly shown as "skipped" (this has a test case).

Also adds a `not` inversion feature to the `TestBuilder`, which makes negative testing a lot easier. Currently, it's only active for the `preprocessor` features, but we can extend it to other features in the future.


ssmifi (approved):
Nice. LGTM.

Issue #410, opened 2025-09-18
Remove support for `%SKIP` statements


asatklichov:
Since PL/I compiler is not supporting  %SKIP (N); even mentioned in specification, let's disable this support. 

We have alternative support %DO SKIP;

 
<img width="217" height="281" alt="Image" src="https://github.com/user-attachments/assets/39f973d3-3d4d-42e4-ba26-50520c881311" />

msujew:
Even though the compiler doesn't seem to be aware of the statement, we should keep the parser feature for it. We should instead:

* Remove the skip feature from the "skipped lines" feature (i.e. no longer display the lines as skipped in the editor).
* Add a custom warning saying that the SKIP statement is not supported by the language server.

asatklichov:
done

PR #427, opened 2025-10-07
Refactor preprocessor parser

msujew:
Refactors the `PreprocessorParserState` to a generic `ParserState` that can eventually be used for both parsers. Performs a few changes to improve error recovery behavior and better splitting of the code to support CICS and SQL preprocessors later on (with a few comments explaining how.

Also fixes a few minor issues related to parsing of the `%END` statement (mainly related to labels before and after the statement, which previously led to parser errors).

Deals with https://github.com/zowe/zowe-pli-language-support/issues/410 partially, by removing the functionality of the `%SKIP` statement (but keeps it parseable).


VitGottwald:
Big change. Do we have some tests that measure performance on PRs to make sure we are not accidentally degrading it?

msujew:
@VitGottwald Not yet, it would be nice to have something that does a few benchmarks on GitHub's hardware and then posts a result in the PR. I'm pretty sure these changes don't impact the performance at all, since they don't actually touch the main parsing step (still done in Chevrotain) yet. I'll create some performance benchmark soon for that.

packages/language/src/parser/parser-entry.ts
```diff
@@ -0,0 +1,63 @@
...
+    } else {
+      // TODO: Handle other preprocessors (SQL, CICS) here!
+      // state.canConsume(EXEC_SQL/DFHRESP/EXEC_CICS/...)
+
+      // Otherwise construct a token statement
+      statements.push(consumeTokenStatement(state));
+    }
+  }
+  for (const statement of statements) {
+    recursivelySetContainer(statement);
```
ssmifi: _Suggestion_: Set the containers in the first loop when the `stmt` is pushed, so that we can avoid the second loop.

packages/language/src/parser/parser-state.ts
```diff
@@ -0,0 +1,242 @@
...
+  private inProcedure = false;
+
+  constructor(tokens: t.Token[], mode: ParserStateMode) {
+    this.tokens = tokens;
+    this.mode = mode;
+    this.diagnostics = [];
+    this.index = 0;
+  }
+
+  pushProcedure(): void {
```
ssmifi: _Minor_: Naming might be confusing if the assumption is that procedures cannot be nested. 
Is this case handled (gracefully)?

packages/language/src/parser/parser-state.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  tryConsume(
+    element: SyntaxNode | undefined,
+    kind: CstNodeKind | undefined,
+    tokenType: TokenType,
+  ): t.Token | null {
+    if (!this.canConsume(tokenType)) {
+      return null;
+    }
+    const token = this.token!;
```
ssmifi, line 169: _Suggestion_: If `canConsume` returns a type guard like `this is ParserState & { token: t.Token }`, we can ditch the non-null assertion here.

packages/language/src/parser/parser.ts
```diff
@@ -24,17 +24,22 @@ export class PliParser extends AbstractParser {
...
   }
 
-  parse(): ast.PliProgram {
-    const result = this.PliProgram();
+  parse(): ast.Program {
+    let result: ast.Program | undefined;
+    try {
+      result = this.PliProgram();
+    } catch {
+      // Sometimes, this happens
```
ssmifi: _Minor_: Can we be more specific here? Do we know, when this fails?

packages/language/test/fourslash-harness/harness-interface.ts
```diff
@@ -18,12 +18,14 @@ import { SemanticTokenTypes } from "vscode-languageserver-types";
...
 import { CompilerOptionsCodes } from "../../src/preprocessor/compiler-options/codes";
 import { CompilerOptions } from "../../src/preprocessor/compiler-options/options";
+import { SyntaxKind } from "../../src/syntax-tree/ast";
 
 type SemanticTokenTypesValues = `${SemanticTokenTypes}`;
 
 export type Not<T> = Omit<T, "not">;
 
 export interface HarnessTesterInterface {
+  Syntax: typeof SyntaxKind;
```
ssmifi, line 30: Nice! :)

packages/language/src/parser/preprocessor-parser.ts
```diff
@@ -0,0 +1,1455 @@
...
+          state.popProcedure();
+        }
+        break;
+      case t.RETURN.tokenTypeIdx:
+        unit = returnStatement(state);
+        break;
+      case t.NOTE.tokenTypeIdx:
+        unit = noteStatement(state);
+        break;
+      case t.REPLACE.tokenTypeIdx:
```
ssmifi, line 232: _Minor_: I would order the cases alphabetically for maintainability.

ssmifi (commented):
Looks quite nice. Thank you. I just have a few minor questions.

packages/language/src/parser/parser-state.ts
Lotes: Suggestion: enter/leaveProcedure?

packages/language/src/parser/parser-state.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  pushProcedure(): void {
+    this.inProcedure = true;
+  }
+
+  popProcedure(): void {
+    this.inProcedure = false;
+  }
+
+  peek(la: number): t.Token | undefined {
```
Lotes, line 67: May add a JSDoc that 1 is the minimum. Or even a guard with a throw?
The value 0 would be the last token? Feels weird.

packages/language/src/parser/tokens.ts
```diff
@@ -18,7 +18,11 @@ export interface Token {
   image: string;
   originalImage: string;
   startOffset: number;
+  startLine: number;
```
Lotes, line 21: Why adding these fields here?

packages/language/test/fourslash-harness/execute.test.ts
```diff
@@ -253,5 +260,5 @@ if (!HARNESS_TEST_FILE) {
   }
 
   const fullPath = path.resolve(HARNESS_TEST_FILE);
-  runSingleHarnessTest(fullPath);
+  runSingleHarnessTest(fullPath, 600_000); // 10 minute timeout for debugging
```
Lotes, line 263: Maybe remove it or add an environment variable. Otherwise we will have long running tests in the pipeline.

`env.DEBUG ? 600_000 : undefined`

msujew: This code is only executed if the `HARNESS_TEST_FILE` env variable exists, which is exclusively set during debugging with the respective launch config.

packages/language/src/parser/tokens.ts
msujew: We need them to recover to the next line:

https://github.com/zowe/zowe-pli-language-support/blob/70dd2eac5868d7bef0b3c5cfec9c87820b89e777/packages/language/src/parser/parser-state.ts#L140-L143

packages/language/src/parser/parser-state.ts
msujew: It's the same semantics as Chevrotain, with its `LA` method.

packages/language/src/parser/parser-state.ts
msujew: While this is a good idea in general, in this case I would disagree. That would mess with the type system in ways that are not good:

```ts
if (state.canConsume(someTokenType)) {
  state.consume(someTokenType);
  const nextToken = state.token // <-- this token is typed as "Token", even though it might be undefined
}
```

packages/language/src/parser/parser.ts
msujew: I've added a better comment.

ssmifi (approved):
LGTM

Issue #414, opened 2025-09-18
Change error-message  type for implicit variables and compiler options


asatklichov:
- [x] 1. **Downgrade** the message to warning for implicit variables. 

<img width="609" height="163" alt="Image" src="https://github.com/user-attachments/assets/2663194f-39bb-40e1-8cc1-d62160523fb1" />

- [ ] 2. **Upgrade** compiler options messages to ERROR if it is really an error.  

<img width="679" height="153" alt="Image" src="https://github.com/user-attachments/assets/58bc7429-3070-4621-8e9f-1bfd2913a7b5" />

- [x] 3. Leave 'Duplicate compiler option ' as it is: WARNING  (no change)

PR #428, opened 2025-10-08
Downgrade implicit vars error to warning

ssmifi:
Fixes https://github.com/zowe/zowe-pli-language-support/issues/414 1)


msujew (approved):
Makes sense to me, thanks 👍

PR #429, opened 2025-10-08
Improve compiler option reporting

### LINKED ISSUE #374: Findings 28.8

### LINKED ISSUE #414: Change error-message  type for implicit variables and compiler options

### LINKED ISSUE #419: [Feature] Add reporting for duplicate and exclusive compiler suboptions
Currently, only  top level compiler options are checked for dupes and mutexes. In the near future, also duplicated or excluding settings within one option should be reported as a warning. Usually, this is still valid on the mainframe and the last option takes effect.

ssmifi:
Upgrades most of the compiler option codes to `Error`.

Fixes https://github.com/zowe/zowe-pli-language-support/issues/414 (compiler option errors are errors)
Fixes https://github.com/zowe/zowe-pli-language-support/issues/419 (dupes and mutex for sub options) and therefore also fixes https://github.com/zowe/zowe-pli-language-support/issues/374 b

Also removes old compiler option case sensitivity tests.


packages/language/src/preprocessor/compiler-options/codes.ts
```diff
@@ -104,30 +104,46 @@ export const CompilerOptionsCodes = {
...
+    fullCode: "COOP13W",
+  } as ParametricPLICode,
+
+  MutexSubOptionIssue: {
+    code: "COOP14",
+    severity: Severity.W,
+    message: (name: string) =>
+      `Mutually exclusive compiler sub options found for ${name}, only the last one will take effect.`,
+    fullCode: "COOP14W",
+  } as ParametricPLICode,
```
msujew: Question: Does it perhaps make more sense to use the default dupe issue code, but adjust the `name` to be something like `CHECK(STORAGE)` instead of just `STORAGE`? That would also make the diagnostic shown in the "problems" view of VSCode a bit more helpful.

packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -156,15 +159,24 @@ export class Translator<T extends CompilerOptionsPP = CompilerOptionsPP> {
       }
 
       try {
-        translate?.(option, this.options);
+        const localDiagnostics: CompilerOptionIssue[] = [];
```
msujew, line 162: Suggestion: While you're already at it, can you remove the `CompilerOptionIssue` and replace it with the `Diagnostics` type? AFAIK `CompilerOptionIssue` does not track the `code` and other properties. The `TranslationError` could be removed as well, since we have stuff like `diagnosticFromCode` and other utility functions.

msujew (commented):
Looks quite good to me, I just have a few suggestions I would like to see as part of this refactoring :)

packages/language/src/preprocessor/compiler-options/translator.ts
ssmifi: Good point. I started on this, but realized that to do it completely it would also require some more refactoring on the older compiler options (adding codes and tests and potentially doing tests on the mainframe). So, I would prefer to do the refactoring in a follow-up PR.

msujew (approved):
LGTM

PR #430, opened 2025-10-09
Fix SQL builtins errors

### LINKED ISSUE #413: Resolve SQLCA, SQLDA build-in members
Currently built-in support for SQLCA and SQLDA is working fine but members are not resolved. 
All members should be resolved. 

e.g.1

Image

e.g.2
Image

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/413

This is rather hacky - a real solution would involve splitting the different builtins into individual files, but we currently don't have the tech to facilitate this. In the meantime, this works pretty well.

### DIFF
--- a/packages/language/src/validation/pli-validator.ts
+++ b/packages/language/src/validation/pli-validator.ts
@@ -377,19 +377,8 @@ export class PliValidator implements Validator {
     }
   }
 
-  private knownBuiltins = new Set([
-    "TRUE",
-    "FALSE",
-
-    "SQLCA",
-    "SQLDA",
-    "SQLDA2",
-    "SQLSIZE",
-    "SQLDAPTR",
-    "SQLTRIPLED",
-    "SQLDOUBLED",
-    "SQLSINGLED",
-  ]);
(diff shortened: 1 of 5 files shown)


packages/language/src/validation/pli-validator.ts
```diff
@@ -399,18 +391,29 @@ export class PliValidator implements Validator {
...
+    // This is a rather hacky way to determine what are "valid" builtins
+    // We should have a separate file for that
+    // TODO: Refactor this, probably when fixing the LSP features for builtins
+    if (this.knownBuiltinsOffset === 0) {
+      const file = this.compilationUnit.files.getDocument(uri);
+      if (!file) {
+        return;
+      }
+      const text = file.getText();
+      this.knownBuiltinsOffset = text.indexOf("/* Known Builtins */");
```
ssmifi: _Minor_: Should we at least use a constant for the marker?

ssmifi (approved):

ssmifi:
As you said. Works. ;-)

Issue #431, opened 2025-10-09
preprocessor-findings 9.10


asatklichov:
- [x] 1. Builtin` datetime` is not recognized, typo - DATEIME (https://github.com/zowe/zowe-pli-language-support/pull/430)

- [x] 2. ` ENTRY `support in preprocessor e.g. (https://github.com/zowe/zowe-pli-language-support/pull/440)
`%dcl x(2) fixed entry; `     same as in , #411  
- [x] 3. Validation is missing in %DECLARE statement, e.g.
```
%dcl A(4) FIXED INTERNAL RESCAN SCAN; /*must be one option, not both */
%dcl (b), (c), b(lbound(2):5) fixed fixed internal internal;  /*must be not be repeatable */
%dcl (b),(c), b(lbound(2):5) fixed,(m), (n),b(lbound(2):5) fixed fixed; /*must be not be repeatable */
```
- [x] 4.  has issues with LABEL definition, can not be closed with END; (https://github.com/zowe/zowe-pli-language-support/pull/427)
 ```
%DECLARE I FIXED BIN(31);
   
  %LOOP: DO I=1 TO 5;  
    PUT LIST ('ITER', I);
  %END LOOP; /*error, spec page 620 */
```

or another example

  ```
%DECLARE I FIXED BIN(31);
   
  %LOOP: DO I=1 TO 5;  
    PUT LIST ('ITER', I);
  %LOOP: END LOOP; /*error, spec page 620 */
```

- [x] 5. Wrong interpretation of SKIPPING (https://github.com/zowe/zowe-pli-language-support/pull/433)

<img width="499" height="209" alt="Image" src="https://github.com/user-attachments/assets/278979e9-f538-464c-8d4e-f4b332b6dc37" />

- [x] 6. Once 'NOT' compiler option used with '¬' and in code we use '!', it does not report error, or it shows  misleading error (https://github.com/zowe/zowe-pli-language-support/pull/435)

`%PROCESS OR('|'), NOT ('¬') ; /* NOT ('!');*/`
 

<img width="416" height="189" alt="Image" src="https://github.com/user-attachments/assets/c4d5a3ec-e90a-4def-84b3-5a57facde052" />

or misleading error 'Character '!' with decimal value 33 does not belong to the PL/I character set. It will be ignored'

<img width="413" height="185" alt="Image" src="https://github.com/user-attachments/assets/e25911ff-2e12-4f0a-b254-d93cd228a06c" />

[...]

PR #433, opened 2025-10-09
Fix comparison operation in macro interpreter

msujew:
Related to (5) in https://github.com/zowe/zowe-pli-language-support/issues/431.

Adds a bunch of tests that confirm that this now behaves as expected (with a few negative tests as well).


ssmifi (approved):
LGTM

Issue #432, opened 2025-10-09
Abbreviated 'stmt' is not supported


asatklichov:
Currently, below code is compiling with 'statement', but it should also support abbreviated form 'stmt' 

```
    %Statement_function: procedure( key1 ) stmt returns( fixed );
        dcl key1 fixed;
        return( key1 + key1 );
    %end;
```

montymxb:
Closed by #434

PR #434, opened 2025-10-09
Support STMT shorthand token

msujew:
Fixes https://github.com/zowe/zowe-pli-language-support/issues/432.

### DIFF
--- a/packages/language/src/parser/tokens.ts
+++ b/packages/language/src/parser/tokens.ts
@@ -423,7 +423,7 @@ export const PROCEDURE = registerKeyword({
   names: ["PROCEDURE", "PROC", "XPROCEDURE", "XPROC"],
 });
 export const STATEMENT = registerKeyword({
-  name: "STATEMENT",
+  names: ["STATEMENT", "STMT"],
 });
 export const CHARACTER = registerKeyword({
   names: ["CHARACTER", "CHAR"],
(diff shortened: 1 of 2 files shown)


packages/language/src/parser/tokens.ts
```diff
@@ -423,7 +423,7 @@ export const PROCEDURE = registerKeyword({
   names: ["PROCEDURE", "PROC", "XPROCEDURE", "XPROC"],
 });
 export const STATEMENT = registerKeyword({
-  name: "STATEMENT",
+  names: ["STATEMENT", "STMT"],
```
ssmifi: _Question_: This change made we wonder, why we use two fields to separate between array or not in the first place? Couldn't we just use `name(s): string | string[];` and check if it is an array or not?

msujew: I've updated the typing, see latest commit :)

ssmifi (approved):
LGTM
By the way, personally, I would prefer plural for "one-or-more" cases, but I figured you settled for the singular form for to be in line with Chevrotain? :)

msujew:
> By the way, personally, I would prefer plural for "one-or-more" cases, but I figured you settled for the singular form for to be in line with Chevrotain? :)

Not necessarily, but this form just felt more natural to me ¯\\\_(ツ)\_/¯

PR #436, opened 2025-10-13
Prepare 1.0.3 release

msujew:
* Updates the package.json to version `1.0.3`
* Adds a changelog to include all the (meaningful) changes for version 1.0.3 with links to the respective PRs.
* Updates our pnpm lock file to use the latest versions of all dependencies.

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -10,8 +10,8 @@
     "type": "git",
     "url": "https://github.com/zowe/zowe-pli-language-support"
   },
-  "version": "1.0.0",
-  "displayName": "PL/I Programming Language Support",
+  "version": "1.0.3",
+  "displayName": "PL/I Language Support",
   "engines": {
     "vscode": "^1.67.0"
   },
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/package.json
```diff
@@ -10,7 +10,7 @@
     "type": "git",
     "url": "https://github.com/zowe/zowe-pli-language-support"
   },
-  "version": "1.0.0",
+  "version": "1.0.2",
```
asatklichov: Version will be 1.0.3 (as you know we tried internal 2 releases because of display name)

msujew: The latest [public release](https://marketplace.visualstudio.com/items?itemName=broadcomMFD.pli-language-support) is still at `1.0.1`:

image


Isn't `1.0.2` more appropriate?

asatklichov (approved):

packages/vscode-extension/package.json
asatklichov: regarding display update  name we keep 1.0.3

PR #440, opened 2025-10-14
Support macro parse of ENTRY attribute

msujew:
Adds preliminary support to parse the `entry` attribute keyword on preprocessor declarations. Does not implement actual support in the interpreter for it.


Lotes (approved):
LGTM

PR #443, opened 2025-10-15
Resolve Libs Recursively

montymxb:
Closes the 2nd point of #394 by adding support for recursive libs resolution. When your process group config has a libs entry like so: `{libs: ["cpy"]}`, then we resolve all valid libs files contained in any sub directories too. For example, `cpy/folder2/mylib.pli` is resolvable as `%INCLUDE "mylib.pli"` without needing to add `cpy/folder2` explicitly to the config.

When a process group config is loaded or set via the plugin config proivder, we post-process each lib entry to explore & add all sub directories as valid lib entries. This changes from the user supplied `libs` entry to be something that is expanded with all reachable sub directories as well.

### DIFF
--- a/packages/language/src/workspace/plugin-configuration-provider.ts
+++ b/packages/language/src/workspace/plugin-configuration-provider.ts
@@ -331,7 +331,7 @@ export class PluginConfigurationProvider {
 
       if (processGrpConfig !== undefined) {
         try {
-          this.parseProcessGroupConfigs(processGrpConfig);
+          await this.parseProcessGroupConfigs(processGrpConfig);
           this.postProcessProgramConfigs();
           await this.postProcessProcessGroups();
           return;
@@ -470,11 +470,11 @@ export class PluginConfigurationProvider {
    * @param text Raw text content of .pliplugin/proc_grps.json to parse
    * @returns Whether parsing was successful
    */
-  public parseProcessGroupConfigs(text: string): boolean {
+  public async parseProcessGroupConfigs(text: string): Promise<boolean> {
     try {
       const serializedData: SerializedProcessGroup[] = JSON.parse(text).pgroups;
       const groupConfigs = serializedData.map(deserializeProcessGroup);
-      this.setProcessGroupConfigs(groupConfigs);
+      await this.setProcessGroupConfigs(groupConfigs);
       this.postProcessProgramConfigs();
       return true;
     } catch {
@@ -569,7 +569,7 @@ export class PluginConfigurationProvider {
   public getProcessGroupConfigFromLib(libUri: URI): ProcessGroup | undefined {
     const dirname = UriUtils.basename(UriUtils.dirname(libUri));
     for (const config of this.processGroupConfigs.values()) {
-      if (config.$computedLibs?.includes(dirname)) {
+      if (config.$computedLibs.includes(dirname)) {
         return config;
       }
     }
--- a/packages/language/src/workspace/plugin-configuration-provider.ts
+++ b/packages/language/src/workspace/plugin-configuration-provider.ts
@@ -308,6 +308,7 @@ export class PluginConfigurationProvider {
         try {
           this.parseProcessGroupConfigs(processGrpConfig);
           this.postProcessProgramConfigs();
+          await this.postProcessProcessGroups();
           return;
         } catch (e) {
           console.error("Failed to load process group config, skipping:", e);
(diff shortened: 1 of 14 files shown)


packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -72,6 +72,14 @@ export interface ProcessGroup {
   compilerOptions: string[];
   pliOptions: PliOptions;
   libs: string[];
+
+  /**
+   * Sub directories of libs, populated after reading the configs.
+   * A computed property that is not serialized like regular 'libs',
+   * but is also used to resolve includes.
+   */
+  _subLibs: string[];
```
msujew: Suggestion: We always seem to concat both the `libs` and `_subLibs` folders when accessing them (which makes sense). Why don't we merge them from the start - that removes a lot of redudancy in the accessing code. Also, can we name this to something like `$computedLibs`? `_` is usually the indicator for a private property.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -324,6 +342,39 @@ export class PluginConfigurationProvider {
...
+        if (lib) {
+          // read all files in this lib path
+          // add any contained directories to the libs list, as well as the toProcess list
+          const libUri = UriUtils.joinPath(URI.parse(this.workspacePath), lib);
+          const entries = await FileSystemProviderInstance.readDir(libUri);
+          if (entries.length) {
+            for (const dirEntry of entries) {
+              const fileName = dirEntry.name;
+              const fileType = dirEntry.type;
+              if (fileType === 2) {
```
msujew:
```suggestion
              if (fileType === DirEntryType.Directory) {
```

packages/language/test/fourslash/preprocessor/include/from-lib-path-recursive.ts
```diff
@@ -0,0 +1,38 @@
...
+////         "name": "default",
+////         "lsp-options": {
+////             "check-margins": true
+////         },
+////         "libs": [
+////             "cpy"
+////         ]
+////         }
+////     ]
+//// }
```
msujew, line 28: Suggestion: Can we instead instruct the fourslash framework to reparse the config after adding all the files to the virtual file system? This seems like a workaround for an issue that should be solved within the framework.

msujew (approved):
This looks quite good, but I have 2 improvements I would like to see in there.

packages/language/test/fourslash/preprocessor/include/from-lib-path-recursive.ts
montymxb: Yeah that's a good idea. This pattern is something we've been using for some time due to needing to modify specific details, but we'll end up tripping on the same issue whenever testing sub-libs if we can't reparse on demand. I'll put that in there.

packages/language/src/workspace/plugin-configuration-provider.ts
montymxb: Ah nice thanks for catching that.

packages/language/src/workspace/plugin-configuration-provider.ts
montymxb: Can do with the renaming.

If we merge them we also need to divide them up on serialization, otherwise we'll end up changing the libs entry on disk. It won't be incompatible or anything, but it's probably unwanted. I can filter out the computed sub libs from the core libs on serialization, and then it's just `libs` being used everywhere else as it was before.

msujew: Why would we divide them up on serialization? `libs` is kept as it is read from the original config, whereas `$computedLibs` contains all of them. We just need to serialize `libs`.

montymxb: Because you just said to merge them from the start, so libs would be merged with $computedLibs. I presume I'm misunderstanding you and you mean the start of the include checking logic or something else? I'm under the impression you mean the start of the extension, if that helps clarify how I'm looking at it.

montymxb: Ahhh just looked at this again today, I see what you mean.`$computedLibs` is the new total switching away from `libs`, leaving it unchanged. That's makes more sense, was wondering what you meant there. I'll add that in too.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -497,12 +564,12 @@ export class PluginConfigurationProvider {
...
    * process group associated with a library file. It is rather fuzzy and might not be 100% accurate.
    * @param libUri URI of the including file (likely a library file)
-   * @returns Associated process group config, or undefined if not found
+   * @returns First process group config that includes a matching lib path, or undefined if not found
    */
   public getProcessGroupConfigFromLib(libUri: URI): ProcessGroup | undefined {
     const dirname = UriUtils.basename(UriUtils.dirname(libUri));
     for (const config of this.processGroupConfigs.values()) {
-      if (config.libs?.includes(dirname)) {
+      if (config.$computedLibs?.includes(dirname)) {
```
msujew: Minor:

```suggestion
      if (config.$computedLibs.includes(dirname)) {
```

packages/language/test/fourslash/preprocessor/include/from-lib-path-recursive.ts
montymxb: Following up here, I'm starting to think this might make a good follow-up PR instead. The change requires us to be able to await the async reload of the config provider, which we can't do just yet. This requires a bit more work, and I think that might be good to split off then. Thoughts on this?

msujew: Sure, fine with me 👍

PR #444, opened 2025-10-15
Type system: testing utils and structure types

Lotes:
## About the code

* added testing utils for checking the type of an underlying symbol
* structure types, but no unions, (yet)
* needed to split out `precision` from `scale mode` in order to model both independently from each other
* concept of type attribute implications (see the table in `descriptions.ts`), e.g.
  * if `BINARY` is given, this implies that the data type is an `arithmetic`
  * if `PRECISION` is given with two numbers passed, this implies that the scale mode is `FIXED`
* removed and also improved parts of the type inferer
  * the code wasn't used right now (dead code)
  * type resolution for declared statement and sub elements was optimized

## About the tests

* unfortunately, some tests are broken - they throw the wrong diagnostics (IBM2462 instead of IBM2424), affected tests contain a `TODO` - I think this can be also fixed later. IBM2462 informs the user about a conflict in the attributes, IBM2424 says more about the conflict (scaling factor not allowed for FLOAT)


packages/language/src/typesystem/descriptions.ts
```diff
@@ -30,23 +30,28 @@ export enum DataType {
...
+  Format: DataType.Format as const,
+  Label: DataType.Label as const,
+  Locator: DataType.Locator as const,
+  Ordinal: DataType.Ordinal as const,
+  Picture: DataType.Picture as const,
+  String: DataType.String as const,
+  Structure: DataType.Structure as const,
+  Task: DataType.Task as const,
+  Unknown: DataType.Unknown as const,
+};
```
msujew: Question: Why have another object for this? The `DataType` enum already is an object. Same with the other objects that are just enums.

packages/language/test/test-builder.ts
```diff
@@ -932,6 +938,61 @@ export class TestBuilder {
     });
   }
 
+  expectTypeAt(label: string, expectedType: TypeExpectation): void {
+    const ranges = this.getLabelRanges(label);
+    for (const [start] of ranges) {
+      const node = this.syntaxNodesByOffset.get(start);
```
msujew: Suggestion: I don't necessarily think we need the `syntaxNodesByOffset` map or the `collectSyntaxNodesByOffset` function, since we don't really care about the range in the `expectTypeAt` testing method (you only use `start` anyway). We really only need an index:

```
DCL <|1>VAR_NAME FIXED(10);
```

With the index in place, we can perform a binary search for the token and retrieve its `element`. That fulfills the exact same purpose, but without the overhead of collecting and storing all offsets/nodes in a map.

packages/language/test/test-builder.ts
```diff
@@ -932,6 +938,61 @@ export class TestBuilder {
...
+  ) {
+    if (expectedType.type === DataType.Structure) {
+      if (actualType.type !== DataType.Structure) {
+        throw new Error(
+          `Expected type to be a ${TypeDescriptions.Names[DataType.Structure]}, but got ${TypeDescriptions.Names[actualType.type]}`,
+        );
+      }
+      for (const [name, expectedMemberType] of Object.entries(
+        expectedType.members ?? {},
+      )) {
```
msujew, line 968: Suggestion: This doesn't catch any additional properties that might be inside of `actualType`, but that aren't expected within `expectedType`. I.e. this will result in a false positive when we have something like `expectedType: {}`, `actualType: { name: String }`.

msujew (commented):
This looks pretty good. I'm happy with the functional changes, but have a few improvements for the changes for the type-testing.

packages/language/src/typesystem/descriptions.ts
Lotes: I had a knowledge gap. Thank you!

Lotes:
My current two concerns are the type caching and cyclic types.

Caching: I think it can be better. `DeclareStatements` can have multiple variables, but they are also the entry point for gathering information (like for structures which span over multiple `DeclaredItems`). One solution could be multiple entry points in the type inferer.

Cyclic types: I saw that they are possible with `define structure`. Need to put an eye on it. My current approach will not "endless loop", because DEFINE, HANDLE and TYPE are not supported right now.

Both things can be done in a follow up PR. Also because I have no much time left before my 2 weeks of vacation. I do not want to block the development with my absence.

msujew (approved):
Looks good to me, thanks 👍

Issue #445, opened 2025-10-15
%include directive related findings


asatklichov:
- [ ]  As per spec, %INCLUDE directive can not be empty, but currently it does not do a error reporting. 

<img width="387" height="189" alt="Image" src="https://github.com/user-attachments/assets/f0c7a883-9815-456b-9285-49933b56187f" />

Spec.

<img width="575" height="176" alt="Image" src="https://github.com/user-attachments/assets/9a9f63f9-cd17-4af2-bfc8-f539a6787c65" />

- [ ]  Once you delete, one of or all items under "libs" entry, program can not resolve copybooks. You have to re-load VSCode. 

```
  "libs": [
        "cpy",
        "inc" 
      ],
```

asatklichov:
@msujew both fixed, thanks

PR #446, opened 2025-10-15
Show error on missing token after include

msujew:
Fixes part (1) of https://github.com/zowe/zowe-pli-language-support/issues/445.

Adjusts the test-builder to improve the parsing of test files when lacking a range marker name.

### DIFF
--- a/packages/language/src/parser/parser-state.ts
+++ b/packages/language/src/parser/parser-state.ts
@@ -20,6 +20,7 @@ import {
   Severity,
 } from "../language-server/types";
 import { PLICodes } from "../validation/messages";
+import { SimplePLICode } from "../validation/messages/pli-codes";
 
 export function preprocessorParserState(tokens: t.Token[]): ParserState {
   return new ParserState(tokens, ParserStateMode.Preprocessor);
@@ -108,17 +109,21 @@ export class ParserState {
    * until a successful consume() happens or the recover() method is called.
    */
   error(
-    message?: string,
+    message?: string | SimplePLICode,
     token = this.token || this.last,
     severity = Severity.S,
   ) {
     if (!this.inError) {
-      const msg =
-        message ??
-        (this.eof
-          ? "Unexpected end of file."
-          : `Unexpected token '${generateTokenErrorName(token)}'.`);
-      this.diagnostics.push(diagnostic(severity, msg, token));
+      if (!message || typeof message === "string") {
+        const msg =
+          message ??
+          (this.eof
+            ? "Unexpected end of file."
+            : `Unexpected token '${generateTokenErrorName(token)}'.`);
+        this.diagnostics.push(diagnostic(severity, msg, token));
+      } else {
+        this.diagnostics.push(diagnosticFromCode(message, token));
+      }
       this.inError = true;
     }
   }
@@ -177,12 +182,17 @@ export class ParserState {
     element: SyntaxNode | undefined,
     kind: CstNodeKind | undefined,
     tokenType: TokenType,
+    code?: SimplePLICode,
   ): t.Token | null {
     const token = this.token;
     if (!token || !this.canConsume(tokenType)) {
       if (!this.inError) {
-        const message = `Expected token "${tokenType.name}", but received ${generateTokenErrorName(token)} instead.`;
-        this.error(message);
+        if (code) {
+          this.error(code);
+        } else {
+          const message = `Expected token "${tokenType.name}", but received ${generateTokenErrorName(token)} instead.`;
+          this.error(message);
+        }
       }
       return null;
     } else {
--- a/packages/language/src/parser/preprocessor-parser.ts
+++ b/packages/language/src/parser/preprocessor-parser.ts
@@ -582,45 +582,58 @@ function labelReference(state: ParserState): ast.LabelReference {
 
 function includeStatement(state: ParserState): ast.IncludeDirective {
   const directive = ast.createIncludeDirective();
-  const token = state.consume(
+  const includeToken = state.consume(
     directive,
     CstNodeKind.IncludeDirective_INCLUDE,
     t.INCLUDE,
   );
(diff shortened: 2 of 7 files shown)


packages/language/test/utils.ts
```diff
@@ -313,10 +313,11 @@ export function replaceNamedIndices(text: string): {
   const ranges: Record<string, Array<[number, number]>> = {};
   const rangeStack: {
     index: number;
-    label: string;
+    label?: string;
   }[] = [];
 
-  const regex = /<\|(?<indexMarker>\w+)>|<\|(?<rangeStartMarker>\w+):|\|>/;
+  const regex =
+    /<\|(?<indexMarker>\w+)>|(?<rangeStartMarker><\|)((?<rangeLabel>\w+):)?|\|>/;
```
montymxb, line 325: Could you clarify why was the testing logic adjusted here? I can see the adjustments above but I'm not sure why this needed.

montymxb (commented):
Looks pretty good, tested separately as well. One question and I'll defer to @Lotes for final review.

packages/language/test/utils.ts
msujew: On `development`, we *always* need a range label. I.e. `<|label:DCL|>`. Using this kind of syntax would associate the `"label"` label with the range that `DCL` takes up in the text file. My change adjusts the test-builder to automatically generate labels if none is provided. Meaning that the range marker `<|DCL|>` inserts `DCL` into the text and auto generates the range label `"DCL"`. It just makes writing tests for diagnostics that only affect a single token a bit easier.

packages/language/src/parser/parser-state.ts
```diff
@@ -177,12 +182,17 @@ export class ParserState {
     element: SyntaxNode | undefined,
     kind: CstNodeKind | undefined,
     tokenType: TokenType,
+    code?: SimplePLICode,
```
Lotes, line 185: Feels a bit hacky, because we suddenly care about error codes. But I have no good idea for an alternative (only promises came up, like you do an `await consume(...).catch(() => diagnosticFromCode(...))`).

Apart from that, we are excluding ParametricPLICodes (not sure if we need them here really). An alternative could be a lambda as a factory for a diagnostic.

Lotes: In synchronous code you would create a try-catch around it, which make the code bigger/more complex.

packages/language/src/parser/preprocessor-parser.ts
```diff
@@ -582,45 +582,58 @@ function labelReference(state: ParserState): ast.LabelReference {
...
     directive.items.push(item);
     // Optional comma
     state.tryConsume(directive, CstNodeKind.IncludeDirective_Comma, t.Comma);
   }
-  state.consume(directive, CstNodeKind.IncludeDirective_Semicolon, t.Semicolon);
+  const semicolon = state.consume(
+    directive,
+    CstNodeKind.IncludeDirective_Semicolon,
+    t.Semicolon,
+    PLICodes.Severe.IBM1618I,
```
Lotes, line 630: Do we have a test for `IBM1618I`?

packages/language/test/fourslash/parser/pp/procedure-with-percent-in-body.ts
```diff
@@ -12,11 +12,11 @@
 /// <reference path="../../framework.ts" />
 
 //// %TEST: PROC;
-////   <|1:%|>DCL I FIXED;
+////   <|%|>DCL I FIXED;
```
Lotes, line 15: Nice usage of the new marker feature.

packages/language/test/utils.ts
Lotes: I like this feature. What do you think about splitting the `RegExp` into two or three(?) new `RegExps`?
The intention is clear, but the alternatives are hard to parse for me. I think the code would benefit readability from that. Maybe introduce some helper functions to make the loop body shorter.

msujew: That would make it rather difficult to work with. The benefit of having all of these in a single regexp is that we can just get a list of ordered occurrences of each match in `groups`. If we were to split them, we would have to merge and sort the groups outselves.

packages/language/src/parser/parser-state.ts
msujew: My latest refactoring was explicitly done to remove all `try-catch` code in the parser since it made the control-flow in case of parsing errors incredibly complicated. The issue with the `code` is that we definitely want to supply it with the `consume` function, because we would otherwise generate two diagnostics for the failed consume. The `code` is just there to improve the error message in case this is a "known" issue.

> Apart from that, we are excluding ParametricPLICodes (not sure if we need them here really).

AFAIK, none of the parser errors are parametric.

packages/language/src/parser/preprocessor-parser.ts
msujew: I've added a test.

packages/language/test/utils.ts
Lotes: OK, another alternative could be a documented, multiline RegExp, but I think that is only a [feature in C#](https://learn.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regexoptions?view=net-9.0#system-text-regularexpressions-regexoptions-ignorepatternwhitespace).

Somehow like this, if I remember correctly...
```csharp
new Regex("""
a #reads 'a'
b #followed by 'b'
""", RegexOptions.IgnorePatternWhitespace)
```

Or add some short comment (at the if conditions maybe?), what cases there are.
I see there are 3 labels in the Regexp, but it is a bit difficult to see how they start and end, what alternative ends where and so on.

msujew: I've added some documentation to explain a bit better what each part stands for.

Lotes (approved):
LGTM, thanks

PR #451, opened 2025-10-17
[Feature] Implement diagnostic and quick fix for missing configuration

Wagner-Laranjeiras:
### [Feature] Implement diagnostic and quick fix for missing configuration and [Bugfix] Fix non existent lib breaking all INCLUDES.

Issue: #463 #462

**Summary:**
[Feature] This PR introduces a new Diagnostic (LSPIR001) and Quick Fix action (pli.applyQuickFixCreateConfig) for missing files. When the config files are not present, the diagnostic will be thrown at the file in the %INCLUDE. The Quick Fix will then offer creating the configuration, having the current file as the program entry point.
[Bugfix] Fix INCLUDES resolution breaking because of non existent directory added to libs. 

**Changes**
- apply-quick-fixes.ts file now has a generic function which calls more specific functions based on the current diagnostic.
- New commands.ts file, which contains functions called at the connection-handler.
- New constants.ts file, which contains namespaces for the Commands codes and PluginConfiguration.
- New check for throwing diagnostic at the instruction-interpreter.ts file. Now, before trying to resolve an INCLUDE, we check if the **full** configuration is present and throw the new Diagnostic if it's not.
- Catch error scenario for when trying to read directory and return an empty array if it's the case.
- Refactor searchPath value handling at search function.

### DIFF
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -18,10 +18,14 @@ import {
   Diagnostic,
 } from "vscode-languageserver-types";
 import {
-  PluginConfigurationProvider,
   PluginConfigurationProviderInstance,
   serializeProcessGroup,
 } from "../../workspace/plugin-configuration-provider";
+
+import { Commands, PluginConfiguration } from "../constants";
+import { PLICodes } from "../../validation/pli-codes";
+import { LspCodes } from "../../validation/lsp-codes";
+import { fullCode } from "../types";
 
 export async function quickFixResolveInclude(
   diagnostic: Diagnostic,
@@ -73,17 +77,38 @@ export async function quickFixResolveInclude(
 
   const procGrpsFileUri = UriUtils.joinPath(
     workspaceFolderUri,
-    PluginConfigurationProvider.PROCESS_GROUP_CONFIG_FILE,
+    PluginConfiguration.PROCESS_GROUP_FILE_PATH,
   );
 
   const action: CodeAction = {
-    title: `Add '${parentFolder}' to INCLUDE libs`,
+    title: `Add '${parentFolder}' to INCLUDE libs.`,
     kind: CodeActionKind.QuickFix,
     diagnostics: [diagnostic],
     command: {
       title: "Apply INCLUDE fix",
-      command: "pli.applyIncludeFix",
+      command: Commands.RESOLVE_INCLUDE,
       arguments: [procGrpsFileUri.toString(), newContent],
+    },
+  };
+
+  return action;
+}
+
+export async function quickFixCreateConfig(
+  diagnostic: Diagnostic,
+): Promise<CodeAction | undefined> {
+  const entryUri = UriUtils.basename(URI.parse(diagnostic.data.entryUri));
+  if (!diagnostic.data.entryUri || !entryUri) {
+    return undefined;
+  }
+  const action: CodeAction = {
+    title: `Create a plugin configuration folder for this file.`,
+    kind: CodeActionKind.QuickFix,
+    diagnostics: [diagnostic],
+    command: {
+      title: "Create configuration folder",
+      command: Commands.CREATE_CONFIG,
+      arguments: [entryUri],
     },
   };
 
@@ -95,12 +120,25 @@ export async function applyQuickFixes(
 ): Promise<CodeAction[] | undefined> {
   const actions: CodeAction[] = [];
   // PLI CODES LIST
-  const CODE_UNRESOLVED_INCLUDE = "IBM3841IS"; // The INCLUDE file could not be found, or if found, it could not be opened.
+  const CODE_UNRESOLVED_INCLUDE = fullCode(PLICodes.Severe.IBM3841I); // The INCLUDE file could not be found, or if found, it could not be opened.
+  const CODE_MISSING_CONFIG = fullCode(
+    LspCodes.IncludeResolution.MissingConfiguration,
+  ); // "Could not resolve include directive. Plugin configuration is missing"
 
   for (const diagnostic of diagnostics) {
-    if (diagnostic.code === CODE_UNRESOLVED_INCLUDE) {
-      const action = await quickFixResolveInclude(diagnostic);
-      if (action) actions.push(action);
+    if (!diagnostic.code) {
+      return;
+    }
+    let action: CodeAction | undefined;
+    switch (diagnostic.code) {
+      case CODE_UNRESOLVED_INCLUDE:
+        action = await quickFixResolveInclude(diagnostic);
+        if (action) actions.push(action);
+        break;
+      case CODE_MISSING_CONFIG:
+        action = await quickFixCreateConfig(diagnostic);
+        if (action) actions.push(action);
+        break;
     }
   }
 
--- a/packages/language/src/language-server/commands.ts
+++ b/packages/language/src/language-server/commands.ts
@@ -1,0 +1,75 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { URI } from "vscode-uri";
+import { FileSystemProviderInstance } from "../workspace/file-system-provider";
+import { Mutex } from "../workspace/mutex";
+import { PluginConfigurationProviderInstance } from "../workspace/plugin-configuration-provider";
+import { ExecuteCommandParams } from "vscode-languageserver";
+import { UriUtils } from "../utils/uri";
+import { PluginConfiguration } from "./constants";
+
+export async function commandResolveInclude(params: ExecuteCommandParams) {
+  return Mutex.run(async () => {
+    const [uri, content] = params.arguments as string[];
+    try {
+      await FileSystemProviderInstance.writeFile(URI.parse(uri), content);
+    } catch (err) {
+      console.error("Failed to write proc_grps.json:", err);
+    }
+  });
+}
+
+export async function commandCreateConfig(params: ExecuteCommandParams) {
+  return Mutex.run(async () => {
+    const workspaceFolderUri = URI.parse(
+      PluginConfigurationProviderInstance.getWorkspacePath(),
+    );
+    if (!workspaceFolderUri || !params.arguments) {
+      return;
+    }
+    try {
+      const entryUri = params.arguments[0];
+      await FileSystemProviderInstance.writeFile(
+        UriUtils.joinPath(
+          workspaceFolderUri,
+          PluginConfiguration.PROGRAM_FILE_PATH,
+        ),
+        JSON.stringify(
+          {
+            ...PluginConfiguration.DEFAULT_PROGRAM_FILE_CONTENT,
+            pgms: [
+              {
+                ...PluginConfiguration.DEFAULT_PROGRAM_FILE_CONTENT.pgms[0],
+                program: entryUri,
+              },
+            ],
+          },
+          null,
+          2,
+        ),
+      );
+      await FileSystemProviderInstance.writeFile(
+        UriUtils.joinPath(
+          workspaceFolderUri,
+          PluginConfiguration.PROCESS_GROUP_FILE_PATH,
+        ),
+        JSON.stringify(
+          PluginConfiguration.DEFAULT_PROCESS_GROUP_FILE_CONTENT,
+          null,
+          2,
+        ),
+      );
+    } catch (err) {
+      console.error("Failed to create configuration: ", err);
+    }
+  });
+}
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -34,6 +34,7 @@ import { Diagnostic, diagnosticFromCode } from "../language-server/types";
 import { PreprocessorTokens } from "./pli-preprocessor-tokens";
 import { tokenMatcher } from "chevrotain";
 import { PLICodes } from "../validation/pli-codes";
+import { LspCodes } from "../validation/lsp-codes";
 
 interface Variable {
   name: string;
@@ -2286,6 +2287,7 @@ async function resolveIncludeFileUri(
         const match = await FileSystemProviderInstance.search({
           path: libFileUri,
           extensions: pgroup.includeExtensions,
+          global: true,
         });
         if (match) {
           return match;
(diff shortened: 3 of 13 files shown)


packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -82,25 +82,58 @@ export async function quickFixResolveInclude(
...
 
+export async function quickFixCreateConfig(
+  diagnostic: Diagnostic,
+): Promise<CodeAction | undefined> {
+  const entryUri = UriUtils.basename(URI.parse(diagnostic.data.entryUri));
+  if (!entryUri) {
+    return;
+  }
+  const action: CodeAction = {
+    title: `Create a '.pliplugin' folder in the project root using the current file as the entry point in 'pgm_conf.json'`,
```
msujew: Minor: IMO too long of a title for this code action:
```suggestion
    title: `Create a plugin configuration folder for this file.`,
```

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -82,25 +82,58 @@ export async function quickFixResolveInclude(
...
+  return action;
+}
+
 export async function applyQuickFixes(
   diagnostics: Diagnostic[],
 ): Promise<CodeAction[] | undefined> {
   const actions: CodeAction[] = [];
   // PLI CODES LIST
   const CODE_UNRESOLVED_INCLUDE = "IBM3841IS"; // The INCLUDE file could not be found, or if found, it could not be opened.
+  const CODE_MISSING_CONFIG = "_TB002E";
```
msujew: Suggestion: Please don't refer to the same constants multiple times as strings.
```suggestion
  const CODE_MISSING_CONFIG = Internal.DiagnosticMissingConfiguration.code;
```

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -47,6 +47,7 @@ export interface FileSystemProvider {
    * Returns a singular URI if found, otherwise undefined.
    */
   search(options: SearchOptions): Promise<URI | undefined>;
+  mkdir(uri: URI): Promise<void>;
```
msujew: Suggestion: I don't think there is a specific use case to generate a directory - instead, we should generate the directory on demand when calling `writeFile`.

packages/language/src/language-server/commands.ts
```diff
@@ -0,0 +1,91 @@
...
+        // if I'm on the right direction
+        const configFolder = workspaceFolderUri + "/.pliplugin";
+        if (!configFolder || !params.arguments) return;
+        const entryUri = params.arguments[0];
+
+        await FileSystemProviderInstance.mkdir(URI.parse(configFolder));
+        await FileSystemProviderInstance.writeFile(
+          URI.parse(
+            `${workspaceFolderUri}/${PluginConfigurationProvider.PROGRAM_CONFIG_FILE}`,
+          ),
```
msujew: Minor: We have a `UriUtils.joinPath` method exactly for this use case.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -82,25 +82,58 @@ export async function quickFixResolveInclude(
     diagnostics: [diagnostic],
     command: {
       title: "Apply INCLUDE fix",
-      command: "pli.applyIncludeFix",
+      command: "pli.applyQuickFixResolveInclude",
```
msujew: Suggestion: Please move this into a constant that is accessible to both this function AND the language server. Ideally, you can create a `namespace Commands` that contains these.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1882,28 +1884,60 @@ async function runInclude(
...
+
+    const hasProgConfig =
+      await FileSystemProviderInstance.fileExists(progConfigUri);
+    const hasProcsConfig =
+      await FileSystemProviderInstance.fileExists(procsConfigUri);
+
+    if (!hasProcsConfig && !hasProgConfig) return false;
+
+    return true;
+  }
```
msujew: Suggestion: The function has no dependency on any of the local variables, it can be moved out of the function.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -175,7 +175,28 @@ export class PluginConfigurationProvider {
...
+        "compiler-options": [],
+        libs: ["cpy", "inc"],
+        "include-extensions": [".pli", ".pl1", ".inc"],
+        "implicit-builtins": ["SUBSTR"],
+        "lsp-options": {
+          "check-margins": true,
+        },
+      },
+    ],
+  };
```
msujew: Suggestion: Export these from constants (same for the variables above actually).

packages/vscode-extension/src/extension/main.ts
```diff
@@ -22,6 +22,7 @@ import { Settings } from "./settings";
 import { registerCustomDecorators } from "./decorators";
 import { WorkspaceDidChangePlipluginConfigNotification } from "pli-language";
 import TelemetryReporter from "@vscode/extension-telemetry";
+import { PluginConfigurationProvider } from "../../../language/src/workspace/plugin-configuration-provider";
```
msujew: Suggestion: Do not import code from another package using a relative import. This can break in unexpected ways. Export the constants that you need from the `pli-language` package.

msujew (commented):
I've noticed that our `.pluplugin` directory handling is a bit inaccurate in case the user opens multiple workspace folders. See https://github.com/zowe/zowe-pli-language-support/issues/454 as a potential follow up on this.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Wagner-Laranjeiras: Yes, that also crossed my mind. Thanks!

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Wagner-Laranjeiras: Good catch, thanks

packages/language/src/language-server/commands.ts
Wagner-Laranjeiras: Perfect, gonna use it :)

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Wagner-Laranjeiras: Got it!

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: On it ;)

packages/language/src/workspace/file-system-provider.ts
Wagner-Laranjeiras: Hey @msujew , could we discuss this on Monday? I have some questions regarding this implementation. :)

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2017,25 +2019,36 @@ async function runInclude(
...
-      item.fileName,
-    );
-    if (item.fileName)
-      diagnostic.data = {
-        unresolvedFile: item.fileName,
-        entryUri: context.entryUri.toString(),
-      };
+    let diagnostic: Diagnostic;
+
+    if (!(await hasFullConfig())) {
```
msujew: Suggestion: I don't think we necessarily need to wait here for the file system. Assuming the current unit has a config available to it, it will be accessible via the `processGroup`/`programConfig` field. Accessing the file system might considerably slow down the interpreter in the error case.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2017,25 +2019,36 @@ async function runInclude(
...
+      diagnostic = diagnosticFromCode(
+        InternalCodes.DiagnosticMissingConfiguration,
+        item.token,
+      );
+      diagnostic.data = { entryUri: context.entryUri.toString() };
```
msujew: Suggestion: This shouldn't be implemented as an internal code. Internal codes are internal, because they shouldn't be displayed to the user. Generating this diagnostic however will show the internal code to the user. Reuse the `IBM3841I` code instead and add additional data into the `.data` field to differentiate between the quick fixes that are applied to these diagnostics.

packages/language/src/validation/internal-codes.ts
```diff
@@ -20,6 +20,12 @@ export const InternalCodes = {
       `Expected diagnostic at label "${label}" to be in file "${file}" but received: ${uri}`,
     fullCode: "_TB0001E",
   } as ParametricPLICode,
+  DiagnosticMissingConfiguration: {
+    code: "_TB002",
+    severity: Severity.E,
+    message: "Configuration is missing.",
+    fullCode: "_TB002E",
+  } as SimplePLICode,
```
msujew: Please remove, see reason above.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -365,15 +361,20 @@ export class PluginConfigurationProvider {
           // add any contained directories to the libs list, as well as the toProcess list
           const libUri = UriUtils.joinPath(URI.parse(this.workspacePath), lib);
           const entries = await FileSystemProviderInstance.readDir(libUri);
+          console.log(entries);
```
msujew: Please remove before undrafting this PR.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -365,15 +361,20 @@ export class PluginConfigurationProvider {
...
-                // also add to the full libs list
-                computedLibs.add(`${lib}/${fileName}`);
+                try {
+                  // directory to add for handling
+                  libsToProcess.push(`${lib}/${fileName}`);
+                  // also add to the full libs list
+                  computedLibs.add(`${lib}/${fileName}`);
+                } catch (err) {
+                  console.error(`Failed to add ${lib}/${fileName}: `, err);
+                }
```
msujew: Question: How can an array push/a set add operation fail?

packages/vscode-extension/src/language/main.ts
```diff
@@ -34,15 +35,22 @@ class NodeFileSystemProvider implements FileSystemProvider {
...
+          name: dirent.name,
+          type: dirent.isDirectory()
+            ? DirEntryType.Directory
+            : DirEntryType.File,
+        };
+      });
+    } catch (err) {
+      console.error(`Error reading directory: ${uri.fsPath}`, err);
+      return [];
+    }
```
msujew: Suggestion: IMO the call site should handle any file system failure. Only the caller knows what to do with a failure:

1. Do nothing (which would be essentially the same as returning `[]`).
2. Log the error and continue (perhaps displaying it to the user)
3. Bubble up the error

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: Ah ok. I understood that Internal codes were custom ones we had generate outside of the PLI-Codes.

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: I was testing if somehow that could be the problem and ended up letting this there because I wasn't sure it couldn't fail. Gonna remove it though.

packages/language/src/language-server/lsp-codes.ts
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { Severity } from "../language-server/types";
+import { SimplePLICode } from "../validation/pli-codes";
+
+export const lspCodes = {
```
msujew: Minor: Exported constants are usually either CamelCase or SNAKE_CASE.
```suggestion
export const LspCodes = {
```

packages/language/src/language-server/lsp-codes.ts
```diff
@@ -0,0 +1,24 @@
...
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { Severity } from "../language-server/types";
+import { SimplePLICode } from "../validation/pli-codes";
```
msujew: Suggestion: Please put this file in the `validation` part of the package.

Wagner-Laranjeiras:
No need to review it currently ;) Idea is to clean up more the tests (unnecessary comments and one more mock that I'm trying to get rid of)

packages/language/test/validating.test.ts
```diff
@@ -369,10 +370,19 @@ describe("Validating", () => {
...
-        severity: Severity.S,
-      });
+      assertDiagnostic(
+        doc,
+        // TODO: Implement more tests and differentiate when the config is present or not.
+        // This is going to be done for both the current and previous quick fix.
+        // {
+        // message: "The INCLUDE file nonexistent.pli could not be opened.",
+        // severity: Severity.S,
+        // },
```
Wagner-Laranjeiras: Just catch this that I forgot. Gonna work on it right now.

packages/language/test/validating.test.ts
```diff
@@ -373,6 +390,18 @@ describe("Validating", () => {
         message: "The INCLUDE file nonexistent.pli could not be opened.",
         severity: Severity.S,
       });
+      setPluginConfigurationProvider(undefined);
```
Wagner-Laranjeiras, line 392: I'm placing this here instead of in a afterEach because I don't want to risk messing up with other tests logic.

packages/language/test/lsp/apply-quick-fixes.test.ts
```diff
@@ -8,132 +8,216 @@
...
+//
+// ----------------------------------------------------------
+// quickFixResolveInclude tests
+// ----------------------------------------------------------
 describe("quickFixResolveInclude", () => {
-  beforeEach(() => {
-    vi.resetAllMocks();
+  test("returns undefined when no unresolved file in diagnostic", async () => {
+    const diagnostic = {
+      data: { entryUri: "/workspace/main.pli" },
```
msujew: Suggestion: The `entryUri` and `unresolvedFile` variables are supposed to be URI strings (like `file:///workspace/main.pli`). However, in your tests, they seem to be simple file paths. This does not reflect how they are used in the language server.

msujew (approved):
Thanks, I think this looks good 👍 Just a small-ish testing issue, see below.

packages/language/test/lsp/apply-quick-fixes.test.ts
Wagner-Laranjeiras: Thanks Mark! I'm focused on the caching issue currently, but I'll apply your review, solve the conflicts and then merge on Monday, first thing in the morning.

PR #453, opened 2025-10-21
Implement preprocessor for `SQL TYPE IS`

msujew:
Resolves the last point of #392 and follow up on #408.

Implements the `SQL TYPE IS` attribute as a preprocessor step that generates additional properties into the current procedure. Won't generate anything if not inside of procedure.

Also fixes a bug related to the caching of type system diagnostics (see changes in `lifecycle.ts`).


packages/language/src/parser/parser-entry.ts
```diff
@@ -30,31 +31,47 @@ export function preprocessorParse(
...
-      const stmt = ast.createStatement();
-      stmt.value = includeAlt;
-      recursivelySetContainer(stmt);
-      statements.push(stmt);
-    } else {
-      // TODO: Handle other preprocessors (SQL, CICS) here!
-      // state.canConsume(EXEC_SQL/DFHRESP/EXEC_CICS/...)
-
+  let token = state.token;
+  while (token) {
```
ssmifi, line 37: _Question_: Should we assert somehow that the token is actually advancing just in case there is a bug in the parser?

packages/language/src/parser/parser-entry.ts
```diff
@@ -30,31 +31,47 @@ export function preprocessorParse(
...
+          isTokenStatement = false;
+          const sqlAttrStmt = sqlAttributeStatement(state);
+          const sqlAttrStatement = ast.createStatement();
+          sqlAttrStatement.value = sqlAttrStmt;
+          recursivelySetContainer(sqlAttrStatement);
+          statements.push(sqlAttrStatement);
+        }
+        break;
+    }
+    if (isTokenStatement) {
```
ssmifi, line 70: _Minor Suggestion_: Using the default case of the switch as `isTokenStatement` would lose the boolean flag.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -484,6 +494,141 @@ function runInstructionSync(
...
+  }
+  if (body.kind === ast.SyntaxKind.SqlAttributeLobLocator) {
+    context.tokens.push(...lex(LOB_LOCATOR_TYPE));
+  } else if (body.kind === ast.SyntaxKind.SqlAttributeRowId) {
+    context.tokens.push(...lex(ROWID_TYPE));
+  } else {
+    // The LOB_FILE and LOB attributes require additional declarations to be inserted
+    // They are always inserted after the semicolon of the procedure statement
+    const procSemicolonIndex = findProcSemicolon(context);
+    if (procSemicolonIndex !== undefined) {
```
ssmifi: _Minor_: Could reduce nesting depth by exiting prematurely here.

ssmifi (approved):
LGTM, just minor comments.

packages/language/src/parser/parser-entry.ts
msujew: Not possible, since tokens like `SQL` don't directly trigger the parsing of the respective statements. They need an inner `if`, and in all other cases the token statement should be parsed.

PR #455, opened 2025-10-23
Support DD Style Includes

montymxb:
Closes the last of #394 by:

- supporting `%INCLUDE ddname(member);` includes, that resolve within libs
- adding parser support for segmented ddnames in includes, ex. `%include a.b.c(member)`
- `%include member;` includes that resolve within the first matching `ddname(member)` entry that is reachable from libs
- adding ddname style lib entries, such as `cpy/a.b.c`, where one or more `a.b.c(...)` files exist within `cpy`. Collectively they constitute a single ddname entry that members can be resolved within
    - these, along with recursive lib resolution, are what allow `%include member` to work
- supporting resolution of ddnames in recursive libs
- updating search to support member resolution with & without a ddname
- refactor existing IncludeItem types to split up member & file cases
- support process group lib resolution without tripping on ddnames
- numerous other changes to support these includes across the stack

The later case is in supplement to our existing file resolution strategy, where `%include abc` would resolve to a file named `abc` with one of the supported extensions. This is still the case, but we'll first attempt to match to a member of an existing ddname entry before trying files.

This does _not_ add:
- quick fix support for resolving unresolved `%include ddname(member);` or `%include member` entries, but the existing quick fix support for files works as is
- proper range for unresolved `ddname(member)` diagnostics (diagnostic is still on the member at the moment). The regular `member` case works as expected.

There's a _lot_ of changes here for a relatively small feature, so I'm very open to feedback for issues, improvements, thoughts, etc.. I've done my best to try and document new additions & logic clearly to make it clear why changes were made at various stages, but feel free to make further suggestions if there's anything else that's missing or unclear in terms of documented behavior.

Some other notes:
- Noticed syntax highlighting needs to be improved for some ddname includes, such as `%INCLUDE D.E.F`. This should be spun out into a separate issue, could benefit from semantic highlighting support.

### DIFF
--- a/packages/language/src/language-server/definition-request.ts
+++ b/packages/language/src/language-server/definition-request.ts
@@ -69,7 +69,8 @@ export function definitionRequest(
     ];
   } else if (
     isIncludeItemToken(token.kind) &&
-    (token.element?.kind === SyntaxKind.IncludeItem ||
+    (token.element?.kind === SyntaxKind.IncludeItemFile ||
+      token.element?.kind === SyntaxKind.IncludeItemMember ||
       token.element?.kind === SyntaxKind.InscanDirective)
   ) {
     // allow jumping to the resolved file when present
--- a/packages/language/src/language-server/definition-request.ts
+++ b/packages/language/src/language-server/definition-request.ts
@@ -73,6 +73,8 @@ export function definitionRequest(
       token.element?.kind === SyntaxKind.IncludeItemMember ||
       token.element?.kind === SyntaxKind.InscanDirective)
   ) {
+    // TODO @montymxb Nov 7th, 2025: Highlighting for members is incomplete, only gets the member & not the full ddname.
+    // see: https://github.com/zowe/zowe-pli-language-support/issues/466
     // allow jumping to the resolved file when present
     const filePath = token.element.filePath;
     if (filePath) {
--- a/packages/language/src/parser/preprocessor-parser.ts
+++ b/packages/language/src/parser/preprocessor-parser.ts
@@ -596,23 +596,15 @@ function includeStatement(state: ParserState): ast.IncludeDirective {
     // collect all ID tokens that can contribute to a ddname
     const idTokens: t.Token[] = [];
     let nextIdToken: t.Token | null = null;
-    while (true) {
-      nextIdToken = state.tryConsume(
+    while (
+      (nextIdToken = state.tryConsume(
         item,
         CstNodeKind.IncludeItem_FileID,
         t.ID,
-      );
-      if (!nextIdToken) {
-        break;
-      }
+      ))
+    ) {
       idTokens.push(nextIdToken);
-
-      const dotToken = state.tryConsume(
-        item,
-        CstNodeKind.IncludeItem_Dot,
-        t.Dot,
-      );
-      if (!dotToken) {
+      if (!state.tryConsume(item, CstNodeKind.IncludeItem_Dot, t.Dot)) {
         break;
       }
     }
--- a/packages/language/src/parser/preprocessor-parser.ts
+++ b/packages/language/src/parser/preprocessor-parser.ts
@@ -591,7 +591,8 @@ function includeStatement(state: ParserState): ast.IncludeDirective {
   directive.idempotent = isXInstruction(includeToken);
   let parseError = false;
   while (true) {
-    let item: ast.IncludeItemFile | ast.IncludeItemMember | undefined = undefined;
+    let item: ast.IncludeItemFile | ast.IncludeItemMember | undefined =
+      undefined;
     // collect all ID tokens that can contribute to a ddname
     const idTokens: t.Token[] = [];
     let nextIdToken: t.Token | null = null;
--- a/packages/language/src/parser/preprocessor-parser.ts
+++ b/packages/language/src/parser/preprocessor-parser.ts
@@ -591,9 +591,7 @@ function includeStatement(state: ParserState): ast.IncludeDirective {
   directive.idempotent = isXInstruction(includeToken);
   let parseError = false;
   while (true) {
-    const item = ast.createIncludeItem() as
-      | ast.IncludeItemFile
-      | ast.IncludeItemMember;
+    let item: ast.IncludeItemFile | ast.IncludeItemMember | undefined = undefined;
     // collect all ID tokens that can contribute to a ddname
     const idTokens: t.Token[] = [];
     let nextIdToken: t.Token | null = null;
@@ -620,6 +618,11 @@ function includeStatement(state: ParserState): ast.IncludeDirective {
 
     if (idTokens.length > 1) {
       // ddname which requires a parenthesized member portion
+      item = ast.createIncludeItemMember();
+      for (const idToken of idTokens) {
+        idToken.element = item;
+      }
+
       state.consume(item, CstNodeKind.IncludeItem_OpenParen, t.OpenParen);
 
       const memberToken = state.consume(
@@ -628,16 +631,19 @@ function includeStatement(state: ParserState): ast.IncludeDirective {
         t.ID,
       );
 
-      const itemWithMember = item as ast.IncludeItemMember;
-
       state.consume(item, CstNodeKind.IncludeItem_CloseParen, t.CloseParen);
       // joint ddname
-      itemWithMember.ddname = idTokens.map((t) => t.image).join(".");
-      itemWithMember.ddnameTokens = idTokens;
+      item.ddname = idTokens.map((t) => t.image).join(".");
+      item.ddnameTokens = idTokens; // <-- TODO use these tokens to report chained diagnostics, do I already do this?
 
-      itemWithMember.memberName = memberToken?.image ?? null;
-      itemWithMember.token = memberToken;
+      item.memberName = memberToken?.image ?? null;
+      item.token = memberToken;
     } else if (idTokens.length === 1) {
+      item = ast.createIncludeItemMember();
+      for (const idToken of idTokens) {
+        idToken.element = item;
+      }
+
       // either ddname w/ member, or raw member
       const ddnameOrMember = idTokens[0].image;
 
@@ -659,20 +665,22 @@ function includeStatement(state: ParserState): ast.IncludeDirective {
 
         state.consume(item, CstNodeKind.IncludeItem_CloseParen, t.CloseParen);
 
-        const itemWithMember = item as ast.IncludeItemMember;
+        item.ddname = ddnameOrMember;
+        item.ddnameTokens = idTokens;
 
-        itemWithMember.ddname = ddnameOrMember;
-        itemWithMember.ddnameTokens = idTokens;
-
-        itemWithMember.memberName = memberToken?.image ?? null;
-        itemWithMember.token = memberToken;
+        item.memberName = memberToken?.image ?? null;
+        item.token = memberToken;
       } else {
         // member include case
-        const itemWithMember = item as ast.IncludeItemMember;
-        itemWithMember.memberName = ddnameOrMember;
-        itemWithMember.token = nextIdToken;
+        item.memberName = ddnameOrMember;
+        item.token = nextIdToken;
       }
     } else {
+      item = ast.createIncludeItemFile();
+      for (const idToken of idTokens) {
+        idToken.element = item;
+      }
+
       // direct file include, not a member (from string)
       const stringToken = state.tryConsume(
         item,
@@ -680,10 +688,9 @@ function includeStatement(state: ParserState): ast.IncludeDirective {
         t.STRING_TERM,
       );
       if (stringToken) {
-        const itemWithFile = item as ast.IncludeItemFile;
         const fileName = unpackCharacterValue(stringToken.image);
-        itemWithFile.fileName = fileName;
-        itemWithFile.token = stringToken;
+        item.fileName = fileName;
+        item.token = stringToken;
       } else {
         // At least one include item is required
         // If none is found, we will report an error at the end of the statement
@@ -719,7 +726,7 @@ export function includeAltStatement(
     CstNodeKind.IncludeAltDirective_INCLUDE_ALT,
     t.INCLUDE_ALT,
   );
-  const item = ast.createIncludeItem() as ast.IncludeItemFile;
+  const item = ast.createIncludeItemFile();
   const token = state.consume(item, CstNodeKind.IncludeItem_FileID, t.ID);
   const fileName = token?.image ?? null;
   item.fileName = fileName;
--- a/packages/language/src/parser/preprocessor-parser.ts
+++ b/packages/language/src/parser/preprocessor-parser.ts
@@ -591,27 +591,99 @@ function includeStatement(state: ParserState): ast.IncludeDirective {
   directive.idempotent = isXInstruction(includeToken);
   let parseError = false;
   while (true) {
-    const item = ast.createIncludeItem();
-    const idToken = state.tryConsume(
-      item,
-      CstNodeKind.IncludeItem_FileID,
-      t.ID,
-    );
-    if (idToken) {
-      const fileName = idToken.image;
-      item.fileName = fileName;
-      item.token = idToken;
+    const item = ast.createIncludeItem() as
+      | ast.IncludeItemFile
+      | ast.IncludeItemMember;
+    // collect all ID tokens that can contribute to a ddname
+    const idTokens: t.Token[] = [];
+    let nextIdToken: t.Token | null = null;
+    while (true) {
+      nextIdToken = state.tryConsume(
+        item,
+        CstNodeKind.IncludeItem_FileID,
+        t.ID,
+      );
+      if (!nextIdToken) {
+        break;
+      }
+      idTokens.push(nextIdToken);
+
+      const dotToken = state.tryConsume(
+        item,
+        CstNodeKind.IncludeItem_Dot,
+        t.Dot,
+      );
+      if (!dotToken) {
+        break;
+      }
+    }
+
+    if (idTokens.length > 1) {
+      // ddname which requires a parenthesized member portion
+      state.consume(item, CstNodeKind.IncludeItem_OpenParen, t.OpenParen);
+
+      const memberToken = state.consume(
+        item,
+        CstNodeKind.IncludeItem_MemberID,
+        t.ID,
+      );
+
+      const itemWithMember = item as ast.IncludeItemMember;
+
+      state.consume(item, CstNodeKind.IncludeItem_CloseParen, t.CloseParen);
+      // joint ddname
+      itemWithMember.ddname = idTokens.map((t) => t.image).join(".");
+      itemWithMember.ddnameTokens = idTokens;
+
+      itemWithMember.memberName = memberToken?.image ?? null;
+      itemWithMember.token = memberToken;
+    } else if (idTokens.length === 1) {
+      // either ddname w/ member, or raw member
+      const ddnameOrMember = idTokens[0].image;
+
+      // check to see if we have an opening parenthesis for a following member
+      // making this a ddname + member
+      const openParenToken = state.tryConsume(
+        item,
+        CstNodeKind.IncludeItem_OpenParen,
+        t.OpenParen,
+      );
+
+      if (openParenToken) {
+        // ddname(member) case
+        const memberToken = state.consume(
+          item,
+          CstNodeKind.IncludeItem_MemberID,
+          t.ID,
+        );
+
+        state.consume(item, CstNodeKind.IncludeItem_CloseParen, t.CloseParen);
+
+        const itemWithMember = item as ast.IncludeItemMember;
+
+        itemWithMember.ddname = ddnameOrMember;
+        itemWithMember.ddnameTokens = idTokens;
+
+        itemWithMember.memberName = memberToken?.image ?? null;
+        itemWithMember.token = memberToken;
+      } else {
+        // member include case
+        const itemWithMember = item as ast.IncludeItemMember;
+        itemWithMember.memberName = ddnameOrMember;
+        itemWithMember.token = nextIdToken;
+      }
     } else {
+      // direct file include, not a member (from string)
       const stringToken = state.tryConsume(
         item,
         CstNodeKind.IncludeItem_FileString,
         t.STRING_TERM,
       );
       if (stringToken) {
+        const itemWithFile = item as ast.IncludeItemFile;
         const fileName = unpackCharacterValue(stringToken.image);
-        item.fileName = fileName;
-        item.string = true;
-        item.token = stringToken;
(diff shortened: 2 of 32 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1813,17 +1816,27 @@ async function runIncludeInstruction(
...
+    } else if (ast.isIncludeItemMember(item)) {
+      includeItem = {
+        memberName: item.memberName,
+        ddname: item.ddname,
+        ddnameTokens: item.ddnameTokens,
+        token: item.token,
+        idempotent: instruction.idempotent,
+      };
+    } else {
+      continue;
```
ssmifi, line 1836: _Question_: Is this `continue` only executed if the ast node is incomplete? Perhaps we should add a comment here.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1880,13 +1931,26 @@ async function runInclude(
...
+          : item.memberName,
     );
-    if (item.fileName)
+
+    if (isFileIncludeItem(item)) {
       diagnostic.data = {
         unresolvedFile: item.fileName,
         entryUri: context.entryUri.toString(),
       };
+    } else if (isMemberIncludeItem(item)) {
```
ssmifi, line 1945: _Minor_: Isn't it a `MemberIncludeItem` automatically?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1993,31 +2057,96 @@ async function resolveIncludeFileUri(
...
   if (pgroup) {
     // lib file as either a string or a member from a known process group
-    const absPathRegex = /^(?:\/|\\|[A-Z]:)/i;
     const computedLibs = pgroup.$computedLibs;
-    for (const lib of computedLibs) {
-      let libFileUri: URI;
-      if (!absPathRegex.test(lib)) {
+
+    // construct the appropriate file name or partial name for members
+    let fileNameOrPartial: string;
```
ssmifi: _Minor_: If we provide a naming function outside of this function, it could be reused in the diagnostics above.

packages/language/src/syntax-tree/ast.ts
```diff
@@ -1613,24 +1613,64 @@ export function createIncludeAltDirective(): IncludeAltDirective {
...
+ * Include item by file name
+ */
+export type IncludeItemFile = IncludeItem & {
+  fileName: string | null;
+};
+
+/**
+ * Include item by member name + optional ddname
+ */
+export type IncludeItemMember = IncludeItem & {
```
msujew: Suggestion: I'm not a big fan of breaking with the convention that we have throughout the file of having one `SyntaxKind` per type. IMO these should be distinct interfaces. Also, please add them to the `SyntaxNode` union type, which currently only contains `IncludeItem`.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1993,31 +2057,96 @@ async function resolveIncludeFileUri(
...
+    // used to denote whether we're working with a pure member (requires special lookup handling)
+    let isPureMember = false;
```
ssmifi: _Minor_: Could we describe what this means in more detail? Is this always located in `ddLib`? Add an example to the comment that shows how a pure member looks like.

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -136,21 +221,26 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
         }
       }
     }
+    if (entries.length === 0) {
+      // when no files are found, we assume the directory does not exist
+      // as we don't currently support creating empty dirs in the vfs
```
ssmifi, line 182: _Question_: Even if we would, should we create a new directory automatically on a read request?

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -160,34 +250,59 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
...
-    const extensions = options.extensions ?? [];
-
-    if (!options.global) {
-      if (this.files.has(searchPath)) {
-        return options.path;
+    if (isPathSearch(options)) {
+      // perform a search with a uri
+      const searchPath = options.path
+        .toString(true)
+        .toLowerCase()
```
ssmifi, line 213: _Question_: Throughout the whole function: How is case-sensitivity handled exactly? Different cases are not supported? What about cases where the original token contained a string with different cases?

packages/vscode-extension/src/language/main.ts
```diff
@@ -60,19 +62,49 @@ class NodeFileSystemProvider implements FileSystemProvider {
...
+        `${options.dirPath.fsPath}**/*\\(${options.member}\\)`,
+        {
+          nodir: true,
+          absolute: true,
+          nocase: true,
+        },
+      );
+    } else {
+      // member lookup w/ dd path (partial lib file as entry)
+      const uri = URI.parse(options.ddPath + `(${options.member})`);
```
msujew: Suggestion: Isn't this just the normal, non-global file search using a `<ddPath>(<member>)` syntax? IMO we can represent this using the normal `PathSearch` object.

Note that using `fs.access` will perform a case-sensitive match on the file system. This might fail on Linux, since PL/I is generally case insensitive. That why we use glob even for simple file lookups (for the `nocase` option).

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -82,11 +82,19 @@ export interface ProcessGroup {
...
    */
-  $computedLibs: string[];
+  $computedLibs: LibsEntry[];
+
+  /**
+   * Set of computed libs for fast lookup.
+   * Only contains directory entries, not DD name entries, which are partial.
+   * Used to find process groups from a lib URI
+   */
+  $computedLibsSet: Set<string>;
```
ssmifi, line 95: _Question_: Couldn't `computedLibs` be a set from the start? Feels a bit odd to have two computed libs "sets" that only partially overlap.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -356,30 +391,66 @@ export class PluginConfigurationProvider {
...
+                  // found a ddname-style entry, add full lib & break out, only need one to confirm
+                  computedLibs.add({
+                    ddLib: lib,
+                  });
+                  break;
+                }
+              }
+            } catch (parentError) {
+              // parent directory also failed to read, skip this lib
+              console.warn(`Failed to resolve library entry "${lib}"`);
```
ssmifi: _Question_: Will this eventually (maybe in some function further down the stack) result in a proper diagnostic or popup? I can imagine that mistakes can happen quite easily in these configurations and be confusing without proper notification.

packages/vscode-extension/src/common/file-search.ts
```diff
@@ -18,39 +23,92 @@ export async function searchFiles(
...
+    } else {
+      start = "/";
+    }
+
+    // iterate through segments except last to find correct casing for each
+    for (let i = 0; i < segments.length - 1; i++) {
+      const dir = await readDirSafe(start);
+      const segment = segments[i];
+      const fsSegment = dir.find(
+        (d) => d.toLowerCase() === segment.toLowerCase(),
```
ssmifi: _Question_: Same as before, can we just assume, everything is lowercase?

packages/language/test/fourslash/preprocessor/include/from-lib-path-recursive.ts
```diff
@@ -12,6 +12,7 @@
 /// <reference path="../../framework.ts" />
 
 // trigger a reparsing of the config, so we can pick up the sub libs folder
+// TODO @montymxb Oct 17th, 2025: Replace with a fourslash utility to trigger config reload
```
ssmifi, line 15: _Question_: How would that look like? What do you have in mind here? Perhaps this should be included in this PR to help manage the file system's complexity?

packages/language/test/fourslash/preprocessor/include/test-ddmember-include-members.ts
```diff
@@ -0,0 +1,45 @@
...
+////             "cpy2/A.B.C"
+////         ]
+////         }
+////     ]
+//// }
+
+// @filename: cpy2/MYLIB(m1)
+//// DECLARE LIB_VAR1 FIXED;
+
+// @filename: cpy2/A.B.C(m2)
```
ssmifi, line 41: _Comment_: This looks rather confusing by mixing different file systems. Should be ok in the test system, because it is just a name, but is this also a realistic case, e.g., on zOS?

ssmifi (commented):
Thanks for the extensive work. The system is getting quite complex/detailed. I have a couple of questions mostly regarding future maintainability. We can also discuss in the internal meeting tomorrow. 

In general, I would be quite happy if we could provide more tests for the different file systems because I have the feeling that things can break quite easily as the system grows. Same is true for the different clients, at least eventually.

packages/language/test/fourslash/preprocessor/include/test-ddmember-includes-registered-libs.ts
```diff
@@ -0,0 +1,45 @@
...
+
+// @filename: cpy2/MYLIB(m1)
+//// DECLARE LIB_VAR1 FIXED;
+
+// @filename: cpy2/A.B.C(m2)
+//// DECLARE LIB_VAR2 FIXED;
+
+// @filename: main.pli
+//// %INCLUDE m1;
+//// %INCLUDE m2;
```
msujew: Suggestion: Can we also test that the diagnostic marker appears here?

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -356,30 +391,66 @@ export class PluginConfigurationProvider {
...
-            for (const dirEntry of entries) {
-              const fileName = dirEntry.name;
-              const fileType = dirEntry.type;
-              if (fileType === DirEntryType.Directory) {
-                // directory to add for handling
-                libsToProcess.push(`${lib}/${fileName}`);
-                // also add to the full libs list
-                computedLibs.add(`${lib}/${fileName}`);
+          try {
+            const entries = await FileSystemProviderInstance.readDir(libUri);
```
msujew, line 422: Suggestion: Add a `stat` method to the `FileSystemProvider` to resolve this.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1993,31 +2057,96 @@ async function resolveIncludeFileUri(
...
+        if (isPureMember) {
+          // attempt an early resolution for any DDName that introduces this member
+          const memberMatch = await FileSystemProviderInstance.search({
+            dirPath: resolveLibFileUri(lib.dir),
+            member: fileNameOrPartial,
+          });
+          if (memberMatch) {
+            return memberMatch;
+          }
+        }
```
msujew: Question: Is this a performance optimization? It seems like that, because otherwise it will just fall through to the standard search.

msujew (commented):
I think we really need some documentation in addition to the code/tests that explain how each style of include is supposed to be resolved. I also believe that we can consolidate some of the include resolution logic, because this has become a lot of "if this, then that" on multiple levels (i.e. search - with its various implementations, URI resolution in the interpreter, etc.).

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: Yep, that's it. If for any reason we're lacking a legitimate entry we shouldn't proceed to resolve it. I'll add a note to clarify the purpose.

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: Based on the type it is, but for the same reason as the continue usage above we may not always have a complete object, which would prevent us from producing a good data payload here.

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: Totally agree with this, I'll factor that out & reuse

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: Can do! Yeah it's quite complicated, but I can boil it down to a some comments that should help factor clarify what each case would look like.

packages/language/src/workspace/file-system-provider.ts
montymxb: I would avoid automatically generating a directory on request, otherwise we get a bunch of test failures. This is one of those points that I'm not too happy with at the moment given this isn't entirely correct. It's a stop-gap to avoid a rather large chunk of failures across our test suite, otherwise I would prefer to omit it. I'll need to go further into this one & find a better solution most likely.

packages/language/src/workspace/file-system-provider.ts
montymxb: Currently case is generally lower-cased, but you had a good point again before about ext4 and such, where that's invalid. For context in the vfs we probably want this to normalize, but for the node fs implementation that's probably not be sensible. Hmm, need to think about this a bit more.

packages/language/src/workspace/plugin-configuration-provider.ts
montymxb: We have the other one as an array since we need to retrieve those values for directory look-ups. This one is just for checking whether a dir entry is present that's an actual directory (and not a ddname entry). We could implement the same check using the array alone (I had this initially at some point), but we have to walk through it & skip the ddname entries. This just avoids the redundant checking.

packages/language/src/workspace/plugin-configuration-provider.ts
montymxb: Yep 100% we should have this be a diagnostic. The question is still where to put it, since it's derived from the config again, but having at least _some_ feedback that there's an issue outside of the output log would be nice. I'm not sure if I want to put it up in this PR however, would it make sense to divide up as a follow-up task?

packages/vscode-extension/src/common/file-search.ts
montymxb: In general, no we can't assume that, you're right about that. But this implementation is just leveraged in the browser at the moment with the vfs, where we're doing case insensitive look-ups (at least I'm pretty sure that's the case). However this code is in the common section, which implies it probably should _not_ make the browser based assumption, which means this wouldn't work if used elsewhere on an ext4 fs & such 🤔 . So this probably should be changed, I'll get back to it next week most likely.

packages/language/test/fourslash/preprocessor/include/from-lib-path-recursive.ts
montymxb: This came up as a request by @msujew in a previous PR for testing configs, since it's becoming a pattern to re-trigger a config reload by rewriting the file again. It's still on my TODO list, but while implementing this PR the same tactic is used in these 4slash tests as well, so I just wanted to extend the same mention here.

My plan is to do that as a separate task so it doesn't expand this one further. There's about 3 or 4 tests that leverage this right now, and they work as is, it would just be an improvement to the testing infra. The problem is that we need to await async handling for the plugin configuration provider to reload fully, as far as I'm aware of we can't do that yet.

packages/language/test/fourslash/preprocessor/include/test-ddmember-include-members.ts
montymxb: It is an actual use case that was presented some weeks ago, but it's also confusing at first glance too. The file `A.B.C(m2)` is synced over from zOS to the user's machine, and corresponds to a lib entry on the mainframe (akin to `A/B/C/m2`). So both the unix & mainframe lib resolution strategies need to be present as far as I'm aware. There will be cases where large quantities of these ddname entries need to be recursively resolved from a given libs entry.

packages/language/src/syntax-tree/ast.ts
montymxb: That's fine, I can split up those up further, probably for the best. Early on this was causing problems as the same node type takes on entirely different semantics depending on how it's constructed.

packages/vscode-extension/src/language/main.ts
montymxb: Yes it is just the full path combined in this case, that can be done via a regular path search, I can swap that over.

It probably will fail. The ext4 case on Linux is something that needs further checking for sure, @ssmifi mentioned this as well too, there are a number of points where I'm too aggressively lower-casing things to get stuff to behave in the tests, and I haven't yet verified how that will behave in practice on linux, I need to go over that still.

packages/language/test/fourslash/preprocessor/include/test-ddmember-includes-registered-libs.ts
montymxb: Yep definitely, I can add a check for that.

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: It's a requested case for include resolution order in case we have a file `cpy/member.pli` and `cpy/ddname(member)`. I had to hash this out a bit, but it appears that in such a case we should be resolving the ddname w/ a matching member first when present over the literal `member` file entry. And when we can't we just fall through as you've noted. But the question points out that this isn't well documented, I'll add some additional comments to clarify the intent.

packages/language/src/workspace/file-system-provider.ts
montymxb: Ahh actually just realized this is the lower check and not the upper regarding `this.files.size === 0`, which is the one that I was thinking of. In this case we do want to throw since that matches the existing `fs` behavior for reading a directory that doesn't exist, and then our exception handling provides the appropriate recovery mechanism in both the nfs & vfs cases.

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: Renamed this to `isMemberWithoutDDName`, which is much clearer in purpose (plus comment details too)

packages/language/src/parser/preprocessor-parser.ts
```diff
@@ -591,27 +591,99 @@ function includeStatement(state: ParserState): ast.IncludeDirective {
...
+
+      const dotToken = state.tryConsume(
+        item,
+        CstNodeKind.IncludeItem_Dot,
+        t.Dot,
+      );
+      if (!dotToken) {
+        break;
+      }
+    }
```
msujew, line 610: Minor:
```suggestion
    let nextIdToken: t.Token | null = null;
    while (nextIdToken = state.tryConsume(
        item,
        CstNodeKind.IncludeItem_FileID,
        t.ID,
      )) {
      idTokens.push(nextIdToken);
      if (!state.tryConsume(
        item,
        CstNodeKind.IncludeItem_Dot,
        t.Dot,
      )) {
        break;
      }
    }
```

packages/language/src/workspace/file-system-provider.ts
montymxb: Thought about this a bit more, but I'm planning to leaving this as is for now, but I did add docs to the VFS indicating it's case-insensitive by design. For ext4 we should be okay via the existing node file system provider, as that doesn't lower case the uris. However we can definitely do a bit more to preserve the original casing in addition to the lower cased form here.

packages/language/src/workspace/plugin-configuration-provider.ts
montymxb: Setup diagnostic reporting for this one case, and that should be a nice step towards reporting additional config related diagnostics as well.

packages/vscode-extension/src/language/main.ts
montymxb: This has been since simplifies to align w/ the existing PathSearch approach.

montymxb:
~Pending documentation & and resolving the out-standing Windows fail.~

Resolved & ready for re-review.

packages/vscode-extension/src/common/file-search.ts
montymxb: Modified searchFiles to accept an additional option to determine whether checking is done w/ respect to case. By default it's case-insensitive, matching what it currently does, but leaves the door open to change this based on the circumstance & fs we're working within later on.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -348,39 +391,104 @@ export class PluginConfigurationProvider {
...
+            libUri = UriUtils.joinPath(URI.parse(this.workspacePath), lib);
+          }
+
+          try {
+            const entries = await FileSystemProviderInstance.readDir(libUri);
+            if (entries.length) {
+              for (const fileName of entries) {
+                const stats = await FileSystemProviderInstance.stat(
+                  UriUtils.joinPath(libUri, fileName),
+                );
```
msujew, line 429: Suggestion: This is more of a TODO in a separate PR. We should parallelize this, because as it is, this will block the language server for a relatively long time in case we have to go through a lot of directories here. Can you add a todo/create an issue for this?

packages/language/src/language-server/definition-request.ts
```diff
@@ -69,7 +69,8 @@ export function definitionRequest(
     ];
   } else if (
     isIncludeItemToken(token.kind) &&
-    (token.element?.kind === SyntaxKind.IncludeItem ||
+    (token.element?.kind === SyntaxKind.IncludeItemFile ||
+      token.element?.kind === SyntaxKind.IncludeItemMember ||
```
msujew, line 73: Todo: Highlight full source range when performing the definition request on the item:

Image

msujew (approved):
Looks good to me. I have a minor code improvement and some todos, which I'd be glad to see in a separate PR.

packages/language/src/workspace/plugin-configuration-provider.ts
montymxb: Can do. This was my initial concern when I wrote this up too, I had `withFileTypes` in mind with the prior iteration, but without the flag on our end. Perhaps that might be a better approach going forward to avoid the extra stat checks. I'll add an issue + TODO here for that momentarily.

montymxb: Added #465 for tracking.

packages/language/src/language-server/definition-request.ts
montymxb: Right, still needs to be done, I'll open an issue for that too.

I think we have a similar issue w/ the underlining on string includes as well from what I can see. Would that make sense to track them together?
Screenshot 2025-11-07 at 09 16 43

montymxb: Added #466 with both points for now.

Issue #442, opened 2025-10-14
Refactor compiler option TranslatorErrors to the generic Diagnostic interface


ssmifi:
This is a follow up from https://github.com/zowe/zowe-pli-language-support/pull/429: Refactor compiler option `TranslatorErrors` to the generic `Diagnostic` interface. To do this smoothly, all of the older compiler options should be refactored to the newer functions, including dedicated codes and harness tests. Potentially, in corner cases tests must be performed on the mainframe.

ssmifi:
Second part is to check, validate and add tests for:
- [x] ARCH, 
- [x] ASSERT, 
- [x] ATTRIBUTES, 
- [x] BACKREG, 
- [x] BITPREC, 
- [x] BLKOFF, 
- [x] CMPAT, 
- [x] COMMON, 
- [x] DBCS, 
- [x] DBRMLIB, 
- [x] DECOMP, 
- [x] DLL, 
- [x] DLLINIT, 
- [x] EXIT, 
- [x] EXPORTALL, 
- [x] EXTRN, 
- [x] FILEREF, 
- [x] FLOAT, 
- [x] FLOATINMATH, 
- [x] GOFF, 
- [x] recheck old PP tests, 
- [x] SYSPARM
- [x] SYSTEM

PR #456, opened 2025-10-28
Refactor TranslationError to the Diagnostic type

ssmifi:
Main part of https://github.com/zowe/zowe-pli-language-support/issues/442.
To keep the refactoring/reviews of for the old compiler options (+adding new tests) manageable, I split the task.
The PR contains the majority of the work:

- Refactored the TranslationError infrastructure to our global diagnostic type.
- Refactored old compiler options to the latest changes and added dedicated tests and confirmed behavior on the mainframe:
  - Aggregate, Attribute, Blank, Brackets, Case, CaseRules, Check, CodePage, Compile, CSect, CSectCut, Currency, DD, DDSQL, Decimal, Default, Deprecate, DeprecateNext, Display, Flag
- Removed old compiler options translator tests (except PP and SYSPARM).

_Suggestion_: Do we really need the `fullCode` field in the codes? As far as I can tell it is (at least currently) always `code + severity`, and it is getting a bit tedious to keep that in sync. For the compiler options, I tested it without `fullCode` in this PR. If you agree, I can remove it from the other codes as well. If you disagree, I will add the codes to the compiler options again.

The remaining work of https://github.com/zowe/zowe-pli-language-support/issues/442 is verifying and refactoring the options A - G and the SYSPARM which were not handled in this PR. I will add the list to the ticket. Since the main part of the refactoring should be concluded with this PR, the priority can be lowered, though.


packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -51,7 +52,7 @@ enum RuleAlignment {
 export class Translator<T extends CompilerOptionsPP = CompilerOptionsPP> {
   options: T;
   defaults: T;
-  issues: CompilerOptionIssue[] = [];
+  issues: Diagnostic[] = [];
```
msujew:
```suggestion
  diagnostics: Diagnostic[] = [];
```

packages/language/src/preprocessor/compiler-options/translator.ts
```diff
@@ -38,7 +39,7 @@ type Translate<T extends CompilerOptionsPP> = (
   acceptor: TranslationErrorAcceptor,
 ) => void;
 
-type TranslationErrorAcceptor = (error: TranslationError) => void;
+type TranslationErrorAcceptor = (error: Diagnostic) => void;
```
msujew:
```suggestion
type TranslationDiagnosticAcceptor = (diagnostic: Diagnostic) => void;
```

packages/language/src/validation/pli-codes.ts
msujew: Question: Can we also remove those `fullCode` fields? We should be able to use search-and-replace for them.

ssmifi: I removed all `fullCode` from the pli-codes. The harness now also accepts the codes directly to make the handling in the tests easier.

msujew (approved):
Thanks, looks good to me 👍

Issue #450, opened 2025-10-17
Improve Hover Response Caching


montymxb:
This is already noted in the [hover-request.ts](https://github.com/zowe/zowe-pli-language-support/blob/development/packages/language/src/language-server/hover-request.ts#L251-L252) file, which should be changed to a separate caching service. This is particularly important since previously computed hover results for libs will stay put even if the libs are no longer reachable from our config.

An example of this is checking the hover result for a lib that can be reached, forcing it to compute & cache, and then checking it again after its corresponding `libs` entry has been removed from the config. Hovering again will show the diagnostic for the unresolved include along with the cached hover result. We should invalidate those cache entries when the libs change.

PR #457, opened 2025-10-30
[Feature] Implement caching for hover

Wagner-Laranjeiras:
Issue: [#450](https://github.com/zowe/zowe-pli-language-support/issues/450)

This is a work in progress.

### DIFF
--- a/packages/language/src/language-server/cache/include-cache.ts
+++ b/packages/language/src/language-server/cache/include-cache.ts
@@ -32,9 +32,12 @@ export const includeCache = new LRUCache<string, IncludePreview>({
   sizeCalculation: (v) => (v.partialContent ? v.partialContent.length : 1),
   fetchMethod: async (key, staleValue, { context }) => {
     if (!context) throw new Error("Missing context for include preview");
-    const { unit, node } = context as { unit: CompilationUnit; node: IncludeItemNode };
+    const { unit, node } = context as {
+      unit: CompilationUnit;
+      node: IncludeItemNode;
+    };
     return loadIncludePreview(unit, node);
-  }
+  },
 });
 
 export function makeIncludeKey(
@@ -44,18 +47,6 @@ export function makeIncludeKey(
 ) {
   return `${unitId}::inc::${relativePath}::cfg=${configVersion ?? "v0"}`;
 }
-
-// export async function fetchIncludePreview(
-//   key: string,
-//   fetchFn: () => Promise<IncludePreview>,
-// ) {
-//   const cached = includeCache.get(key);
-//   if (cached) return cached;
-
-//   const value = await fetchFn();
-//   includeCache.set(key, value);
-//   return value;
-// }
 
 export async function fetchIncludePreviewWithCache(
   key: string,
@@ -85,11 +76,10 @@ export async function loadIncludePreview(
   const fileUri = URI.parse(node.filePath);
   const doc = unit.services.files.getDocument(fileUri);
   if (!doc) {
-    // unresolved now
     return { fileUri: null, partialContent: null, fetchedAt: Date.now() };
   }
 
-  // read up to N lines (same logic as before)
+  // read up to N lines
   const lineCutoff = 20;
   const fileContent = doc.getText({
     start: { line: 0, character: 0 },
--- a/packages/language/src/language-server/cache/include-cache.ts
+++ b/packages/language/src/language-server/cache/include-cache.ts
@@ -1,0 +1,102 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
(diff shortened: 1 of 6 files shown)


Wagner-Laranjeiras:
Please don't review it now Ben. I'm coming back here with a fresh pair of eyes and there's a lot that I can improve already. I'll let you know when it's decent enough, so you don't waste your time ;)

packages/language/src/language-server/cache/include-cache.ts
```diff
@@ -0,0 +1,92 @@
...
+  filePath: string | null;
+  relativeFilePath: string | null;
+}
+
+export const includeCache = new LRUCache<string, IncludePreview>({
+  max: 500,
+  ttl: 1000 * 60 * 5,
+  maxSize: 1000,
+  sizeCalculation: (v) => (v.partialContent ? v.partialContent.length : 1),
+  fetchMethod: async (key, staleValue, { context }) => {
```
montymxb, line 33: Would recommend we use `get` over `fetch` to avoid async modifying things, and set when a key is expired or missing. Otherwise this also complicates the test logic for getting hover results, since they'll be async as well.

packages/language/src/language-server/cache/include-cache.ts
```diff
@@ -0,0 +1,92 @@
...
+  configVersion?: string,
+) {
+  return `${unitId}::inc::${relativePath}::cfg=${configVersion ?? "v0"}`;
+}
+
+export async function fetchIncludePreviewWithCache(
+  key: string,
+  asyncLoader: () => Promise<IncludePreview>,
+) {
+  return (includeCache as any).fetch(key, { fetchMethod: asyncLoader });
```
montymxb, line 55: We should avoid `any` usage here, since it may very well be the case that `includeCache` may not be of the given type. A good approach would be to verify the type before using it, and handling the case where it's not as expected.

packages/language/src/language-server/hover-request.ts
```diff
@@ -229,50 +235,42 @@ function decodeBound(bound: Bound | null): string | null {
...
-      Array.from(lineCount).length > lineCutoff
-        ? fileContent + "\n...\n"
-        : fileContent;
-    // cache for later requests
-    node.sourceText = partialContent;
+  const doc = unit.services.files.getDocument(URI.parse(node.filePath));
+  if (!doc) {
+    return null;
+  }
+  const configVersion = String(
```
montymxb, line 254: Not sure why we're including this, but I presume it's to respond to the include keyword changing? Even so I'm not sure we'll need it in practice, but I'm curious what your reasoning was here.

In addition we wouldn't need to use the `String` constructor either, it's rarely used unless you really need a String object over a string primitive. Good rule of thumb is to always first use the `string` primitive as is, unless you have a very specific reason to use the object.

packages/language/src/language-server/hover-request.ts
```diff
@@ -355,28 +359,40 @@ const generateNameTokenMarkup: MarkupGenerator = ({ unit, token }) => {
...
-
-  return generators
-    .map(tryGenerate)
-    .filter((response): response is string => response !== null);
+): Promise<MarkupResponse[]> {
+  const results = await Promise.all(
+    // REVIEW THIS LATER
+    generators.map(async (gen) => {
+      try {
+        return await gen(context);
```
montymxb, line 370: No need to await if we want these to run concurrently, but I'm curious if there was a reason for wanting to rewrite this? If there wasn't an issue with it before I would favor leaving it in its original form.

I presume the problem was dealing with the async rewrite, but if you leverage `get` then this can part can be reverted.

packages/language/src/language-server/cache/include-cache.ts
```diff
@@ -0,0 +1,92 @@
...
+  fetchedAt: number;
+}
+
+export interface IncludeItemNode {
+  sourceText: string | null;
+  filePath: string | null;
+  relativeFilePath: string | null;
+}
+
+export const includeCache = new LRUCache<string, IncludePreview>({
```
montymxb, line 28: Instead of exporting the cache directly, we can export a smaller set of functions for r/w operations on the cache. The benefit of this is related to the fetch note below too, since we can then handle read (get) operations that would result in a cache miss, by generating the include preview, setting it in the cache, and also returning the result.

packages/language/src/language-server/hover-request.ts
```diff
@@ -229,50 +235,42 @@ function decodeBound(bound: Bound | null): string | null {
...
-  filePath: string | null;
-  relativeFilePath: string | null;
-}
-
 /**
  * Converts an IncludeItem or InscanDirective to a string representation.
  * Ex. %INCLUDE "/path/to/file.pli"
  */
-function getIncludeItemRepresentation(
+async function getIncludeItemRepresentation(
```
montymxb, line 242: If you use `get` for the cache retrieval you can drop all of the async propagation that's happening.

montymxb (commented):
Hey cool stuff here! Always a fan of lru-cache. I left some feedback for what's been setup so far. The biggest suggestion I could make is to use `get` over `fetch` to avoid switching everything over to async. You can handle the get in the include-cache.ts file with a function that can check to retrieve the preview when it's missing or stale, and externally everything will remain as is.

PR #458, opened 2025-11-03
Properly support type references

msujew:
Fixes point (1) of https://github.com/zowe/zowe-pli-language-support/issues/392.

Does quite a few things:
1. Add a reference `type` field that differentiates between type and variable scope objects.
2. Add type symbols in the symbol table.
3. Allow to resolve references for types.
4. Improve syntax highlighting for types.
5. Add a bunch of tests for (3) and (4).
6. Removes the `OrdinalTypeAttribute` - it is interchangeable with the existing `TypeAttribute`.
7. Removes the usage of FQN - it is not clear how they are supposed to be used for ordinals (and there are no examples).


packages/language/src/language-server/semantic-tokens.ts
```diff
@@ -156,13 +206,50 @@ function isVariableType(token: Token): boolean {
...
+  }
+  return false;
+}
+
+function isEnumToken(token: Token): boolean {
+  return token.kind === CstNodeKind.DefineOrdinalStatement_Name;
+}
+
+function isEnumMemberToken(token: Token): boolean {
+  if (token.kind === CstNodeKind.OrdinalValue_Name) {
```
Lotes: Could be a one-liner

Lotes: Minor: Also two-liners for the following functions, but I think there it is fine, because the expression is more complex.

packages/language/src/language-server/completion/keywords.ts
```diff
@@ -115,7 +115,7 @@ export const CompletionKeywords = {
     [CstNodeKind.EntryAttribute_Limited0, kw("LIMITED")],
     [CstNodeKind.LikeAttribute_LIKE, kw("LIKE")],
     [CstNodeKind.TypeAttribute_TYPE, kw("TYPE")],
-    [CstNodeKind.OrdinalTypeAttribute_ORDINAL, kw("ORDINAL")],
+    [CstNodeKind.TypeAttribute_TYPE, kw("ORDINAL")],
```
Lotes, line 118: The rename looks suspicious. It is a multimap, ok... but why using the same key? "Ordinals" are enums and "Type" are all other types, or?

packages/language/src/linking/symbol-table.ts
```diff
@@ -65,7 +62,7 @@ export class SymbolTable {
   ) {
     this.addSymbolDeclaration(
       text,
-      QualifiedSyntaxNode.createImplicit(this.unit, token, node),
```
Lotes, line 68: Good thing, that you can reduce complexity by removing this unit parameter.

packages/language/src/linking/resolver.ts
```diff
@@ -232,7 +232,11 @@ function getMatchingSymbols(
   qualifiedName: string[],
   reference: Reference,
   reporter: LinkerErrorReporter,
-): QualifiedSyntaxNode[] {
+): readonly QualifiedSyntaxNode[] {
+  if (reference.type === ReferenceType.Type) {
```
Lotes, line 236: This is a very short branch. Looking at the non-type branch: Why is it not possible to have a duplicated name / ambiguous symbols in the type branch?

Lotes (commented):
I have more questions than advices. See below.

packages/language/src/language-server/completion/keywords.ts
msujew: See https://github.com/zowe/zowe-pli-language-support/blob/ee782c766d0d61e624013b8842236fa17395e159/packages/language/src/parser/parser.ts#L6803-L6812

"TYPE" can also be used for ordinals. Originally found this behavior [here](https://www.microfocus.com/documentation/enterprise-developer/ed90/ED-VS2022/GUID-2F99C754-71C7-4A3A-BA55-C364CFB1B41C.html) and validated this against our compiler.

packages/language/src/linking/resolver.ts
msujew: I've rechecked this with the compiler. Instead of complaining about an ambiguous symbol, it will instead raise `IBM1432I`: `The type A is already defined. The redefinition is ignored.` on the declaration of the duplicate type.

IMO, we can tackle this in a separate PR.

Lotes (approved):

PR #459, opened 2025-11-03
False-Positive-Fix: FILE-typed variables with INPUT

Lotes:
`INPUT` / `OUTPUT` / `UPDATE` are transmission direction attributes.
Before this change they were only taking care that the variable becomes `FILE`-typed, which caused a false positive when the variable was already declared as `FILE` (it was flagged as a duplicated attribute).


msujew (approved):
Great, looks good to me 👍

PR #460, opened 2025-11-04
Support LIKE/TYPE attribute linking

msujew:
Fixes point (2) of https://github.com/zowe/zowe-pli-language-support/issues/392.

This change does quite a lot of things:
1. It splits linking into two phases for two different "priorities" of references
    1. References within `LIKE` and `TYPE` attributes receive the high priority
    2. All other references receive the "normal" priority
2. In between those two linking phases, the newly resolved references go through the symbol table generation again.
3. In order to support linking to structure members, they are now parsed as `DeclaredItem` elements. This massively simplifies the symbol table generation.
4. Since we might run symbol table generation multiple times on the same element, we now cache symbol keys
5. Same with validations

### DIFF
--- a/packages/language/src/linking/declared-item-parser.ts
+++ b/packages/language/src/linking/declared-item-parser.ts
@@ -22,7 +22,7 @@ import { MultiMap } from "../utils/collections";
 import { getNameToken } from "./tokens";
 import { LinkerErrorReporter } from "./error";
 import { Token } from "../parser/tokens";
-import { getExtendingDeclaredItems } from "./util";
+import { getExtendingDeclaredItems } from "../syntax-tree/ast-utils";
 
 type UnrolledItem = {
   kind: SyntaxKind;
@@ -66,6 +66,12 @@ function getDeclaredItemToken(node: SyntaxNode): Token | undefined {
 
   return getNameToken(node);
 }
+
+/**
+ * The starting index for levels in PL/I.
+ * Levels are 1-based.
+ */
+const LEVEL_START_INDEX = 1;
 
 /**
  * Unrolls factorized declarations.
@@ -129,7 +135,7 @@ function unrollFactorized(
               extended.extendingItems,
               extended.target,
               // If no level is specified on the extended item, we assume level 1.
-              item.level ?? 1,
+              item.level ?? LEVEL_START_INDEX,
               // Create a copy of the visited set for each branch
               new Set(visited),
             ),
@@ -173,13 +179,14 @@ function unrollSubtree(
       }
       // Virtually increase the level of the child item
       // This ensures that the child item is always deeper than the parent item.
-      item.level += level - 1;
+      // Subtract the LEVEL_START_INDEX since the levels are 1-based.
+      item.level += level - LEVEL_START_INDEX;
       // Still in the subtree. Push the child item
       result.push(item);
     } else if (item.node === node) {
       // We have found the starting node of the subtree.
       // Assume level 1 if not specified
-      foundLevel = item.level ?? 1;
+      foundLevel = item.level ?? LEVEL_START_INDEX;
     }
   }
   return result;
@@ -246,7 +253,7 @@ export class DeclaredItemParser {
 
     // When level is not set, we assume 1 like the PL/I compiler seems to do.
     if (level === null) {
-      return 1;
+      return LEVEL_START_INDEX;
     }
 
     // TODO: get max level from compilation unit? If e.g. compilation flags can change this.
--- a/packages/language/src/linking/declared-item-parser.ts
+++ b/packages/language/src/linking/declared-item-parser.ts
@@ -22,13 +22,13 @@ import { MultiMap } from "../utils/collections";
 import { getNameToken } from "./tokens";
 import { LinkerErrorReporter } from "./error";
 import { Token } from "../parser/tokens";
+import { getExtendingDeclaredItems } from "./util";
 
 type UnrolledItem = {
   kind: SyntaxKind;
   level: number | null;
   levelToken: Token | null;
   node: DeclaredVariable | WildcardItem;
-  container: SyntaxNode;
 };
 
 /**
@@ -88,7 +88,7 @@ function getDeclaredItemToken(node: SyntaxNode): Token | undefined {
  */
 function unrollFactorized(
   rolledItems: readonly DeclaredItemElement[],
-  parent: DeclaredItem | null = null,
+  visited: Set<DeclaredVariable>,
 ): UnrolledItem[] {
   const items = rolledItems.slice(); // Explicitly make a copy of the items, to use `.shift()`
   const variables: UnrolledItem[] = [];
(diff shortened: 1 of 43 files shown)


packages/language/src/linking/qualified-syntax-node.ts
```diff
@@ -91,6 +91,24 @@ export class QualifiedSyntaxNode {
...
+      parts.push("implicit");
+    }
+    let current: QualifiedSyntaxNode | null = this;
+    while (current) {
+      // The ID of the token uniquely identifies the node in the source code
+      // It can sometimes be undefined, still push something to avoid collisions
+      parts.push(current.token.id?.toString() || "NaN");
+      current = current.parent;
+    }
+    return parts.join("_");
```
ssmifi, line 109: _Minor_: Why not push `undefined` if undefined? 
Also, makes no real difference, but why use `_` instead `.` for some kind of qualified name?

_Edit_: Ok, `id` is a number, so I suppose `NaN` (and probably the underscore) makes sense. :)

packages/language/src/linking/resolver.ts
```diff
@@ -127,10 +135,19 @@ export class ReferencesCache {
     return this.reverseMap.get(node) || [];
   }
 
-  allReferences(): Reference[] {
+  priorityReferences(): Reference[] {
```
ssmifi, line 141: _Question_: Should we add the information from the PR to the functions (or elsewhere) as comments to make the explanation of the the priorities persistant?

_Edit_: Ok, this is more or less explained in `resolveReferences`. Shouldn't we add a hint here as well?

packages/language/src/linking/symbol-table.ts
```diff
@@ -321,21 +332,47 @@ function assignRedeclaredSymbols(scopeCache: ScopeCache) {
...
+    referencesCache: unit.referencesCache,
+    scopeCache: unit.scopeCaches.regular,
+    statementOrderCache: unit.statementOrderCache,
+    reporter,
+  };
+  for (const node of symbols) {
+    const scope = unit.scopeCaches.regular.get(node);
+    if (!scope) {
+      // Something went wrong, skip this node.
+      continue;
```
ssmifi, line 355: _Question_: Shouldn't this be reported anywhere? At least in debug?

packages/language/src/linking/symbol-table.ts
```diff
@@ -321,21 +332,47 @@ function assignRedeclaredSymbols(scopeCache: ScopeCache) {
...
+    }
+    // Second iteration over the node to handle newly added symbols.
+    // The node should already be present in the scope cache.
+    // Duplicate symbols will be filtered out by the symbol table.
+    handleNode(node, scope, context);
```
ssmifi, line 360: _Question_: This isn't really asserted from the perspective of the function, isn't it? Can we add some sort of assertion for future maintenance or isn't this favorable?

packages/language/src/syntax-tree/ast-utils.ts
```diff
@@ -0,0 +1,88 @@
...
+  DeclarationAttribute,
+  DeclaredItem,
+  DeclaredVariable,
+  DefineStructureStatement,
+  getContainer,
+  SyntaxKind,
+  TypeExtendingAttribute,
+} from "../syntax-tree/ast";
+
+export function getTypeExtendingAttribute(
```
ssmifi, line 22: _Minor Structure Question_: True, all functions here are currently only used in linking, but aren't these general ast utils and shouldn't be located there?

packages/language/src/validation/diagnostics-store.ts
```diff
@@ -0,0 +1,78 @@
...
+  TypeSystem,
+  Validation,
+}
+
+export class DiagnosticsStore {
+  private keys = new Set<string>();
+  private diagnostics: Diagnostic[][] = [];
+
+  constructor() {
+    for (let i = 0; i <= DiagnosticCategory.Validation; i++) {
```
ssmifi: _Minor_: Assumes that `Validation` is the last index. Could make this safer for future extensions or add comment to `DiagnosticCategory`.

packages/language/src/workspace/lifecycle.ts
```diff
@@ -92,22 +97,16 @@ export function parse(compilationUnit: CompilationUnit): Program {
 }
 
 export function generateSymbolTable(compilationUnit: CompilationUnit) {
-  compilationUnit.diagnostics.symbolTable = iterateSymbols(compilationUnit);
+  iterateSymbols(compilationUnit);
```
ssmifi, line 100: _Question_: Hm. Feels a bit odd that half of the diagnostics are added to the store in `lifecycle` and some are now added directly via a reporter from the unit. Should we normalize this (maybe in a follow-up PR)?

packages/language/src/linking/declared-item-parser.ts
```diff
@@ -111,28 +111,80 @@ function unrollFactorized(
...
+  let foundLevel: number | undefined = undefined;
+  for (const item of unrolled) {
+    if (foundLevel !== undefined) {
+      if (item.level === null || item.level <= foundLevel) {
+        // We've reached the end of the subtree.
+        break;
+      }
+      // Virtually increase the level of the child item
+      // This ensures that the child item is always deeper than the parent item.
+      item.level += level - 1;
```
ssmifi: _Question_: Sorry, don't see it. Why -1?

ssmifi (commented):
Nice work. Thanks. I just have a couple of small questions.

packages/language/src/linking/declared-item-parser.ts
msujew: I wasn't quite sure how to explain this, so here's an example:

```
DCL 1 A, 2 B, 3 C;
DCL 1 X, 2 Y LIKE A;
```

Since `Y` will "take up" all of the subelements of `A`, it will essentially become:
```
DCL 1 A, 2 B, 3 C;
DCL 1 X, 2 Y, 3 B, 4 C;
```

Since `X` already has 1 level (it's the default, even if not explicit), we need to subtract that one level to correctly arrive at 3, 4 for B, C respectively.

packages/language/src/workspace/lifecycle.ts
msujew: Yep, let's do this in a follow up :)

packages/language/src/linking/symbol-table.ts
msujew: What kind of assertion do you mean? For what?

ssmifi: The assumptions in the comments: second iteration, node should be present, filtered symbols.

packages/language/src/workspace/lifecycle.ts
ssmifi: Do we follow up immediately or should we add a ticket for this?

packages/language/src/linking/symbol-table.ts
msujew: Asserting all of that is quite costly... Also, a bunch of tests will fail (in rather obvious ways) in case anything of that behavior breaks.

packages/language/src/linking/resolver.ts
msujew: I've added a hint to the top of the class 👍

packages/language/src/syntax-tree/ast-utils.ts
msujew: I've moved them to a `ast-utils` file.

ssmifi (approved):
LGTM, thanks! 
I leave it up to you if you want to track the follow-up. :)

packages/language/src/workspace/lifecycle.ts
msujew: I've created https://github.com/zowe/zowe-pli-language-support/issues/468 for this.

PR #469, opened 2025-11-10
Compiler options refactorings

ssmifi:
Concludes https://github.com/zowe/zowe-pli-language-support/issues/442

- Mostly added tests and checked the old compiler options. 
- Moved remaining tests to fourslash.
- Removed old compiler options tests.


msujew (approved):
Thanks, looks good to me 👍

PR #470, opened 2025-11-10
[Feature] Implement caching for hover

Wagner-Laranjeiras:
Issue #450

**Summary**
This PR introduces an in-memory caching layer for include file content previews, improving performance and reducing redundant file reads. The implementation uses an LRUCache with a five-minute TTL to store truncated file content for hover previews and related operations.

**Key Changes**

- Added an LRU cache (includeCache) for storing file preview content.
- Implemented a new utility function getFileContentPreview() that:
- Returns cached preview content if available.
- Reads the first N lines of a file (default: 100) when not cached.
- Automatically stores new entries in the cache.
- Exposed a limited API (includeCacheApi) with only the necessary read/write operations (get, set, has, clear), rather than exporting the entire cache object.
- Added constants for TTL (FIVE_MINUTES) and DEFAULT_LINE_CUTOFF for easier configuration and testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,5 +33,8 @@
     "overrides": {
       "isexe": "3.1.1"
     }
+  },
+  "dependencies": {
+    "lru-cache": "^11.2.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -33,8 +33,5 @@
     "overrides": {
       "isexe": "3.1.1"
     }
-  },
-  "dependencies": {
-    "lru-cache": "^11.2.2"
   }
 }
--- a/packages/language/src/language-server/cache/include-cache.ts
+++ b/packages/language/src/language-server/cache/include-cache.ts
@@ -21,7 +21,7 @@ export function getFileContentPreview(
 ) {
   const partialContent = unit.services.includeCache.get(key);
   if (partialContent) {
-    return partialContent.toString();
+    return partialContent;
   }
   // load up the first (LINE_CUTOFF) lines of content from the file (semi-arbitrary cutoff)
   const fileContent = document.getText({
--- a/packages/language/src/language-server/cache/include-cache.ts
+++ b/packages/language/src/language-server/cache/include-cache.ts
@@ -1,0 +1,38 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+import { TextDocument } from "vscode-languageserver-textdocument";
+import { CompilationUnit } from "../../workspace/compilation-unit";
+
+const DEFAULT_LINE_CUTOFF = 100;
+
+export function getFileContentPreview(
+  unit: CompilationUnit,
+  key: string,
+  document: TextDocument,
+  lineCutoff = DEFAULT_LINE_CUTOFF,
+) {
+  const partialContent = unit.services.includeCache.get(key);
+  if (partialContent) {
+    return partialContent.toString();
+  }
+  // load up the first (LINE_CUTOFF) lines of content from the file (semi-arbitrary cutoff)
+  const fileContent = document.getText({
+    start: { line: 0, character: 0 },
+    end: { line: lineCutoff, character: 0 },
+  });
+  const partialContentText =
+    document.getText().length > fileContent.length
+      ? fileContent + "\n...\n"
+      : fileContent;
+
+  unit.services.includeCache.set(key, partialContentText);
+  return partialContentText;
+}
--- a/packages/language/src/language-server/hover-request.ts
+++ b/packages/language/src/language-server/hover-request.ts
@@ -37,6 +37,7 @@ import { URI } from "../utils/uri";
 import { retrieveProcedureFromLabelPrefix } from "../validation/utils";
 import { CompilationUnit } from "../workspace/compilation-unit";
 import { HoverResponse, tokenToRange } from "./types";
+import { getFileContentPreview } from "./cache/include-cache";
 
 type MarkupResponse = string | null;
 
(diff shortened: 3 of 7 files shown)


packages/language/src/language-server/cache/include-cache.ts
```diff
@@ -0,0 +1,51 @@
...
+  // load up the first (LINE_CUTOFF) lines of content from the file (semi-arbitrary cutoff)
+  const fileContent = document.getText({
+    start: { line: 0, character: 0 },
+    end: { line: lineCutoff, character: 0 },
+  });
+  const lineCount = fileContent.matchAll(/\n/g);
+  const partialContentText =
+    Array.from(lineCount).length > lineCutoff
+      ? fileContent + "\n...\n"
+      : fileContent;
```
msujew: Suggestion: This is a rather inefficient way of testing the length of the `fileContent`:
```suggestion
  const partialContentText =
    document.getText().length > fileContent.length
      ? fileContent + "\n...\n"
      : fileContent;
```

packages/language/test/lsp/include-cache.test.ts
```diff
@@ -0,0 +1,155 @@
...
+import { URI } from "../../src/utils/uri";
+import {
+  getFileContentPreview,
+  includeCacheApi,
+} from "../../src/language-server/cache/include-cache";
+
+const mockCode = `PTASK38: PROC OPTIONS(MAIN) REORDER ;
+ DCL SYSPRINT FILE PRINT STREAM OUTPUT   ;/* FILE SYSPRINT          */
+ DCL REP3800  FILE OUTPUT RECORD         ;/* FILE REP3800           */
+ DCL VSR404   FILE RECORD SEQUENTIAL INPUT ENV(VSAM RECSIZE(64));
```
msujew: Suggestion: Instead of using some of our samples, we should generate some synthetic data instead:

```
DCL Var1 FIXED;
DCL Var2 FIXED;
DCL Var3 FIXED;
DCL Var4 FIXED;
...
```

This makes the test easier to maintain in the long run.

packages/language/test/lsp/include-cache.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      150,
+    );
+    expect(includeCacheApi.get(`${mockPath}v02`)).toBeDefined();
+
+    expect(resultOne).toBe(mockCode);
+    expect(resultTwo).toBe(mockCode);
+  });
+
+  test("Returns partial content when not cached / when file has more than specified lines", () => {
+    includeCacheApi.clear();
```
msujew: Suggestion: The `includeCacheApi` clear should be part of a `afterEach`.

packages/language/src/language-server/cache/include-cache.ts
```diff
@@ -0,0 +1,51 @@
...
+const includeCache = new LRUCache({
+  max: 500,
+  ttl: FIVE_MINUTES,
+});
+
+export const includeCacheApi = {
+  get: includeCache.get.bind(includeCache),
+  set: includeCache.set.bind(includeCache),
+  has: includeCache.has.bind(includeCache),
+  clear: includeCache.clear.bind(includeCache),
```
msujew: Suggestion: The fourslash test framework should clear the cache after every test.

packages/language/src/language-server/cache/include-cache.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { LRUCache } from "lru-cache";
+import { TextDocument } from "vscode-languageserver-textdocument";
+
+const FIVE_MINUTES = 1000 * 60 * 5;
+const DEFAULT_LINE_CUTOFF = 100;
+const includeCache = new LRUCache({
+  max: 500,
+  ttl: FIVE_MINUTES,
+});
```
msujew: Suggestion: This should be a service on the `CompilationUnit`.

msujew (commented):
Looks quite good already. I have an architectural remark and then a few minor suggestions to improve the code.

msujew:
Thinking further about this - how does this deal with file changes in between cache accesses? If the file under the include has changed, does the cache entry get disposed? Otherwise, it could lead to outdated hover information.

packages/language/test/lsp/include-cache.test.ts
```diff
@@ -8,106 +8,120 @@
...
-  PUT SKIP(2) FILE (SYSPRINT) EDIT         /*                       */
-        ('***  NORMAL END OF PTASK38  ***') (A,A);
-  CLOSE FILE (REP3800)                ;    /*                       */
-  CLOSE FILE (VSR404)                ;     /*                       */
-  CLOSE FILE (SYSPRINT)               ;    /*                       */
-                                           /*                       */
- END PTASK38;`;
+const mockCode = `DCL Var1 FIXED;
+DCL Var2 FIXED;
+DCL Var3 FIXED;
```
msujew: Suggestion: Great, now we can generate this file programmatically in a loop :)

packages/language/src/language-server/hover-request.ts
```diff
@@ -253,7 +253,11 @@ function getIncludeItemRepresentation(
   if (!doc) {
     return null;
   }
-  const partialContent = getFileContentPreview(node.relativeFilePath, doc);
+  const partialContent = getFileContentPreview(
+    unit,
+    node.relativeFilePath,
+    doc,
```
msujew: Suggestion: Using `node.relativeFilePath` as the key here is relatively problematic. Imagine the following theoretical structure:

```
main.pli
test/file.pli
test/test/file.pli
```
With both `main.pli` and `test/file.pli` using something like a `%INCLUDE "./test/file.pli"` statement (relative includes are currently disabled, but that's beside the point). In both cases, they will include the relative file "./test/file.pli", but these probably have different contents. Therefore, the key is not unique for that specific file. We need something better like the URI of the file to ensure that we don't run into conflicts.

msujew (commented):
Much better, thank you 👍

packages/language/src/language-server/hover-request.ts
Wagner-Laranjeiras: Got it! Thanks

package.json
```diff
@@ -33,5 +33,8 @@
     "overrides": {
       "isexe": "3.1.1"
     }
+  },
+  "dependencies": {
+    "lru-cache": "^11.2.2"
```
msujew: Suggestion: This should be a dependency of the `language` package.

packages/language/src/language-server/cache/include-cache.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export function getFileContentPreview(
+  unit: CompilationUnit,
+  key: string,
+  document: TextDocument,
+  lineCutoff = DEFAULT_LINE_CUTOFF,
+) {
+  const partialContent = unit.services.includeCache.get(key);
+  if (partialContent) {
+    return partialContent.toString();
```
msujew: Question: Why `toString()`?

msujew (approved):
Looks good to me, just two points, see below.

packages/language/src/language-server/cache/include-cache.ts
Wagner-Laranjeiras: I just wanted to ensure we would get a string, but it's redundant because we are already enforcing this when setting the parameters (I still have some vanilla JS habits I'm working on).

Issue #441, opened 2025-10-14
Implement Macro Preprocessor Effects


ssmifi:
Implement effects for DEPRECATE, DEPRECATENEXT, NAMEPREFIX, (maybe ID, have to check), and diagnostics + quick fix for CASE(UPPER).

asatklichov:
@ssmifi  please provide a warning message for those below, not to be supported
 'DBCS(EXACT | INEXACT) - DBCS(INEXACT), EOLCOMM | NOEOLCOMM - EOLCOMM, INCONLY | NOINCONLY - NOINCONLY
'

And implement:  
**RESCAN(ASIS | UPPER) **

PR #471, opened 2025-11-12
Macro pp effects

ssmifi:
Implements https://github.com/zowe/zowe-pli-language-support/issues/441

* Implements macro pp validations for DEPRECATE, DEPRECATENEXT, NAMEPREFIX and CASE.
* Provides quickfix and source action for CASE.
* Adds a "not supported" message for DBCS, EOLCOMM, and INCONLY.

The Macro PP diagnostics use internal codes until we can verify the official codes on the mainframe. Afterwards they should be replaced with the official ones provided they carry the same information.

### DIFF
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -16,6 +16,7 @@ import {
   CodeAction,
   CodeActionKind,
   Diagnostic,
+  TextEdit,
 } from "vscode-languageserver-types";
 import {
   PluginConfigurationProviderInstance,
@@ -115,6 +116,31 @@ export async function quickFixCreateConfig(
   return action;
 }
 
+export function quickFixUppercaseText(
+  diagnostic: Diagnostic,
+): CodeAction | undefined {
+  const range = diagnostic.range;
+  if (!diagnostic.data || !diagnostic.data.text || !diagnostic.data.uri) {
+    console.error("Diagnostic data is missing for quick fix uppercase text.");
+    return undefined;
+  }
+  const text = diagnostic.data.text;
+  const uri = diagnostic.data.uri;
+
+  const action: CodeAction = {
+    title: `Convert to uppercase`,
+    kind: CodeActionKind.QuickFix,
+    diagnostics: [diagnostic],
+    edit: {
+      changes: {
+        [uri]: [TextEdit.replace(range, text.toUpperCase())],
+      },
+    },
+  };
+
+  return action;
+}
+
 export async function applyQuickFixes(
   diagnostics: Diagnostic[],
 ): Promise<CodeAction[] | undefined> {
@@ -124,6 +150,7 @@ export async function applyQuickFixes(
   const CODE_MISSING_CONFIG = fullCode(
     LspCodes.IncludeResolution.MissingConfiguration,
   ); // "Could not resolve include directive. Plugin configuration is missing"
+  const CODE_MACRO_CASE = fullCode(LspCodes.UpperCase);
 
   for (const diagnostic of diagnostics) {
     if (!diagnostic.code) {
@@ -137,6 +164,10 @@ export async function applyQuickFixes(
         break;
       case CODE_MISSING_CONFIG:
         action = await quickFixCreateConfig(diagnostic);
+        if (action) actions.push(action);
+        break;
+      case CODE_MACRO_CASE:
+        action = quickFixUppercaseText(diagnostic);
         if (action) actions.push(action);
         break;
     }
--- a/packages/language/src/language-server/code-actions/apply-source-actions.ts
+++ b/packages/language/src/language-server/code-actions/apply-source-actions.ts
@@ -1,0 +1,106 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import {
+  CodeAction,
+  CodeActionKind,
+  Diagnostic,
+  TextEdit,
+} from "vscode-languageserver-types";
+import { URI } from "../../utils/uri";
+import { CompilationUnitHandler } from "../../workspace/compilation-unit";
+import { diagnosticToLSP, fullCode } from "../types";
+import { LspCodes } from "../../validation/lsp-codes";
+
+export function sourceActionUppercaseAllText(
+  caseDiagnostics: Diagnostic[],
+): CodeAction {
+  // Group diagnostics by URI
+  const diagnosticsByUri: { [uri: string]: Diagnostic[] } = {};
+
+  for (const diagnostic of caseDiagnostics) {
+    if (!diagnostic.data || !diagnostic.data.uri) {
+      console.error(
+        "Diagnostic data is missing for source action uppercase all text.",
+      );
+      continue;
+    }
+    const uri = diagnostic.data.uri;
+    if (!diagnosticsByUri[uri]) {
+      diagnosticsByUri[uri] = [];
+    }
+    diagnosticsByUri[uri].push(diagnostic);
+  }
+
+  // Create text edits for each file
+  const changes: { [uri: string]: TextEdit[] } = {};
+
+  for (const [uri, diagnostics] of Object.entries(diagnosticsByUri)) {
+    changes[uri] = diagnostics
+      .filter((d) => d.range && d.data?.text)
+      .map((diagnostic) =>
+        TextEdit.replace(
+          diagnostic.range!,
+          (diagnostic.data?.text || "").toUpperCase(),
+        ),
+      );
+  }
+
+  const action: CodeAction = {
+    title: `Convert all to uppercase (${caseDiagnostics.length} instances)`,
+    kind: CodeActionKind.SourceFixAll,
+    diagnostics: caseDiagnostics,
+    edit: {
+      changes,
+    },
+  };
+
+  return action;
+}
+
+export function getCaseDiagnosticsFromCompilationUnit(
+  uri: string,
+  compilationUnitHandler: CompilationUnitHandler,
+): Diagnostic[] {
+  try {
+    const unit = compilationUnitHandler.getCompilationUnit(URI.parse(uri));
+    if (!unit) return [];
+
+    const caseDiagnostics = unit.diagnostics
+      .getAll()
+      .filter((d) => d.code === fullCode(LspCodes.UpperCase));
+
+    return caseDiagnostics
+      .map((d) => diagnosticToLSP(unit, d))
+      .filter((d) => d !== undefined) as Diagnostic[];
+  } catch (error) {
+    return [];
+  }
+}
+
+export async function applySourceActions(
+  uri: string,
+  compilationUnitHandler: CompilationUnitHandler,
+): Promise<CodeAction[] | undefined> {
+  const actions: CodeAction[] = [];
+
+  const caseDiagnostics = getCaseDiagnosticsFromCompilationUnit(
+    uri,
+    compilationUnitHandler,
+  );
+  if (caseDiagnostics.length > 0) {
+    const fixAllAction = sourceActionUppercaseAllText(caseDiagnostics);
+    actions.push(fixAllAction);
+  }
+
+  if (!actions.length) return undefined;
+  return actions;
+}
--- a/packages/language/src/preprocessor/compiler-options/options-macro.ts
+++ b/packages/language/src/preprocessor/compiler-options/options-macro.ts
@@ -12,8 +12,8 @@
 export interface CompilerOptions {
   case?: CompilerOptions.Case;
   dbcs?: CompilerOptions.Dbcs;
-  deprecate?: string[];
-  deprecateNext?: string[];
+  deprecate?: Set<string>;
+  deprecateNext?: Set<string>;
   eolComm?: boolean;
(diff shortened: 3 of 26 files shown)


packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -90,17 +93,40 @@ export async function quickFixResolveInclude(
   return action;
 }
 
+export function quickFixUppercaseText(diagnostic: Diagnostic): CodeAction {
+  const range = diagnostic.range;
+  const text = diagnostic.data?.text || "";
+  const uri = diagnostic.data?.uri || "";
```
msujew: Suggestion: IMO, if no text or uri is specified, we should just return with `undefined` (and maybe log an error?).

packages/language/src/language-server/code-actions/apply-source-actions.ts
```diff
@@ -0,0 +1,100 @@
...
+import { MacroCodes } from "../../validation/internal-codes";
+
+export function sourceActionUppercaseAllText(
+  caseDiagnostics: Diagnostic[],
+): CodeAction {
+  // Group diagnostics by URI
+  const diagnosticsByUri: { [uri: string]: Diagnostic[] } = {};
+
+  for (const diagnostic of caseDiagnostics) {
+    const uri = diagnostic.data?.uri || "";
```
msujew: Same here.

packages/language/src/preprocessor/compiler-options/options-macro.ts
```diff
@@ -41,8 +41,8 @@ export namespace CompilerOptions {
 const defaultCompilerOptions: CompilerOptions = {
   case: "ASIS",
   dbcs: "INEXACT",
-  deprecate: [],
-  deprecateNext: [],
+  deprecate: new Set([]),
```
msujew:
```suggestion
  deprecate: new Set(),
```

packages/language/src/syntax-tree/ast-iterator.ts
```diff
@@ -1046,3 +1047,31 @@ export function collectNodesOfKind<T extends SyntaxNode>(
...
+}
+
+function extractDirectTokens(node: SyntaxNode, tokens: Token[]): void {
+  const nodeRecord = node as unknown as Record<string, unknown>;
+  for (const value of Object.values(nodeRecord)) {
+    if (isToken(value)) {
+      tokens.push(value);
+    }
+  }
+}
```
msujew: Question: Am I correct in the assumption that this should extract only the tokens that belong to non-pp code? What about lower-cased pp code?

packages/language/src/validation/macro/case.ts
```diff
@@ -0,0 +1,44 @@
...
+  acceptor: ValidationAcceptor,
+  compilationUnit: CompilationUnit,
+) {
+  if (node !== compilationUnit.preprocessorAst) return;
+  if (compilationUnit.compilerOptions.macroOptions.case !== "UPPER") return;
+
+  const tokens = collectAllTokens(node);
+
+  for (const token of tokens) {
+    if (/[a-z]/.test(token.originalImage)) {
```
msujew, line 39: Question: Doesn't this test also match string tokens? AFAIK strings always retain their original casing in `UPPER` mode.

packages/language/src/validation/macro/deprecate.ts
```diff
@@ -0,0 +1,63 @@
...
+    !compilationUnit.compilerOptions.macroOptions.deprecate &&
+    !compilationUnit.compilerOptions.macroOptions.deprecateNext
+  ) {
+    return;
+  }
+
+  const nameToken =
+    node.container?.kind === AST.SyntaxKind.Statement
+      ? node.container.labels[0]?.nameToken
+      : undefined;
```
msujew: Suggestion: I believe we actually need to look at all procedure labels here:

```
*PROCESS(DEPRECATE(X2))
 X1: X2: PROC...
```

packages/language/src/validation/macro/nameprefix.ts
```diff
@@ -0,0 +1,56 @@
...
+  let name;
+
+  if (node.kind === AST.SyntaxKind.DeclaredVariable) {
+    nameToken = node.nameToken;
+    name = node.name;
+  } else {
+    nameToken =
+      node.container?.kind === AST.SyntaxKind.Statement
+        ? node.container.labels[0]?.nameToken
+        : undefined;
```
msujew: Same here.

packages/language/src/validation/macro/nameprefix.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+import { diagnosticFromCode } from "../../language-server/types";
+import * as AST from "../../syntax-tree/ast";
+import { ValidationAcceptor } from "../validator";
+import { CompilationUnit } from "../../workspace/compilation-unit";
+import { MacroCodes } from "../internal-codes";
+
+export function MACRO_NamePrefix(
+  node: AST.ProcedureStatement | AST.DeclaredVariable,
```
msujew, line 20: Question: Does the `NAMEPREFIX` option also affect ordinal types and aliases?

packages/language/test/lsp/apply-quick-fixes.test.ts
```diff
@@ -137,3 +137,72 @@ describe("applyQuickFixes", () => {
     expect(result).toBeUndefined();
   });
 });
+
+describe("applyQuickFixes with case diagnostics", () => {
```
msujew, line 236: Question: I'm generally fine with testing quick fixes directly. However, do you think there might be a good way to encapsulate these test inside of the fourslash framework?

packages/language/src/validation/macro/case.ts
ssmifi: Good point. I excluded string tokens and added a line to the test to check that strings stay untouched.

packages/language/src/validation/macro/deprecate.ts
ssmifi: Added another test case for this.

packages/language/src/validation/macro/nameprefix.ts
ssmifi: 👍 + added test case.

packages/language/src/validation/macro/nameprefix.ts
ssmifi: Unfortunately, our mainframe does not give any diagnostics for this (also not for the cases that are specified in the spec). 
The spec states 
>The NAMEPREFIX option specifies that the names of preprocessor procedures and variables must start 
with the specified character.

Should we keep this _as is_ for now until we receive further information?

packages/language/test/lsp/apply-quick-fixes.test.ts
ssmifi: As discussed, would be nice, but need to think about it some more. Opened https://github.com/zowe/zowe-pli-language-support/issues/484 to track this.

packages/language/src/syntax-tree/ast-iterator.ts
ssmifi: I refactored this (and removed it from the `ast-iterator`), because we will not get all tokens this way. The case validation now implements a special case to check all tokens if they belong to the preprocessor, which maybe needs to be optimized for large files, but only runs if `CASE(UPPER)` is set. Wdyt?

packages/language/src/validation/macro/nameprefix.ts
msujew: Sure, let's not over complicate thing 👍

packages/language/src/validation/macro/case.ts
```diff
@@ -0,0 +1,71 @@
...
+        text: token.originalImage,
+      };
+      acceptor(diagnostic);
+    }
+  }
+}
+
+function collectPreprocessorTokens(unit: CompilationUnit): Token[] {
+  const tokens: Token[] = [];
+  const tokenMap = new Map<AST.SyntaxNode, Token>();
```
msujew: Suggestion: Using a `Map` only allows a single token per syntax node. We should use a `MultiMap` instead.

packages/language/src/validation/macro/deprecate.ts
```diff
@@ -0,0 +1,66 @@
...
+    ) {
+      acceptor(
+        diagnosticFromCode(
+          {
+            ...MacroCodes.DeprecateProcedure,
+            severity: Severity.W,
+          },
+          nameToken,
+          name,
+        ),
```
msujew: Suggestion: I believe the correct diagnostic for this is `Warning.IBM3334I`.

packages/language/src/validation/macro/deprecate.ts
```diff
@@ -0,0 +1,66 @@
...
+  for (const nameToken of nameTokens) {
+    const name = nameToken.image;
+
+    if (
+      compilationUnit.compilerOptions.macroOptions.deprecate &&
+      compilationUnit.compilerOptions.macroOptions.deprecate.has(name)
+    ) {
+      acceptor(
+        diagnosticFromCode(MacroCodes.DeprecateProcedure, nameToken, name),
+      );
```
msujew: Suggestion: I believe the correct diagnostic for this is `Error.IBM3660I`.

packages/language/src/validation/macro/deprecate.ts
```diff
@@ -0,0 +1,66 @@
...
+export function MACRO_Deprecate(
+  node: AST.ProcedureStatement,
+  acceptor: ValidationAcceptor,
+  compilationUnit: CompilationUnit,
+) {
+  if (
+    !compilationUnit.compilerOptions.macroOptions.deprecate &&
+    !compilationUnit.compilerOptions.macroOptions.deprecateNext
+  ) {
+    return;
```
msujew, line 27: Suggestion: Since the default is a non-null `Set` anyway, shouldn't we test for the size of the sets instead?

packages/language/src/validation/macro/nameprefix.ts
```diff
@@ -0,0 +1,58 @@
...
+      !nameToken ||
+      !name ||
+      name.startsWith(
+        compilationUnit.compilerOptions.macroOptions.namePrefix.character,
+      )
+    ) {
+      continue;
+    }
+
+    acceptor(diagnosticFromCode(MacroCodes.NamePrefix, nameToken, name));
```
msujew: Suggestion: I believe the correct code would be `Error.IBM3518I`.

packages/language/src/workspace/file-store.ts
```diff
@@ -31,6 +31,14 @@ export class FileStore {
     return this.get(uri)?.tokens;
   }
 
+  getAllTokens(): Token[] {
+    const allTokens: Token[] = [];
+    for (const file of this.map.values()) {
+      allTokens.push(...file.tokens);
+    }
+    return allTokens;
+  }
```
msujew: Suggestion: Since we iterate over this anyway, can we use a generator function? Otherwise we probably generate a lot of unnecessary arrays.
```suggestion
  *getAllTokens(): IterableIterator<Token> {
    for (const file of this.map.values()) {
      yield* file.tokens;
    }
  }
```

packages/language/src/validation/macro/case.ts
```diff
@@ -0,0 +1,71 @@
...
+} from "../../syntax-tree/ast-iterator";
+import { PreprocessorTokens } from "../../preprocessor/pli-preprocessor-tokens";
+
+export function MACRO_Case(
+  node: AST.Program,
+  acceptor: ValidationAcceptor,
+  compilationUnit: CompilationUnit,
+) {
+  if (node !== compilationUnit.preprocessorAst) return;
+  if (compilationUnit.compilerOptions.macroOptions.case !== "UPPER") return;
```
msujew, line 31: Maybe a sort of TODO: I would like to have this behind a toggle/feature flag. I can imagine this impacting performance quite noticeably on large projects.

packages/language/src/validation/macro/deprecate.ts
ssmifi: Technically, the compiler options are still optional, but I exit early now, if both sets are empty.

packages/language/src/validation/macro/case.ts
ssmifi: Sounds good to me. Added https://github.com/zowe/zowe-pli-language-support/issues/490

packages/language/src/validation/internal-codes.ts
```diff
@@ -21,6 +21,14 @@ export const InternalCodes = {
   } as ParametricPLICode,
 };
 
+export const MacroCodes = {
+  Case: {
+    code: "MACRO0001",
+    severity: Severity.W,
+    message: (token: any) => `Input text "${token}" must be uppercase.`,
+  } as ParametricPLICode,
+};
```
msujew: Suggestion: Can we move this somewhere else? The internal codes are AFAIK only those used for internal testing purposes.

msujew (approved):
Looks good to me, just one suggestion, see below.

packages/language/src/validation/internal-codes.ts
ssmifi: 👍 Moved to `lsp-codes`.

packages/language/src/validation/lsp-codes.ts
```diff
@@ -21,4 +21,10 @@ export const LspCodes = {
         "Could not resolve include directive. Plugin configuration is missing",
     } as SimplePLICode,
   },
+
+  UpperCase: {
+    code: "LSPUC001",
+    severity: Severity.W,
+    message: (token: any) => `Input text "${token}" must be uppercase.`,
+  } as ParametricPLICode,
```
msujew, line 29: Suggestion: Casting will make the `Parameter<T["message"]>` type behave incorrectly.

```suggestion
    message: (token: string) => `Input text "${token}" must be uppercase.`,
  },
```

PR #473, opened 2025-11-14
Add validation for scan attribute repetition

### LINKED ISSUE #431: preprocessor-findings 9.10

ssmifi:
Fixes https://github.com/zowe/zowe-pli-language-support/issues/431 3

The type system changes already covers repeating attributes that influence the type. (Note that the mainframe is actually fine with the repetition.)
This validation adds a check for the scan attribute.

### DIFF
--- a/packages/language/src/validation/compiler/IBM1352IE-declare-item-scan-repetition.ts
+++ b/packages/language/src/validation/compiler/IBM1352IE-declare-item-scan-repetition.ts
@@ -1,0 +1,60 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { diagnosticFromCode } from "../../language-server/types";
+import {
+  ComputationDataAttribute,
+  DeclaredItem,
+  SyntaxKind,
+  DefaultAttribute,
+} from "../../syntax-tree/ast";
+import { PLICodes } from "../pli-codes";
+import { ValidationAcceptor } from "../validator";
+
+// %dcl A(4) FIXED INTERNAL SCAN RESCAN; // throws warning IBM3252I with RESCAN.
+// dcl A FIXED SCAN RESCAN; // throws generic IBM1352I with SCAN.
+
+const scanAttributes = (node: DeclaredItem): ComputationDataAttribute[] =>
+  node.attributes.filter(
+    (attr) =>
+      attr.kind === SyntaxKind.ComputationDataAttribute &&
+      (attr.type === DefaultAttribute.SCAN ||
+        attr.type === DefaultAttribute.RESCAN ||
+        attr.type === DefaultAttribute.NOSCAN),
+  ) as ComputationDataAttribute[];
+
+export function IBM1352IE_declared_item_pp_scan_repetition(
+  node: DeclaredItem,
+  accept: ValidationAcceptor,
+): void {
+  let hasScanAttribute = false;
+
+  scanAttributes(node).forEach((attr) => {
+    if (hasScanAttribute) {
+      const token = attr.typeToken;
+      if (!token) return;
+      accept(diagnosticFromCode(PLICodes.Warning.IBM3252I, token, token.image));
+    } else {
+      hasScanAttribute = true;
+    }
+  });
+}
+
+export function IBM1352IE_declared_item_pli_scan_repetition(
+  node: DeclaredItem,
+  accept: ValidationAcceptor,
+): void {
+  scanAttributes(node).forEach((attr) => {
+    const token = (attr as ComputationDataAttribute).typeToken;
+    if (!token) return;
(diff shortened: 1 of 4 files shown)


msujew (approved):

packages/language/src/validation/pli-validator.ts
```diff
@@ -31,21 +32,22 @@ import { IBM3323I_IBM3324I_check_argument_count } from "./compiler/IBM3323I-IBM3
  */
 export function registerPliValidationChecks(): ValidationChecks {
   return {
+    CallStatement: [IBM3323I_IBM3324I_check_argument_count],
+    DeclaredItem: [IBM1352IE_declared_item_scan_repetition],
```
msujew: Nevermind, maybe my approval was a bit too fast. In fact, on normal PLI variables, the scan related variables aren't allowed, since they are only valid for the preprocessor. How does the mainframe react to `SCAN` on a normal `DCL`?

ssmifi: 👍 I will verify this.

ssmifi:
Differentiate between IBM3252 for conflicting scan attribute in preprocessor and IBM1352 for invalid statement element in pli now. Please recheck.

packages/language/src/validation/compiler/IBM1352IE-declare-item-scan-repetition.ts
```diff
@@ -0,0 +1,59 @@
...
+
+// %dcl A(4) FIXED INTERNAL SCAN RESCAN; // throws warning IBM3252I with RESCAN.
+// dcl A FIXED SCAN RESCAN; // throws generic IBM1352I with SCAN.
+
+const scanAttributes = (node: DeclaredItem) =>
+  node.attributes
+    .filter((attr) => attr.kind === SyntaxKind.ComputationDataAttribute)
+    .map((attr) => attr as ComputationDataAttribute)
+    .filter((attr) => typeof attr.type === "string" && attr.type !== null)
+    .filter((attr) => scanModeValues.includes(attr.type! as ScanMode));
```
msujew: Suggestion: I've just noticed that this results in 4 separate iterations on the `attributes` array. Can we merge this into one and just use `as ComputationDataAttribute[]` at the end?

packages/language/src/validation/compiler/IBM1352IE-declare-item-scan-repetition.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export function IBM1352IE_declared_item_pp_scan_repetition(
+  node: DeclaredItem,
+  accept: ValidationAcceptor,
+): void {
+  let hasScanAttribute = false;
+
+  scanAttributes(node).forEach((attr) => {
+    if (hasScanAttribute) {
+      const token = (attr as ComputationDataAttribute).typeToken;
```
msujew: Minor: I think we can do without the casting, see above. The same applies to the other validation function.

packages/language/src/syntax-tree/ast.ts
```diff
@@ -565,7 +565,8 @@ export type CommonDeclarationAttribute =
  * Type extending attributes are attributes that extend the type of a variable.
  */
 export type TypeExtendingAttribute = LikeAttribute | TypeAttribute;
-export type ScanMode = "SCAN" | "RESCAN" | "NOSCAN";
+export const scanModeValues = ["SCAN", "RESCAN", "NOSCAN"] as const;
+export type ScanMode = (typeof scanModeValues)[number];
```
msujew: After the latest rebase, this should work a bit differently.

msujew (approved):
LGTM

PR #474, opened 2025-11-17
Type system: Add more attributes

Lotes:
I added some of the missing attributes. Especially DIMENSION and INITIAL.

There were also some discussion, how much information we should infer into the type descriptions. The result was, that we can add functionality later, if needed. It makes currently no sense to compute everything when we do not use it for now.

There are also a bunch of attributes that are not listed in the online documentation, but instead mentioned in syntax diagrams or inside th index of the language reference.

I was able to identify some extra attributes for parameter passing and a handful of flag attributes. I will stop adding attributes after this PR. We should add them when they are required. Some attributes are also very complex. I tended to store their AST node into the type description FTM.

### DIFF
--- a/packages/language/src/syntax-tree/ast.ts
+++ b/packages/language/src/syntax-tree/ast.ts
@@ -1799,6 +1799,7 @@ export interface InitialAttribute extends AstNode {
   procedureCall: ProcedureCall | null;
   to: boolean;
   content: InitialToContent | null;
+  token: Token | null;
 }
 export interface InitialAttributeItemStar extends AstNode {
   kind: SyntaxKind.InitialAttributeItemStar;
--- a/packages/language/src/syntax-tree/ast.ts
+++ b/packages/language/src/syntax-tree/ast.ts
@@ -1297,25 +1297,25 @@ export function createDimensionBound(): DimensionBound {
 export interface Dimensions extends AstNode {
   kind: SyntaxKind.Dimensions;
   dimensions: DimensionBound[];
+  token: Token | null;
 }
 export function createDimensions(): Dimensions {
   return {
     kind: SyntaxKind.Dimensions,
     container: null,
     dimensions: [],
+    token: null,
   };
 }
 export interface DimensionsDataAttribute extends AstNode {
   kind: SyntaxKind.DimensionsDataAttribute;
   dimensions: Dimensions | null;
-  dimensionsToken: Token | null;
 }
 export function createDimensionsDataAttribute(): DimensionsDataAttribute {
   return {
     kind: SyntaxKind.DimensionsDataAttribute,
     container: null,
     dimensions: null,
-    dimensionsToken: null,
   };
 }
 export interface DisplayStatement extends AstNode {
--- a/packages/language/src/syntax-tree/ast.ts
+++ b/packages/language/src/syntax-tree/ast.ts
@@ -1308,12 +1308,14 @@ export function createDimensions(): Dimensions {
 export interface DimensionsDataAttribute extends AstNode {
   kind: SyntaxKind.DimensionsDataAttribute;
   dimensions: Dimensions | null;
+  dimensionsToken: Token | null;
 }
 export function createDimensionsDataAttribute(): DimensionsDataAttribute {
   return {
     kind: SyntaxKind.DimensionsDataAttribute,
     container: null,
     dimensions: null,
+    dimensionsToken: null,
   };
 }
 export interface DisplayStatement extends AstNode {
--- a/packages/language/src/typesystem/computed-attributes.ts
+++ b/packages/language/src/typesystem/computed-attributes.ts
@@ -2,11 +2,12 @@ import { Bound, DimensionBound } from "./descriptions";
 import * as ast from "../syntax-tree/ast";
 import { evaluateExpression } from "./evaluate";
 
+/** @see https://www.ibm.com/docs/en/epfz/6.1.0?topic=arrays-dimension-attribute */
 export function computeDimensions(dimension: ast.Dimensions): Array<DimensionBound> {
     const dims = dimension.dimensions;
     const result: Array<DimensionBound> = [];
     for(const dim of dims) {
-        function computeExpression(bound: ast.Bound | undefined | null, defaultValue: number): Bound {
+        function computeExpression(bound: ast.Bound | undefined | null, defaultValue: number|undefined): Bound {
             const expr = bound?.expression;
             if (!expr) {
                 return {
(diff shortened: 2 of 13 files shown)


packages/language/src/typesystem/computed-attributes.ts
```diff
@@ -0,0 +1,31 @@
+import { DimensionBound } from "../../test/fourslash-harness/harness-interface";
```
msujew: I don't think you intended to import this from the fourslash interface, did you? :D

packages/language/src/typesystem/computed-attributes.ts
```diff
@@ -0,0 +1,31 @@
...
+import { DataType, TypeDescriptions } from "./descriptions";
+import * as ast from "../syntax-tree/ast";
+import { evaluateExpression } from "./evaluate";
+import { assertType } from "../preprocessor/util";
+
+export function computeDimensions(description: TypeDescriptions.Any): Array<DimensionBound> | undefined {
```
msujew: Suggestion: I would argue that's something we should immediately compute during the type inference. Otherwise, I would assume that we need to compute this multiple times when doing work like testing assignability during the validation. Am I right with that? Or is there a benefit of computing the bounds after the type inference?

packages/language/src/syntax-tree/ast.ts
```diff
@@ -1308,12 +1273,14 @@ export function createDimensions(): Dimensions {
...
   kind: SyntaxKind.DimensionsDataAttribute;
   dimensions: Dimensions | null;
+  dimensionsToken: Token | null;
 }
 export function createDimensionsDataAttribute(): DimensionsDataAttribute {
   return {
     kind: SyntaxKind.DimensionsDataAttribute,
     container: null,
     dimensions: null,
+    dimensionsToken: null,
```
msujew: Question: Is there a specific reason to store the token on the attribute itself? Otherwise, I would recommend to store it on the `Dimensions` ast node, which would also make the parsing less complicated.

packages/language/src/typesystem/computed-attributes.ts
Lotes: Good catch.

packages/language/src/syntax-tree/ast.ts
Lotes: True!

packages/language/src/typesystem/computed-attributes.ts
Lotes: The reason is not visible, yet, true. For dimensions, we could also rely on the `INITIAL` attribute. Then, when `DIMENSION(*)` and `INTIAL(1,2,3)` are given, the `DIMENSION` could be resolved as `DIMENSION(1:3)`. But maybe I think too complicated and keep `DIMENSION(*)` and `INTIAL(1,2,3)` separated.

Lotes: Discussion summary: Let's keep it simple, we can compute constants and store them additionally in the type as value or undefined, if not constant.

packages/language/src/typesystem/computed-attributes.ts
```diff
@@ -0,0 +1,46 @@
...
+                return {
+                    value: undefined,
+                    expression: expr,
+                    refersTo: null,
+                };
+            }
+        }
+        result.push({
+            lowerBound: computeExpression(dim.lower, 1),
+            upperBound: computeExpression(dim.upper, 1),
```
msujew: Suggestion: Shouldn't this be undefined? IMO it should not be able to perform a comparison at all later on if no expression is set (i.e. a parser error occurs). Otherwise users get downstream errors from issues that already have a diagnostic attached to them.

Lotes: Good question! The [documentation](https://www.ibm.com/docs/en/epfz/6.1.0?topic=arrays-dimension-attribute) says at least about the lower bound: `If only the upper bound is given, the lower bound defaults to 1.`. For the upper bound, it should not matter, since I think it is always defined, as an expression or `*`. Let me check if I find an example with an `undefined` upper bound...

Lotes: I have not thought about errors, true....

msujew: Yep, `1` as the default lower bound makes perfect sense. My main concern is that we should put `undefined` as the upper bound in case we run into a parser error.

packages/language/src/typesystem/evaluate.ts
```diff
@@ -0,0 +1,64 @@
...
+        switch (expression.op) {
+          case "+":
+          case "-":
+            assertType<DataType.Arithmetic>(operand.type);
+            return {
+              type: operand.type,
+              value:
+                expression.op === "+"
+                  ? +(operand.value as number)
+                  : -(operand.value as number),
```
msujew: Question: Can we really assert that? What about `DCL X(-"Y")`?

msujew (commented):
Looks quite good, just one question below.

packages/language/src/typesystem/evaluate.ts
Lotes: No, not really. Changed. Thanks.

packages/language/src/typesystem/primitive-type-builder.ts
```diff
@@ -87,7 +91,23 @@ export class DefaultPrimitiveTypeBuilder implements PrimitiveTypeBuilder {
       case ast.SyntaxKind.DateAttribute:
       case ast.SyntaxKind.DefinedAttribute:
         break;
+      case ast.SyntaxKind.InitialAttribute:
+        this.addAttributeWitness(
+          AttributeKind.Initial,
+          attribute,
+          attribute,
+          attribute.token!,
```
msujew: Suggestion: Please don't use `!` in this codebase anywhere (this applies to all `!` used in this PR). Things are typed as `null` for a reason. Tests are essentially incapable of catching these null-reference errors, because they only come up during typing in the editor.

msujew (approved):
Aside from the issue above, this looks good to me 👍

packages/language/src/typesystem/primitive-type-builder.ts
msujew: I'm reviewing the type system code now again. There seems to be quite a few `!` used in there. We should aim to keep the code type safe. We can't assume that any of the code that passes through the type system is syntactically (or even semantically) correct.

Lotes: Thanks, I will take care (probably later this day). Sorry, that I did it again... Need to look at each occurrence...
Is it possible to add a linting rule + an explicit by-pass mechanism for cases where we can use `!`?

PR #476, opened 2025-11-18
Add pp rescan handling

ssmifi:
Concludes https://github.com/zowe/zowe-pli-language-support/issues/441 and hence https://github.com/zowe/zowe-pli-language-support/issues/397

Takes RESCAN(UPPER) as default and handles ASIS as the exception. Please let us discuss if you would like to approach this differently.

The examples are taken from the programming guide.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -1952,7 +1952,10 @@ function replaceTokenWithValue(
   immediateFollow: boolean,
   context: InterpreterContext,
 ): Token[] {
-  const tokens = lex(value);
+  const tokens = lex(
+    value,
+    context.unit.compilerOptions.macroOptions?.rescan !== "ASIS",
+  );
   setImmediateFollowProperty(immediateFollow, tokens);
   return replaceTokensInText(tokens, context);
 }
@@ -1974,8 +1977,8 @@ function generateSyntheticRefItem(
   return refItem;
 }
 
-function lex(text: string): Token[] {
-  return tokenize(text, undefined).tokens;
+function lex(text: string, caseUpper: boolean = true): Token[] {
+  return tokenize(text, undefined, caseUpper).tokens;
 }
 
 function setImmediateFollowProperty(
--- a/packages/language/test/fourslash/preprocessor/rescan-upper.ts
+++ b/packages/language/test/fourslash/preprocessor/rescan-upper.ts
@@ -1,0 +1,31 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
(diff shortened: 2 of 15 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1903,7 +1904,24 @@ function replaceTokenWithValue(
...
+      tokens[0].tokenType === ID &&
+      tokens[0].image === originalText.toUpperCase()
+    ) {
+      const caseToken = {
+        ...tokens[0],
+        image: originalText,
+      };
+      tokens = [caseToken];
+    }
+  }
```
msujew: Suggestion: This looks a bit hacky, especially because it can exclusively operate on single tokens (the values don't necessarily have to be just one token). I have a better solution that covers both the `CASE` and `PP(RESCAN)` compiler options:

1. The `tokenize` function should accept an "upper: boolean" flag that upper cases the input if set. See [here](https://github.com/zowe/zowe-pli-language-support/blob/6e5b0321d138a201f9a269ed117167060b137608/packages/language/src/parser/tokenizer.ts#L200-L203) and [here](https://github.com/zowe/zowe-pli-language-support/blob/6e5b0321d138a201f9a269ed117167060b137608/packages/language/src/parser/tokenizer.ts#L209) for the lines that need to be changed.
2. The `Lexer` class will use the `CASE` compiler option to pass the respective value into it.
3. The `lex` function in this file will use the `PP(RESCAN)` value to compute the flag.

What do you think of that?

ssmifi: Thanks. That was actually my initial approach, but I was thrown off by the PG example, which led me to believe identifiers were always uppercase and opted for a more isolated solution. I’m fine with your proposal; however, we should review the rescan discrepancy between the spec and the mainframe tomorrow.

ssmifi:
The tokenizer now accepts a `caseUpper` flag. The defaults for our implementation are still UPPER. 
Additionally, as suggested in https://github.com/zowe/zowe-pli-language-support/pull/460 closing https://github.com/zowe/zowe-pli-language-support/issues/490, the case(UPPER) validation is now opt-in via the `lspOptions` in the process group. The diagnostics produced by the validation are also capped.

packages/language/test/fourslash/preprocessor/rescan-upper.ts
```diff
@@ -0,0 +1,29 @@
...
+////
+//// %eins = 'zwei';
+//// %text = 'EINS';
+//// display( text );
+////
+//// %text = 'eins';
+//// display( text );
+
+preprocessor.expectTokens(` display( zwei ); 
+  display( zwei ); `);
```
msujew: Minor: Formatting
```suggestion
preprocessor.expectTokens(`
  display(zwei); 
  display(zwei);
`);
```

msujew (approved):
Ok, looks good to me 👍

PR #477, opened 2025-11-18
Stabilize PLIPlugin Initialization in 4Slash Tests

### LINKED ISSUE #475: Add PLIPluginConfig Support for Testing
Note to add built-in config support for our exising 4slash tests. Currently we have to do some tricks to force the config to reload in various cases, which is a bit clunky.

montymxb:
Closes #475 by ensuring the plugin configuration is initialized after we write all test files, and before we parse anything. This resolves a previous issue where the config was setup separately in a beforeEach, which being async sometimes ran before or after we separately wrote files to our VFS. This was especially noticeable when debugging, which would cause a few of these tests to then fail.

Just a sidenote, the original solution in mind was to allow `async` usage in our 4slash tests, but I opted for this way for now, as it fixed the issue more directly. But it wasn't _too_ bad to lift the commands under test into an async context, that worked pretty well. That might still be a helpful change depending on test requirements later on.

The current setup looks to setup the existing default config only when we don't have a user-specified alternative as a file under test. It also modifies the logic in `setProgramConfig` to properly post-process entries, which prevents issues where an already prepared process group config wouldn't update it's associated program config with new compiler options. A bit wordy, but this lead to unexpected test failures after the issue above was fixed, since we didn't have a case that setup just the program config & not the process group right afterwards.

Also patches a small issue in building $computedLibs found while testing, where duplicate lib & ddname entries would be present due to the Set usage w/ separate object refs.

### DIFF
--- a/packages/language/src/workspace/plugin-configuration-provider.ts
+++ b/packages/language/src/workspace/plugin-configuration-provider.ts
@@ -398,7 +398,8 @@ export class PluginConfigurationProvider {
   private async postProcessProcessGroups(): Promise<Diagnostic[]> {
     const diagnostics: Diagnostic[] = [];
     for (const processGroup of this.processGroupConfigs.values()) {
-      const computedLibs: Set<LibsEntry> = new Set();
+      // all computed libs for this group, dirs + ddnames
+      const computedLibsMap: Map<string, LibsEntry> = new Map();
       const libsToProcess = [...processGroup.libs];
       while (libsToProcess.length > 0) {
         const lib = libsToProcess.pop();
@@ -427,14 +428,15 @@ export class PluginConfigurationProvider {
                   // directory to add for handling
                   libsToProcess.push(`${lib}/${fileName}`);
                   // also add to the full libs list
-                  computedLibs.add({
-                    dir: `${lib}/${fileName}`,
+                  const dir = `${lib}/${fileName}`;
+                  computedLibsMap.set(`dir:${dir}`, {
+                    dir,
                   });
                 }
               }
             }
             // add the lib itself, now that we know it exists
-            computedLibs.add({
+            computedLibsMap.set(`dir:${lib}`, {
               dir: lib,
             });
           } catch (e) {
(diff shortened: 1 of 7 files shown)


packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -402,7 +402,8 @@ export class PluginConfigurationProvider {
   private async postProcessProcessGroups(): Promise<Diagnostic[]> {
     const diagnostics: Diagnostic[] = [];
     for (const processGroup of this.processGroupConfigs.values()) {
-      const computedLibs: Set<LibsEntry> = new Set();
+      // all computed libs for this group, dirs + ddnames
+      const computedLibsMap: Map<string, LibsEntry> = new Map();
```
ssmifi, line 405: Shouldn’t `computedLibs` have worked with the Set? If there’s a special case that requires this change, I’d suggest adding a test case.

montymxb: It was working, but not for removing duplicates. I overlooked the behavior of set for objects, which tracks references over values, so every entry was retained as if it were a regular list. Moving it to a map just gives us a way to denote libs that would be identical, and avoid collecting them once we already have an entry.

ssmifi: Makes sense. Thanks. Can we add a test case for removing duplicates to reflect that this is necessary?

montymxb: No prob. There is a test that I added under [plugin-configuration.test.ts](https://github.com/zowe/zowe-pli-language-support/pull/477/files#diff-0ea3be3ca13eab561f43c6fcf12b5aa359578341db171a546f900684cc40b21eR140), which was failing before this was patched since those dupes were present before. Is that what you're thinking of or are you asking for another test?

Also I didn't observe any problems from the duplicates being present, I just happened to notice it while I was working on this now, and figured it would be good to rectify here too.

msujew (approved):
Looks great to me 👍

PR #478, opened 2025-11-18
Adds Member Name Validation to DDName Style Includes

### LINKED ISSUE #399: Add member-name validation
Traditional Mainframe files system has 8 char limitation when naming members(files) - names are limited to 8 character.  

But on ZFS (unix like on Mainframe) can have 200 limits or no limits.  Eg. see below the member names can have anything and any limit. 

Image

So, suggested solution is: 
Add a field  **`member-name-validation: true`**  to config file 'proc_grps.json',

Image

 so 
when we open a document,   or include copybook, we check if it has valid name. 
Validation:  First character must not start with number but can be be a letter(a-z,A-Z), and other 
characters can be a 0-9,a-z,A-Z, and characters plus @, #, _, and $ only.

If   **`member-name-validation: false`**, then we skip the validation.

montymxb:
Closes #399 by adding an additional property to the proc groups configuration to enable member name validation. This applies a couple of checks to `%include member` style includes that:
- are longer than 8 chars
- don't start with A-Z and following with either alphanumeric chars or @, #, _ or $.

A diagnostic is emitted for each case separately.

Right now this only adds the diagnostic when we have a member style include that does _not_ resolve to a literal file (like `member.pli`), in all other non-string cases this assumes the include could be a member that needs validation. I've added a longer test case to verify this specific behavior, so we don't get member-focused diagnostics on regular file includes.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2313,27 +2313,40 @@ async function resolveIncludeFileUri(
       }
     }
 
-    if (isMemberWithoutDDName && pgroup.memberNameValidation) {
+    /**
+     * Helper to check & validate member names when member name validations are enabled.
+     * Pushes diagnostics to the context when validation fails.
+     * Effectively a noop when member name validation is disabled in the process group.
+     * @param memberName Member to validate, implied to be a member of an existing ddlib entry
+     */
+    function checkToValidateMember(memberName: string): void {
+      if (!pgroup?.memberNameValidation) {
+        return;
+      }
       // apply additional validation to the member name
-      if (fileNameOrPartial.length > 8) {
+      if (memberName.length > 8) {
         // emit diagnostic for member names > 8 characters
         context.diagnostics.push(
-          diagnostic(Severity.E, `Member exceeds 8 characters.`, item.token),
+          diagnostic(Severity.E, "Member exceeds 8 characters.", item.token),
         );
       }
 
       const memberNameRegex = /^[A-Z][A-Z0-9@#_$]*$/i;
-      if (!memberNameRegex.test(fileNameOrPartial)) {
+      if (!memberNameRegex.test(memberName)) {
         // emit a diagnostic for invalid member names
         context.diagnostics.push(
           diagnostic(
             Severity.E,
-            `Member must start with a letter, followed by letters, numbers, @, #, _, or $`,
+            "Member must start with a letter, followed by letters, numbers, @, #, _, or $.",
             item.token,
           ),
         );
       }
     }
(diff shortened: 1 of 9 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2264,6 +2269,41 @@ async function resolveIncludeFileUri(
...
+      const memberNameRegex = /^[A-Z][A-Z0-9@#_$]*$/i;
+      if (!memberNameRegex.test(memberName)) {
+        // emit a diagnostic for invalid member names
+        context.diagnostics.push(
+          diagnostic(
+            Severity.E,
+            "Member must start with a letter, followed by letters, numbers, @, #, _, or $.",
+            item.token,
+          ),
+        );
```
msujew, line 2337: Suggestion: Similar to the list of compiler options codes we have, can we add dedicated `SimplePLICode` elements for these? We already have some for LSP only features in [this file](https://github.com/zowe/zowe-pli-language-support/blob/development/packages/language/src/validation/lsp-codes.ts) (from the latest `development` branch). That makes both documentation of these special codes and testing easier.

msujew (commented):
I'm mostly happy with the changes, thanks! In addition to the Copilot comments (which are definitely worth addressing), see my one comment below.

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: Yep yep, that makes sense, I probably should have put special codes there in the first place.

packages/language/test/fourslash/preprocessor/include/with-member-validation.ts
```diff
@@ -0,0 +1,90 @@
...
+//// %INCLUDE <|4:_A1@#|>;
+//// // legitimate file include
+//// %INCLUDE <|5:__legitimatefile|>;
+
+// verify 1 diagnostic on the first include only
+verify.expectExclusiveDiagnosticsAt(1, [
+  {
+    severity: constants.Severity.E,
+    message: "Member exceeds 8 characters.",
+  },
```
msujew: Question: Can we use the `LspCodes` directly in here like we do with the other PLI codes? This should make this test more maintainable in the long run.

msujew (approved):
Looks good to me. Just one question/suggestion for the long term maintainability of the tests.

packages/language/test/fourslash/preprocessor/include/with-member-validation.ts
montymxb: Oh whoops missed that, thanks for catching that. I'll switch that over.

PR #479, opened 2025-11-18
Support `SQL TYPE IS BINARY` syntax

msujew:
Vit noticed this issue during testing. The [PL/I documentation](https://www.ibm.com/docs/en/db2-for-zos/12.0.0?topic=pli-host-variable-arrays-in#db2z_hostvariablearraypli__title__10) and the compiler seem to disagree on this. The documentation says this is a valid statement:

```
DCL X SQL TYPE IS BINARY;
```

Whereas the compiler will reject it with a missing `(` after `BINARY`:

```
IBM3754I S SQL TYPE IS BINARY must be followed by an opening left parenthesis.
```

This seems to be the correct syntax:

```
DCL X SQL TYPE IS BINARY(25);
```


packages/language/src/parser/parser-state.ts
```diff
@@ -107,21 +112,45 @@ export class ParserState {
...
    * until a successful consume() happens or the recover() method is called.
    */
+  error(message?: string, token?: t.Token, severity?: Severity): void;
+  error(message: SimplePLICode, token?: t.Token): void;
+  error<T extends ParametricPLICode>(
+    message: T,
+    args: Parameters<T["message"]>,
+    token?: t.Token,
+  ): void;
   error(
```
ssmifi, line 123: _Suggestion_: The overloaded optional parameters make it a bit hard to follow. For improved readability, consider reducing the nesting depth by returning early and clearly marking each case with SL comments.

packages/language/src/parser/parser-state.ts
```diff
@@ -107,21 +112,45 @@ export class ParserState {
...
+        const severity = (tokenOrSeverity || Severity.S) as Severity;
         const msg =
           message ??
           (this.eof
             ? "Unexpected end of file."
             : `Unexpected token '${generateTokenErrorName(token)}'.`);
         this.diagnostics.push(diagnostic(severity, msg, token));
-      } else {
-        this.diagnostics.push(diagnosticFromCode(message, token));
+      } else if ("message" in message) {
```
ssmifi: _Minor_: Could use `isPLICode` (or even `isParametricPLICode` and `isSimplePLICode`) here.

packages/language/test/fourslash/preprocessor/sql/binary-attribute.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../../framework.ts" />
+
+//// TEST: PROC;
+////   DCL TEST_SQL1 SQL TYPE IS BINARY(20);
+////   DCL TEST_SQL2 SQL TYPE IS VARBINARY(40);
```
ssmifi, line 16: _Suggestion_: Add a test with the old syntax and verify the error codes.

ssmifi (approved):
Looks good to me. 
If possible, I would try to increase the readability of `error` a bit (see below). 
You could also add a test for the new error codes.

Issue #472, opened 2025-11-13
DB2 exec sql support


asatklichov:
Implementation of DB2 sql preprocessors support. As per specification. 

TBD:  https://www.ibm.com/docs/en/db2-for-zos

PR #480, opened 2025-11-19
Add sql pp compiler options support

ssmifi:
Adding compiler option support for the sql preprocessor as part of https://github.com/zowe/zowe-pli-language-support/issues/472


msujew (approved):
Looks good to me

PR #486, opened 2025-11-20
Fix include resolution

### LINKED ISSUE #485: Broken Include resolutions
The commit `384f28aeb585ca0d9ee10cd96301cec82b3ff432` introduced a bug that breaks the include resolutions. Currently they aren't being resolved as they should.

Wagner-Laranjeiras:
Closes https://github.com/zowe/zowe-pli-language-support/issues/485.

Remove the global flag when searching for files for the include resolution.


msujew (approved):
Thanks, any idea why our tests didn't catch the issue?

PR #489, opened 2025-11-21
Replace string literal types in AST with enumerations

### LINKED ISSUE #481: [Refactoring] Translate string constant types into enums
Right now, our parser sets string constants on elements like the `DefaultAttribute` values. These are (1) prone to issues related to case-sensitivity and (2) slow to compare. We should instead use enums for these everywhere.

Lotes:
Closes #481

I added a script that checks if all tokens are mapping to the correct enumeration literals.
We could/should add it to the CI, sometimes there are missing tokens/literals.

So, I checked that every token maps to one or several enums. I also checked enum usages, whether conditions need to be adapted.

As you will review, you will discover that each enum in `ast.ts` has a `TokenType` inside of `tokens.ts`. So, for each `ast.XXX` enum, there are parsing methods: `tokens.XXX.map[To|From]EnumLiteral`.

Draft mode can be disabled. I could not discover any new errors in the given files and tests.


check-token-to-enum-mapping.ts
```diff
@@ -0,0 +1,113 @@
+import * as fs from "fs";
+import * as path from "path";
+import * as readline from "readline";
+import { fileURLToPath } from "url";
+
+/**
+ * With this script, we can test whether all AST-relevant enum are really parsed.
+ * This is useful when adding new tokens to ensure they are properly mapped to enums during the parsing process.
+ * Let's assume that we can have gaps in the mapping, but we want to be aware of them.
```
Lotes: Shall we add this to the CI? It is only relevant when tokens/AST enums are added.

Lotes:
Two further lines... any line also missed?
https://github.com/zowe/zowe-pli-language-support/blob/529080e58e37991a7c8717c5a73ace14c2cfea63/packages/language/src/syntax-tree/ast.ts#L1458
https://github.com/zowe/zowe-pli-language-support/blob/529080e58e37991a7c8717c5a73ace14c2cfea63/packages/language/src/syntax-tree/ast.ts#L2390

msujew:
I believe these are all, except for one:

https://github.com/zowe/zowe-pli-language-support/blob/2c3c2639fab1e96f9662cba81af3c33da27dd550/packages/language/src/syntax-tree/ast.ts#L1653-L1656

Can you replace those strings simply with the `Token` type? The actual values are pretty useless :D

check-token-to-enum-mapping.ts
Lotes: Made a test instead.

packages/language/src/language-server/hover-request.ts
```diff
@@ -17,7 +17,7 @@ import {
   isNameToken,
   isReferenceToken,
 } from "../linking/tokens";
-import { Token } from "../parser/tokens";
+import * as tokens from "../parser/tokens";
```
msujew: Minor: I believe in most other files we simply use `t`. We can simply follow the convention introduced by the other files.
```suggestion
import * as t from "../parser/tokens";
```

packages/language/src/parser/preprocessor-parser.ts
```diff
@@ -1490,21 +1492,25 @@ function parseBinary(state: ParserState): ast.Expression | null {
...
+    if (!operatorToken) {
       break;
     }
 
     const item = primary(state);
     infixOperatorItem.items.push(item);
-    infixOperatorItem.operators.push(operator);
+    infixOperatorItem.operators.push(
+      t.BinaryOperator.mapToEnumLiteral(operatorToken.image),
+    );
```
msujew, line 1512: Suggestion: Now that I think about it, using `image` here is quite problematic in case a user has specified the `OR` or `NOT` compiler option, since this will mess up the mapping. Can we somehow map from the `tokenTypeIdx` (which is always the same, independent of the compiler options)? Meaning that `mapToEnumLiteral` should accept a `number` instead.

msujew (commented):
Looks quite good. However, there is one change I would like to see in there (see below).

packages/language/src/parser/preprocessor-parser.ts
Lotes: Great idea. Done.

Lotes: But mapFromEnumLiteral is still a string!

msujew: > But mapFromEnumLiteral is still a string!

Yes, that makes perfect sense :)

msujew (approved):
Looks great, thank you!

Issue #464, opened 2025-11-06
[Type System] Support `LIKE` attribute


msujew:
Currently, the type system does not seem to take the `LIKE` attribute into account, which results in a false positive validation:

<img width="911" height="178" alt="Image" src="https://github.com/user-attachments/assets/51ad4a31-9b74-421b-a9bb-90f02fe197ab" />

msujew:
Resolved by https://github.com/zowe/zowe-pli-language-support/pull/523.

PR #491, opened 2025-11-24
Prevent false positives in type system

msujew:
We currently get false positive error diangostics on https://github.com/zowe/zowe-pli-language-support/issues/464 on certain attributes. This PR temporarily disables these errors to ensure that we don't display false positives to the user.

This also adds a way to add `todo`/`fail` tests to the fourslash test suite.


Lotes (approved):
LGTM, I like the test types.

PR #492, opened 2025-11-24
Add config option for configurable instruction counter limit

ssmifi:
Suggestion for adding a variable instruction counter limit for the preprocessor interpreter, which came up during todays meeting. Configurable on process groups to make it customizable per project.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -265,6 +265,7 @@ interface InterpreterContext {
    * It is invalid to invoke MACNAME outside of a preprocessor procedure.
    */
   macname: string;
+  instructionCounterLimit: number;
 }
 
 interface SqlAttributeCache {
@@ -296,6 +297,8 @@ export interface InterpreterOptions {
   compilerOptions: CompilerOptionResult | undefined;
   marginsProcessor: MarginsProcessor;
 }
+
+export const MAX_INSTRUCTION_COUNTER = 5000;
 
 export async function runInstructions(
   unit: CompilationUnit,
@@ -334,6 +337,9 @@ export async function runInstructions(
       entries: new Map(),
     },
     macname: "",
+    instructionCounterLimit:
+      unit.processGroup?.lspOptions.instructionCounterLimit ??
+      MAX_INSTRUCTION_COUNTER,
   };
   for (const [key, value] of instruction.procedures.entries()) {
     context.procedures.set(key, value);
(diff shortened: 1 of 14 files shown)


packages/vscode-extension/schemas/proc_grps.schema.json
```diff
@@ -53,16 +58,23 @@
               "check-margins": {
                 "type": "boolean",
                 "description": "Whether to check for margins in the PL/I code."
+              },
+              "instruction-counter-limit": {
+                "type": "number",
+                "description": "Maximum number of tries to run an instructions during preprocessing."
```
msujew: Minor: Not entirely accurate. I would also add hint about the default.
```suggestion
                "description": "Maximum number of times to run an instructions during preprocessing. 5000 by default."
```

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -121,6 +128,8 @@ export function deserializeProcessGroup(
   const libs = obj.libs || [];
   const lspOptions = obj["lsp-options"] || {};
   const checkMargins = lspOptions["check-margins"] ?? false;
+  const instructionCounterLimit =
+    lspOptions["instruction-counter-limit"] ?? 5000;
```
msujew: Suggestion: Since we have an `isNumber` check, we don't need the null check here.
```suggestion
  const instructionCounterLimit =
    lspOptions["instruction-counter-limit"];
```

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -136,6 +145,9 @@ export function deserializeProcessGroup(
       : new Set(),
     lspOptions: {
       checkMargins: isBoolean(checkMargins) ? checkMargins : false,
+      instructionCounterLimit: isNumber(instructionCounterLimit)
+        ? instructionCounterLimit
+        : 5000,
```
msujew: Suggestion: Can we use the constant here?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -334,6 +338,9 @@ export async function runInstructions(
       entries: new Map(),
     },
     macname: "",
+    instructionCounterLimit:
+      getConfiguration(uri, "lspOptions")?.instructionCounterLimit ??
```
msujew: Suggestion: You can use `unit.programConfig.lspOptions` here instead. No need to introduce the `getConfiguration` function.

msujew (commented):
Looks pretty good. A few improvements, see below.

msujew (approved):
Great, thanks!

Issue #454, opened 2025-10-22
[Improvement] Correctly deal with multiple workspace folders in the language server


msujew:
Below should be resolved within #668  

Currently, we mostly act like there is only ever one workspace folder open when accounting for the `.pliplugin` folder. We should instead look for the current workspace folder and then locate its `.pliplugin` folder instead.

Another small enhancement is - to support include (all type xinclude etc. )  or all other language feature 
on a program file opened out-of workspace. 

e.g.  When I open "remote-pgm.pli" which is outside the workspace, grammar is parsed, 
but it does not resolve copybooks.  See below snapshot
 
<img width="1823" height="538" alt="Image" src="https://github.com/user-attachments/assets/7501f534-911e-411b-84f9-dbffc111c0cc" />

asatklichov:
@Wagner-Laranjeiras  I updated this issue with to enhance copybook support for externally opened file.

asatklichov:
@Wagner-Laranjeiras  this item can be part of #668, let's consolidate it as part of 668

PR #493, opened 2025-11-25
Properly await language server startup

### LINKED ISSUE #487: [Improvement] Race conditions when computing libs.
**Problem:**

1. In some cases (ie. my machine) when running the extension, no includes get resolved. Then after modifying the "a.pli" file, the includes get resolved and errors go away.
2. After reloading the extension window the problem occur again, and the cycle repeats itself.

**Reason:**

Includes resolution can be negatively affected if the PluginConfiguration initializes before the Libs are properly computed. This usually isn't a problem, but it's a current consistent behavior on my machine (still figuring out why). The problems go away after modifying the file because, when we reindex it, the Libs are then already computed.

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/487, https://github.com/zowe/zowe-pli-language-support/pull/488.

Splits the language server initialization into two phases:

1. `onInitialize` only stores the workspace folders of the client and sends the capabilities back.
2. `onInitialized` reads the plugin configs for each workspace folder and marks the compilation unit handler as "ready"

The compilation unit handler will now **completely block** the compilation if `listen` has been called until `finalize` has been called to ensure that we don't accidentally start compiling code even though the plugin configs haven't been loaded yet.

This also fixes the startup of the web extension. So please test this as well in addition to the linked issue.

### DIFF
--- a/packages/language/src/language-server/connection-handler.ts
+++ b/packages/language/src/language-server/connection-handler.ts
@@ -25,6 +25,7 @@ import {
   Diagnostic,
   Location,
   TextEdit,
+  WorkspaceFolder,
 } from "vscode-languageserver-types";
 import {
   completionItemToLSP,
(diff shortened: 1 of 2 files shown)


packages/language/src/language-server/connection-handler.ts
```diff
@@ -54,22 +55,11 @@ export const WorkspaceDidChangePlipluginConfigNotification =
 export function startLanguageServer(connection: Connection): void {
   const compilationUnitHandler = new CompilationUnitHandler();
   compilationUnitHandler.listen(connection);
+  let folders: WorkspaceFolder[] = [];
 
   connection.onInitialize(async (params) => {
     // init the plugin config provider in reverse folder order, last plugin config encountered will take precedence
-    // TODO @montymxb Apr 23rd, 2025: Consider addressing multiple workspaces w/ multiple plugin configs
```
montymxb, line 60: We just forgot to remove the TODO from an earlier PR right? Just checking.

montymxb (approved):
:octocat: Approved, nice this looks really good!

packages/language/src/language-server/connection-handler.ts
msujew: The issue is still present, I just moved it into its own GitHub issue https://github.com/zowe/zowe-pli-language-support/issues/454 :)

PR #494, opened 2025-11-25
Refactor main parser

### LINKED ISSUE #482: [Refactoring] Translate the Chevrotain parser into our custom parser infrastructure
Currently, we have three different parsers in use:

1. A Chevrotain based compiler options parser
2. A handwritten preprocessor parser
3. A Chevrotain based "main" parser.

No (1) currently isn't that important. However, we should definitely refactor (3) to improve the error recovery and error message handling.

Lotes:
Closes #482

A lot has happened:

* transform Chevrotain to custom parser approach
* introduced rules using custom parser types with defined first sets (these sets are not always used)
* the lookahead variable K is 1 or 2 using the `sequence()` function. Alternatives can be combined with `choose()` and `orRule`
* custom lookhead function are at the end of the `parser.ts` file: We need one for assignments and one for the end statement
* I added a loop check mechanism for debugging mode, so that we can detect endless loops within tests. During production this loop check is empty.
* GENERIC attribute was duplicated in Chevrotain, I corrected this

* thanks to Mark for a first review
* the migration from Chevrotain was initially done with GPT (bad), later with Claude (better)
* according to a benchmark of Mark, the parser now is twice as fast as before this change


packages/language/src/parser/parser-handwritten.ts
```diff
@@ -0,0 +1,840 @@
...
+            element.labels.push(this.ruleLabelPrefix(state));
+        }
+
+        element.value = this.ruleUnit(state);
+
+        return element;
+    }
+
+    ruleUnit(state: ParserState): ast.Unit {
+        return state.consumeAlternatives<ast.Unit>([
```
msujew: This is a rather interesting one. Similar to the preprocessor parser, we can probably split this into two parts:

1. First, identify whether we have an assignment at hand (there is a function available for this). If this is the case, simply run the assignment parser function.
2. If not, this should be an LL(1) decision, meaning that we can use a `switch` over the current token type.

packages/language/src/parser/parser-handwritten.ts
```diff
@@ -0,0 +1,840 @@
...
+import { CstNodeKind } from "../syntax-tree/cst";
+
+export class HandwrittenParser implements Parser<ast.Program, tokens.Token> {
+    private _input: tokens.Token[] = [];
+    set input(value: tokens.Token[]) {
+        this._input = value;
+    }
+    get errors(): IRecognitionException[] {
+        return [];
+    }
```
msujew: Can we transform this class into a list of individual functions? Since they all pass along the `ParserState` anyway, they have all the context they need available. See also the current iteration of the preprocessor parser.

packages/language/src/parser/parser-handwritten.ts
```diff
@@ -0,0 +1,840 @@
...
+
+
+
+    rulePliProgram(state: ParserState): ast.Program {
+        const program: ast.Program = {
+            kind: ast.SyntaxKind.Program,
+            container: null,
+            statements: [],
+        };
```
msujew: The original implementation had these `create<Type>` functions. Some of them are already migrated into the `ast.ts` file. We should finish this migration as well as part of this PR and then simply use the `create<Type>` functions available there.

packages/language/src/parser/parser-handwritten.ts
```diff
@@ -0,0 +1,840 @@
...
+            kind: ast.SyntaxKind.ExportsItem,
+            container: null,
+            reference: null,
+        };
+        const token = state.consume(element, CstNodeKind.Exports_Procedure, tokens.ID)!;
+        element.reference = ast.createReference(
+            element,
+            token,
+            ast.ReferenceType.Variable,
+        );
```
msujew: IMO the `reference` should just stay `null` in case the token is `undefined`. Meaning that there should be a `if (token) { ... }` wrapper here.

msujew (commented):
I didn't click submit yesterday...

packages/language/src/parser/parser-handwritten.ts
Lotes: Ok, this is just a temporary thing so that I can exchange the parsers easily.
But I can give it a try at some point in time :)

Lotes: Done

packages/language/src/parser/parser-handwritten.ts
Lotes: Still need to do

packages/language/src/parser/parser.ts
```diff
@@ -17,8 +17,9 @@ import * as tokens from "./tokens";
 import * as ast from "../syntax-tree/ast";
 import { CstNodeKind } from "../syntax-tree/cst";
 import { ParserMethod, tokenMatcher } from "chevrotain";
+import { Parser } from "./parser-types";
 
-export class PliParser extends AbstractParser {
```
Lotes, line 21: Delete this file, finally

packages/language/src/parser/parser.ts
Lotes, line 1: Rename to `parser.ts`

packages/language/src/parser/parser.ts
```diff
@@ -0,0 +1,6536 @@
...
+      element.exports = exports.rule(state);
+    }
+    if (state.canConsumeFirst(reserves.first())) {
+      element.reserves = reserves.rule(state);
+    }
+    if (state.canConsumeFirst(options.first())) {
+      element.options = options.rule(state);
+    }
+    state.consume(element, CstNodeKind.Package_Semicolon0, tokens.Semicolon);
+    const { inc } = state.createLoopContext();
```
Lotes: Shall we keep these? They help a lot in endless loops, so that the synchronous test do not break the test process.
```ts
const { inc } = state.createLoopContext();
while (...) {
  inc();
  //...
}
```

packages/language/src/parser/parser.ts
```diff
@@ -0,0 +1,6536 @@
...
+      CstNodeKind.BeginStatement_Semicolon1,
+      tokens.Semicolon,
+    );
+
+    return element;
+  },
+);
+
+const endStatement = rule(
+  choice(
```
Lotes: Add a custom lookahead. Make first set invalid.

packages/language/src/parser/parser.ts
```diff
@@ -0,0 +1,6536 @@
...
+  () => getStatement,
+  () => goToStatement,
+  () => ifStatement,
+  () => iterateStatement,
+  () => leaveStatement,
+  () => locateStatement,
+  () => nullStatement,
+  () => onStatement,
+  () => openStatement,
+  () => procincDirective, // TODO integrate into preprocessor
```
Lotes: What about this TODO? It it only for the preprocessor? Then we can delete it or?

packages/language/src/parser/parser.ts
```diff
@@ -0,0 +1,6536 @@
...
+  () => defaultExpressionPart.first(),
+  (state: ParserState): ast.DefaultExpression => {
+    const element = ast.createDefaultExpression();
+
+    element.expression = defaultExpressionPart.rule(state);
+
+    const { inc } = state.createLoopContext();
+    while (state.canConsumeFirst(defaultDeclarationAttribute.first())) {
+      inc();
+      switch (performGenericAttributeLookahead(state)) {
```
Lotes: There are 2 different branches where GENERIC can appear. Need to check if this custom lookahead is really needed and if we can make it to one single branch.

packages/language/src/parser/parser-state.ts
```diff
@@ -50,6 +51,25 @@ export class ParserState {
     this.index = 0;
   }
 
+  createLoopContext(id: number = 1) {
+    const regex = /parser\-handwritten\.ts:(\d+)/;
```
Lotes: Oha, may find a different Regex to get the line number. Or find a different indicator / id.

packages/language/src/parser/parser-state.ts
```diff
@@ -50,6 +51,25 @@ export class ParserState {
...
+    const stack = new Error().stack!.split("\n");
+    const location = stack.find((line) => regex.test(line))!;
+    const match = location.match(regex)![1] + ` (loop id: ${id})`;
+    let lastIndex = -1;
+    return {
+      inc: () => {
+        if (lastIndex === this.index) {
+          ///this.inError = true;
+          throw new Error(
```
Lotes: This was meant to panic and leave the endless loop during tests. In production there should be no such situation. Is there a better solution?

packages/language/src/parser/parser-state.ts
msujew: I believe this currently breaks the language server if running in vscode. We only have a `main.js` file in there.

packages/language/src/parser/parser-state.ts
msujew: Can we make this conditional? Ideally, in production, we shouldn't even have to perform the check at all. Maybe something like setting an environment variable when using the extension launch config & running the tests. Otherwise return an empty function in `inc`?

packages/language/src/parser/parser.ts
msujew: We can delete this, yes.

packages/language/src/parser/parser.ts
msujew: I believe it makes sense to keep these for now. Eventually, we might want to remove them, but I think it's good for now.

packages/language/src/parser/abstract-parser.ts
```diff
@@ -335,9 +335,24 @@ export function constructBinaryExpression(
...
+        opToken,
+        right: null,
+      };
+      opToken.element = binaryExpr;
+      return binaryExpr;
+    }
     // Captured just a single, non-binary expression
     // Simply return the expression as is.
-    return obj.items[0];
+    return left;
```
msujew: Suggestion: Revert these changes and rebase on the latest `development` state. The issue is already fixed in there.

packages/language/src/parser/parser.ts
```diff
@@ -0,0 +1,6536 @@
...
+  const lookahead: (la: number) => tokens.Token | undefined = (la) => {
+    previousToken = state.peek(la - 1);
+    return state.peek(la);
+  };
+  let i = 1;
+  let token = lookahead(i++);
+  // First token of an assigment needs to be an ID
+  if (!token || !tokenMatcher(token, tokens.ID)) {
+    return false;
+  }
```
msujew, line 6425: Suggestion: If the token is of token `ID` (but not a keyword!) we should be able to directly return true here:
```suggestion
  // First token of an assigment needs to be an ID
  if (!token || !tokenMatcher(token, tokens.ID)) {
    return false;
  } else if (token.tokenTypeIdx === tokens.ID.tokenTypeIdx) {
    return true;
  }
```

There is no other statement that can parse an `ID` token.

packages/language/src/parser/parser-state.ts
```diff
@@ -50,6 +51,25 @@ export class ParserState {
     this.index = 0;
   }
 
+  createLoopContext(id: number = 1) {
+    const regex = /parser\-handwritten\.ts:(\d+)/;
+    const stack = new Error().stack!.split("\n");
+    const location = stack.find((line) => regex.test(line))!;
+    const match = location.match(regex)![1] + ` (loop id: ${id})`;
```
msujew: Suggestion: Using the stack here is quite hacky and breaks in the actual language server. Shall we instead add a name to all `rule` function calls that ensures we can always track the current rule name?

packages/language/src/parser/parser-state.ts
```diff
@@ -246,6 +282,58 @@ export class ParserState {
...
+
+  private consumeAlternativesItem<T>(map: RuleMap<T>, index: number): T | null {
+    if (this.eof || this.inError) {
+      return null;
+    }
+    const token = this.tokens[this.index + index];
+    const lookahead = token.tokenType;
+    const idx = this.mapMatch(map, lookahead);
+    if (idx === undefined) {
+      //TODO this.error(`Expected '', but found '${lookahead.name}'.`, token, Severity.S)
```
msujew: Question: Can we extract a reasonable message from the map to display here?

msujew (commented):
I'm pretty happy with the general parser code. There are some improvements wrt. the surrounding infrastructure, see comments below :)

packages/language/src/parser/parser-state.ts
```diff
@@ -48,12 +53,35 @@ export class ParserState {
...
     this.mode = mode;
     this.diagnostics = [];
     this.index = 0;
+    this.debugLoops = debugLoops;
+  }
+
+  createLoopContext(name: string) {
+    let lastIndex = -1;
+    if (this.debugLoops) {
+      return {
```
msujew: Minor:
```suggestion
    if (this.debugLoops) {
      let lastIndex = -1;
      return {
```

packages/language/src/parser/parser-state.ts
```diff
@@ -166,11 +194,27 @@ export class ParserState {
...
     this.inError = false;
   }
 
+  private performMainParserRecovery(): void {
+    const currentLine = (this.token || this.last)?.startLine ?? 0;
+    while (this.index < this.tokens.length) {
+      const token = this.tokens[this.index];
+      if (token.startLine !== currentLine) {
+        // Moved to next line, stop here
+        break;
```
msujew: Suggestion: Now that I think about this, this doesn't work well in the presence of preprocessor code. Since preprocessor generated code has "unpredictable" start-end-line behavior (i.e. a token with line 1000 can be followed by a token from line 1), we should probably ignore this case. It is still useful in the preprocessor, since all tokens are correctly ordered there.

However, IMO there is another, more useful recovery heuristic: If the next token is a token that signals the start of a new statement, we should immediately recover. I.e. something like `CALL`, `DEFINE`, `DECLARE`, etc. Theoretically, in order to do this completely, we should also run the "assignment-lookahead" here. Since we only do this in recovery, taking the small performance hit here is probably ok!

packages/language/src/parser/parser-state.ts
```diff
@@ -301,6 +345,66 @@ export class ParserState {
     }
     return this.consume(element, kind, tokenType);
   }
+
+  private mapMatch(map: RuleMap, tokenType: TokenType): number | undefined {
+    return [
+      tokenType.tokenTypeIdx ?? -1,
+      ...(tokenType.CATEGORIES ?? []).map((cat) => cat.tokenTypeIdx ?? -1),
+    ].find((index) => map.has(index));
```
msujew: Suggestion: Transforming this into multiple arrrays seems quite slow to me. We can do this better with a guard and a loop:


```suggestion
    const tokenTypeIdx = tokenType.tokenTypeIdx ?? -1;
    if (map.has(tokenTypeIdx) {
      return tokenTypeIdx;
    }
    if (tokenType.CATEGORIES) {
      for (const category of tokenType.CATEGORIES) {
        const categoryIdx = category.tokenTypeIdx ?? -1;
        if (map.has(categoryIdx) {
          return categoryIdx;
        }
      }
    }
    return undefined;
```

packages/language/src/parser/parser-state.ts
```diff
@@ -301,6 +345,66 @@ export class ParserState {
...
+    const lookahead = token.tokenType;
+    const idx = this.mapMatch(map, lookahead);
+    if (idx === undefined) {
+      const tokenTypeNames = [...map.keys()]
+        .map((idx) => tokenIdxToClass(idx))
+        .filter((k) => k !== undefined)
+        .map((tk) => tk!.name)
+        .join(", ");
+      this.error(
+        `Expected any of {${tokenTypeNames}}, but found '${lookahead.name}'.`,
```
msujew: Minor: The image works better here, since it will show `Expected any of {a, b} but found 'c'.` instead of  `Expected any of {a, b} but found 'ID'.`. The user doesn't care about the actual technical token most of the time.
```suggestion
        `Expected any of {${tokenTypeNames}}, but found '${lookahead.image}'.`,
```

packages/language/src/parser/parser.ts
```diff
@@ -1,8357 +1,6407 @@
...
+
+export function parsePli(
+  input: tokens.Token[],
+  debugging: boolean,
+): {
+  tree: ast.Program;
+  diagnostics: any;
+} {
+  const state = finalParserState(input, debugging);
+  const program = pliProgram.rule(state)!;
```
msujew:
```suggestion
  const program = pliProgram.rule(state);
```

packages/language/test/rules.test.ts
msujew, line 1: Suggestion: Assuming these tests are still important, can we move them to the parser-state tests?

packages/language/test/parsing.test.ts
```diff
@@ -959,4 +959,16 @@ describe("PL/I Parsing tests", () => {
...
+    const doc = await parseStmts(`
+      DO LOOP;
+      last:
+      label:
+      before:
+      END;
+    `);
+    assertNoParseErrors(doc);
+    generateAndAssertValidSymbolTable(doc);
+  });
```
msujew: Suggestion: I think we can make this test even better by moving it into the fourslash framework and use the `expectAst` verification function.

packages/language/src/workspace/lifecycle.ts
```diff
@@ -80,19 +79,17 @@ export async function tokenize(
...
 
 export function parse(compilationUnit: CompilationUnit): Program {
-  PliParserInstance.input = compilationUnit.tokens;
-  const ast = PliParserInstance.parse();
-  compilationUnit.ast = ast;
-  compilationUnit.diagnostics.addAll(
-    DiagnosticCategory.Parser,
-    parserErrorsToDiagnostics(PliParserInstance.errors),
-  );
+  const debugging = process.env.NODE_ENV === "development";
```
msujew: Suggestion: Can we move this into a utility somewhere? This way, we can access the info directly from the parser and don't need to pass it into the parser. And also make this a bit more web-proof by doing something like this:


```suggestion
  const debugging = typeof process !== 'undefined' && process.env?.NODE_ENV === "development";
```

msujew (commented):
Really good. Just a few minor improvements, see below. Afterwards, this is good to go!

packages/language/test/rules.test.ts
Lotes: I would prefer them in a separate file, since I am using a fixture that is only relevant here.

Lotes:
`CHART.pli` looks very broken 👀 . But no big change to the old parser, it seems.

msujew:
Yes, `CHART.pli` uses a `PUT` syntax that is undocumented and does not fit the syntax diagram shown in the language reference. This is expected :/

packages/language/src/workspace/environment.ts
```diff
@@ -0,0 +1,2 @@
+export const IsDebugging =
```
msujew, line 12: Suggestion: Please add the license header.

packages/language/test/rules.test.ts
```diff
@@ -0,0 +1,170 @@
+import { TokenType } from "chevrotain";
```
msujew, line 12: Suggestion: Please add the license header.

packages/language/src/parser/parser-types.ts
```diff
@@ -0,0 +1,168 @@
+import { IRecognitionException, IToken, TokenType } from "chevrotain";
+import { ParserState } from "./parser-state";
+import { memoize } from "lodash-es";
+
+export interface Parser<TAst, TToken extends IToken = IToken> {
+  set input(value: TToken[]);
+  get errors(): IRecognitionException[];
+  parse(): TAst;
+}
```
msujew: I don't think this interface is used anywhere, right?

msujew (approved):
Fantastic work, looks good to me! Please don't forget to squash your changes when merging :)

PR #495, opened 2025-11-25
Use full include member range for definition

### LINKED ISSUE #466: Improve Definition Highlighting for Includes
Originally mentioned in #455. Referenced in https://github.com/zowe/zowe-pli-language-support/blob/development/packages/language/src/language-server/definition-request.ts#L76-L77.

> Todo: Highlight full source range when performing the definition request on the item:
Image

Possibly related, underlining for definition requests in literal file includes via strings also appears to have a similar issue.
Image

msujew:
Closes #466

Adds the full range of the respective include item to the AST node. Uses this information to calculate the `sourceRange` for the definition request to fix this issue at hand.

### DIFF
--- a/packages/language/src/language-server/connection-handler.ts
+++ b/packages/language/src/language-server/connection-handler.ts
@@ -24,6 +24,7 @@ import {
   CodeAction,
   Diagnostic,
   Location,
+  LocationLink,
   TextEdit,
 } from "vscode-languageserver-types";
 import {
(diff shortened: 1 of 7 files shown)


montymxb (approved):
:octocat: Great this looks fantastic!

On a side note I was checking multi-includes, such as `%INCLUDE a b c;`, that also works nicely in all cases, but when I accidentally wrote up this invalid case:
```
 %include def"b.pli";
```
The LS completely locked up and stopped responding altogether. That's a separate issue however that I'll add up quickly.

PR #496, opened 2025-11-26
[Cleanup] Remove support for SKIP directive.

Wagner-Laranjeiras:
### [Cleanup] Remove SKIP line-skipping feature and add standalone SKIP warning (LSPIS001)

Issue #410 

**Summary:**

This PR removes support for the old “skipped lines” feature that visually marked lines as skipped when encountering a %SKIP directive. Per new requirements, the language server should no longer interpret SKIP as a line-skipping construct.
In place of the previous behavior, the language server now issues a warning only when SKIP appears as a standalone directive, such as:
```
%SKIP;
%SKIP(n);
```
Other forms such as %DO SKIP, used within valid constructs are unaffected.

**Changes Included**

✔ Removed
- The logic that tracked and displayed “skipped lines” in the editor and related tests.

✔ Added
- A new diagnostic LSPIS001 (Standalone SKIP statement is not supported by the language server.)
- A validator that triggers this warning only for standalone SkipDirective nodes
- Added tests for standalone SKIP warnings.


packages/language/src/validation/language-server/LSPIS001-skip-statement-not-supported.ts
```diff
@@ -0,0 +1,23 @@
...
+import { ValidationAcceptor } from "../validator";
+import * as AST from "../../syntax-tree/ast";
+import { diagnosticFromCode } from "../../language-server/types";
+import { LspCodes } from "../lsp-codes";
+
+export function LSPIS001_standalone_skip_directive_not_supported(
+  node: AST.SkipDirective,
+  acceptor: ValidationAcceptor,
+): void {
+  if (node.kind !== AST.SyntaxKind.SkipDirective || !node.token) return;
```
msujew: Minor: I believe we can always know that the current node is of type `AST.SkipDirective`.
```suggestion
  if (!node.token) return;
```

msujew (approved):
Thanks, looks good to me 👍

PR #497, opened 2025-11-26
Prevent diagnostics pollution of compiler options

msujew:
Currently, the `filteredDiagnosticsWithUri` function will simply *assign* the current `URI` to any diagnostic generated by the compiler options translator. This is incorrect, and will result in issues assigned to the current file URI, even if it doesn't originate from there. This was apparent when using the `SQL` and `MACRO`, compiler sub options.

This change ensures that only the compiler options from the current file get the `URI` attached to the tokens.

FYI: This change does not contain tests, because I plan to refactor some of the compiler options stuff anyway, which would make any test I write now obsolete.

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/parser.ts
+++ b/packages/language/src/preprocessor/compiler-options/parser.ts
@@ -31,6 +31,7 @@ import {
   SyntaxKind,
 } from "../../syntax-tree/ast";
 import { ML_COMMENT, SL_COMMENT, Token } from "../../parser/tokens";
+import { URI } from "../../utils/uri";
 
 const commaToken = createToken({ name: "comma", pattern: "," });
 const semicolonToken = createToken({ name: "semicolon", pattern: ";" });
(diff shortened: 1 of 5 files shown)


packages/language/src/preprocessor/compiler-options/parser.ts
```diff
@@ -285,11 +287,15 @@ export function parseAbstractCompilerOptions(
     0,
     semicolonTokenPosition === -1 ? undefined : semicolonTokenPosition,
   ) as Token[];
+  for (const token of tokens) {
```
Lotes: Minor: If `uri` is undefined, you don't need the loop, I guess. Missing IF statement... no big effect ^^*

packages/language/src/workspace/lifecycle.ts
```diff
@@ -75,7 +74,7 @@ export async function tokenize(
   );
   compilationUnit.diagnostics.addAll(
```
Lotes, line 75: Does this mean that diagnostics with a `undefined` URI will disappear? Do we want this?

msujew: They currently disappear anyway (except for the ones that incorrectly pollute the wrong files). I'm currently working on a change that displays issues correctly on the JSON file itself.

Lotes (approved):

PR #500, opened 2025-11-29
Improve AST robustness

msujew:
In some rare cases, we received the following error:

```
TypeError: Cannot set properties of null (setting 'container')
```

This change adjusts the `AbstractParser` to ensure that we never call the `assign` function with `null` (we previously only checked for `undefined`). It also improves the computation of erroneous binary expressions like `A + `, which should still generate a `BinaryExpression` element with `right = null`.

### DIFF
--- a/packages/language/src/parser/parser-entry.ts
+++ b/packages/language/src/parser/parser-entry.ts
@@ -35,16 +35,13 @@ export function preprocessorParse(
   let index = state.index;
   let token = state.token;
   while (token) {
+    let stmt: ast.Statement | null = null;
     let isTokenStatement = true;
     switch (token.tokenTypeIdx) {
       case t.Percent.tokenTypeIdx:
         isTokenStatement = false;
         // Parse a preprocessor statement
-        const ppStmt = statement(state);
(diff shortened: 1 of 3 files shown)


packages/language/src/parser/parser-entry.ts
```diff
@@ -53,25 +50,25 @@ export function preprocessorParse(
...
     }
     if (isTokenStatement) {
       // Otherwise construct a token statement
-      const tokenStmt = consumeTokenStatement(state);
-      recursivelySetContainer(tokenStmt);
-      statements.push(tokenStmt);
+      stmt = consumeTokenStatement(state);
+    }
+    if (stmt) {
+      recursivelySetContainer(stmt);
```
Lotes, line 70: Nice!

Lotes (approved):
LGTM. Can you imagine a test scenario, where the test was red before the fix?

msujew:
> Can you imagine a test scenario, where the test was red before the fix?

Unfortunately, I wasn't able to reproduce the report.

PR #501, opened 2025-12-01
Support additional SQL preprocessor attributes

msujew:
Supports the missing SQL preprocessor attributes from [here](https://www.ibm.com/docs/en/db2-for-zos/13.0.0?topic=pli-host-variables-in). Previously, we only implemented the ones from [this page](https://www.ibm.com/docs/en/db2-for-zos/13.0.0?topic=pli-host-variable-arrays-in).

In total this PR contains changes to:

1. Support the `SQL TYPE IS TABLE LIKE <table-name> AS LOCATOR` syntax
2. Support the `SQL TYPE IS RESULT_SET_LOCATOR VARYING` syntax
3. Remove `BIN` as an individual keyword again. Even though the syntax diagrams never specify it, but `BIN` is ***always*** a valid short version of `BINARY`, even in the SQL preprocessor.

### DIFF
--- a/packages/language/src/parser/sql-parser.ts
+++ b/packages/language/src/parser/sql-parser.ts
@@ -61,32 +61,36 @@ export function sqlAttributeStatement(
       state,
       attributeStatement.isXml,
     );
-  } else if (state.canConsume(t.CharOrBinary)) {
+  } else if (state.canConsume(t.BINARY) || state.canConsume(t.CHARACTER)) {
     attributeStatement.body = parseSQLAttributeExplicitLob(
       state,
       attributeStatement.isXml,
     );
   } else if (state.canConsume(t.LOBLocator)) {
-    attributeStatement.body = ast.createSQLAttributeLobLocator();
+    attributeStatement.body = ast.createSqlAttributeLobLocator();
     state.consume(
       attributeStatement.body,
       CstNodeKind.SqlAttributeLobLocator_LOB_LOCATOR,
       t.LOBLocator,
     );
   } else if (state.canConsume(t.LOBFile)) {
-    attributeStatement.body = ast.createSQLAttributeLobFile();
+    attributeStatement.body = ast.createSqlAttributeLobFile();
     state.consume(
       attributeStatement.body,
       CstNodeKind.SqlAttributeLobFile_LOB_FILE,
       t.LOBFile,
     );
   } else if (state.canConsume(t.ROWID)) {
-    attributeStatement.body = ast.createSQLAttributeRowId();
+    attributeStatement.body = ast.createSqlAttributeRowId();
     state.consume(
       attributeStatement.body,
       CstNodeKind.SqlAttributeRowId_ROWID,
       t.ROWID,
     );
+  } else if (state.canConsume(t.TABLE)) {
+    attributeStatement.body = parseSqlTableLocator(state);
+  } else if (state.canConsume(t.RESULT_SET_LOCATOR)) {
+    attributeStatement.body = parseSqlResultSetLocator(state);
   }
   return attributeStatement;
 }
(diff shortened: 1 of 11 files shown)


packages/language/src/parser/sql-parser.ts
```diff
@@ -130,25 +134,24 @@ function parseSQLAttributeExplicitLob(
...
+    );
+    if (binToken) {
       lob.type = ast.SQLAttributeLobType.BLOB;
       typename = "BLOB";
-      break;
+    }
   }
+  state.consume(lob, CstNodeKind.SqlAttributeLob_LARGE, t.LARGE);
+  state.consume(lob, CstNodeKind.SqlAttributeLob_OBJECT, t.OBJECT);
   parseSQLAttributeLobSize(state, lob, isXml, typename);
```
Lotes, line 155: Instead of strings for typename you could use the corresponding enum.

packages/language/src/parser/tokens.ts
```diff
@@ -983,21 +980,16 @@ export const CANCEL = registerKeyword({
   name: "CANCEL",
 });
 export const BINARY = registerKeyword({
-  name: "BINARY",
+  name: ["BINARY", "BIN"],
```
Lotes, line 983: Was it not a problem that BIN and BINARY are handled differently? I think Azat showed us some syntax that only works with BINARY and without BIN

packages/language/test/fourslash/preprocessor/sql/binary-attribute-with-bin.ts
```diff
@@ -0,0 +1,23 @@
...
+
+//// TEST: PROC;
+////   DCL TEST_SQL SQL TYPE IS BIN(20);
+//// END;
+
+// BINARY and BIN should be exchangable
+preprocessor.expectTokens(`
+TEST: PROC;
+  DCL TEST_SQL CHAR(20) NONVARYING;
+END;
```
Lotes, line 22: True story? SQL TYPE BIN becomes a CHAR?

packages/language/src/parser/tokens.ts
msujew: I tested a bunch of cases against the compiler. It seems to treat `BIN` the same as `BINARY`, even if the documentation says only `BINARY` is supported. I used to think this as well, but using something like:

```
DCL BIN_TEST SQL TYPE IS BIN(10);
```

results in the exact same output as:

```
DCL BIN_TEST SQL TYPE IS BINARY(10);
```

packages/language/src/parser/sql-parser.ts
msujew: Doesn't really work, see [this case](https://github.com/zowe/zowe-pli-language-support/blob/9680ea5b43e761b1a7549c8c706fa8a5fad43c25/packages/language/src/parser/sql-parser.ts#L114).

packages/language/test/fourslash/preprocessor/sql/binary-attribute-with-bin.ts
msujew: I just ran this against the compiler:

```
DCL BIN_TEST SQL TYPE IS BIN(10);
```

Which results in this code after the SQL preprocessor:

```
DCL BIN_TEST CHAR(10) NONVARYING;
```

Don't ask me why. The [documentation says the same](https://www.ibm.com/docs/en/db2-for-zos/13.0.0?topic=pli-host-variables-in#db2z_hostvariablepli__title__9).

Lotes: OK, I see, thanks

Lotes (approved):
LGTM

PR #502, opened 2025-12-01
Support `INITIAL` attribute in preprocessor

msujew:
This attempts to add support for the `INITIAL` attribute of the `DECLARE` statement. Even though the attribute is not supported in the language reference, the compiler accepts the attribute and sets the initial values as expected when using it.

There is one major feature of this still missing, mainly because the AST structure doesn't support this case well:

```
DCL A(2) CHAR INIT((2)"a");
```

We can add this later in a separate PR.


ssmifi (approved):
LGTM

PR #505, opened 2025-12-02
New diagnostic for unresolved INCLUDE

Wagner-Laranjeiras:
### List of Changes

* Introduced new diagnostic:

  ```ts
  FileNotFound: {
    code: "LSPIR002",
    severity: Severity.S,
    message: (includefilename: string) =>
      `INCLUDE file "${includefilename}" not found.`,
  },
  ```

* Replaced the previous diagnostic:

  ```ts
  /**
   * The INCLUDE file could not be found, or if found, it could not be opened.
   * (see page 171)
   */
  IBM3841I: {
    code: "IBM3841I",
    severity: Severity.S,
    message: (includefilename: string) =>
      `The INCLUDE file ${includefilename} could not be opened.`,
  },
  ```

* Updated all relevant tests to reflect the new diagnostic:

  * Expected code changed from `IBM3841I` → `LSPIR002`
  * Updated expected error message text
  * Updated tests accordingly

### DIFF
--- a/packages/language/src/validation/lsp-codes.ts
+++ b/packages/language/src/validation/lsp-codes.ts
@@ -19,12 +19,6 @@ export const LspCodes = {
       message:
         "Could not resolve include directive. Plugin configuration is missing",
     },
-    // FileNotFound: {
-    //   code: "LSPIR002",
-    //   severity: Severity.S,
-    //   message: (includefilename: string) =>
-    //     `INCLUDE file "${includefilename}" not found.`,
-    // },
   },
 
   UpperCase: {
--- a/packages/language/src/validation/lsp-codes.ts
+++ b/packages/language/src/validation/lsp-codes.ts
@@ -19,12 +19,12 @@ export const LspCodes = {
       message:
         "Could not resolve include directive. Plugin configuration is missing",
     },
-    FileNotFound: {
-      code: "LSPIR002",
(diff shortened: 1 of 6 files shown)


packages/language/src/validation/lsp-codes.ts
```diff
@@ -19,6 +19,12 @@ export const LspCodes = {
       message:
         "Could not resolve include directive. Plugin configuration is missing",
     },
+    FileNotFound: {
+      code: "LSPIR002",
+      severity: Severity.S,
+      message: (includefilename: string) =>
+        `INCLUDE file "${includefilename}" not found.`,
+    },
```
msujew: Question: Is `IBM1848I` close enough to this, or do you think it's better to have a completely separate code (i.e. not one that's thrown by the compiler)?

msujew (commented):
I think that makes sense, though there might be a more fitting, existing error code for this. See below.

packages/language/src/validation/lsp-codes.ts
Wagner-Laranjeiras: `IBM1848I` would be indeed even better, since we wouldn't need to add unnecessary code. Gonna implement it and create a new PR. Thanks!

msujew (approved):
Thanks, LGTM.

Issue #499, opened 2025-11-27
[Improvement] Adjust the relative display path for `INCLUDE` statements


msujew:
Currently, when hovering over an include, it will show the relative file path from the current file. Instead, we want to show two different things depending on the current workspace:

1. If the file included is in the current workspace, show the path relative from the workspace root.
2. If the file is outside of the current workspace, show the absolute path on the file system.

asatklichov:
done

PR #506, opened 2025-12-02
Enhance hover preview for includes with relative and absolute file paths

Wagner-Laranjeiras:
**Issue #499** 

Enhances the hover preview content for included files by showing paths more meaningfully:

    **1. Files inside the workspace:** display the path relative to the workspace root for easier readability and context.
    **2. Files outside the workspace:** display the full absolute path so developers can locate them on the filesystem.
    
**Limitations:**
    There's a syntax limitation with this implementation, which is: whenever we display an absolute path, we are actually assigning the absolute file path value to a relative file path string, which is syntactically incorrect. Any suggestion in that regards (and also in general) is welcome.


msujew (commented):
I don't think this addresses point (1) of https://github.com/zowe/zowe-pli-language-support/issues/499. The relative path is currently still pointing from the original including file to the included file. But it should be relative to the **workspace root**.

msujew (commented):
Please add a test case for each type of include (relative and absolute).

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2069,6 +2073,23 @@ function setFilePath(
...
         relative = "./" + relative;
+      } else if (
+        // Check if the path IS absolute
+        relative.startsWith("../") ||
+        relative.startsWith("/") ||
+        (relative.charAt(1) === ":" && relative.charAt(2) === "/")
+      ) {
+        // In this case, we are setting the absolute path as the relative path
+        // because it's the value showcased on the preview.
+        if (isWindows) {
```
ssmifi: _Question_: There is no link to `relative` anymore, but only to the location of `currentUri`. Is this intended? Shouldn't there be a link to the `filePath` anywhere?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2049,16 +2049,20 @@ async function runIncludeInstruction(
...
 ): void {
   if (filePath) {
     item.filePath = filePath;
     if (context.currentUri) {
-      const dirname = UriUtils.dirname(context.currentUri);
-      let relative = UriUtils.relative(dirname, filePath);
+      const workspace = PluginConfigurationProviderInstance.getWorkspacePath();
+      let relative = UriUtils.relative(workspace, filePath);
+      let isWindows = false;
+      if (context.currentUri.path.charAt(1).toLowerCase() === "c") {
```
ssmifi: _Question/Suggestion_: This looks fragile and only detects content on drive c. Do you need the logic at all? Couldn't we just use uri.fsPath + path below?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2049,16 +2049,20 @@ async function runIncludeInstruction(
...
+        isWindows = true;
```
ssmifi: _Just a hint_: You can simply use the expression for boolean toggles:
`const isWindows = context.currentUri.path.charAt(1).toLowerCase() === "c";`

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2049,16 +2049,20 @@ async function runIncludeInstruction(
  * @param filePath File path to set, if null, no changes are made
  * @param context Used for currentUri to calculate relative paths
  */
-function setFilePath(
+export function setFilePath(
   item: { filePath: string | null; relativeFilePath: string | null },
   filePath: string | null,
   context: InterpreterContext,
 ): void {
   if (filePath) {
```
ssmifi: _Minor_: You can exit early here.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2049,16 +2049,20 @@ async function runIncludeInstruction(
...
     item.filePath = filePath;
     if (context.currentUri) {
```
ssmifi: _Question_: Is it ok to set an `item.filePath` without an `item.relativePath`?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2069,6 +2073,23 @@ function setFilePath(
       ) {
         // Make sure the path is explicitly relative
         relative = "./" + relative;
+      } else if (
+        // Check if the path IS absolute
+        relative.startsWith("../") ||
```
ssmifi: _Question_: This looks confusing to me. Isn't this a relative path?

packages/language/src/utils/uri.ts
```diff
@@ -42,6 +44,13 @@ export namespace UriUtils {
         toParts[0] = toParts[0].toLowerCase();
       }
       if (fromParts[0] !== toParts[0]) {
+        // Here is a check to see if we have the workspace folder somewhere in our path.
+        // The limitation is that, if somehow we have a folder with the same name of the workspace folder, within
+        // an absolute path outside the real workspace, it would be catch.
+        const indexExists = toParts.indexOf(fromParts[0]);
```
ssmifi: This does return if the first segment is found anywhere in `toParts`. This would match even if there is no complete common prefix:

> fromPath = /workspace/project
> toPath   = /random/workspace/file.txt

packages/language/test/preprocessor/set-file-path.test.ts
```diff
@@ -0,0 +1,169 @@
...
+  beforeEach(async () => {
+    item = { filePath: null, relativeFilePath: null };
+
+    vfs = new VirtualFileSystemProvider();
+    pluginConfig = new PluginConfigurationProvider();
+
+    setFileSystemProvider(vfs);
+    setPluginConfigurationProvider(pluginConfig);
+
+    if (["darwin", "linux"].includes(process.platform)) {
```
ssmifi: _Question_: I'm not a big fan of specific tests being os-dependent. Test should also indicate issues on OS' different from the one you are developing on. Is there anything in the config that makes this necessary? Can't you just test different paths/URIs on all systems?

ssmifi (changes requested):
The Windows detection looks a bit brittle, and I’m not sure we need OS-specific logic at all. Please check. Also, it would be great if the tests weren’t OS-dependent.

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: I can second the drive letter check, should definitely be an `[a-z]`. But we already do OS detection in the UriUtils, specifically there's an `isWindows` check that we can rewrite & export so we don't duplicate that. As for here it looks like you're using this to determine that we're not just on windows but also an absolute path w/ a drive letter? If so we should probably rename the var here to reflect that a bit more.

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: Same thought here as well.

packages/language/src/preprocessor/instruction-interpreter.ts
montymxb: Can pull down the `isWindows` var to just before the check, no need to have it above.

Separately it looks like we could split the windows specific behavior off to make it easier to test (came up in a conversation with @ssmifi )

packages/language/src/utils/uri.ts
```diff
@@ -27,9 +27,11 @@ export namespace UriUtils {
   }
 
   export function relative(from: URI | string, to: URI | string): string {
-    const fromPath =
-      typeof from === "string" ? URI.parse(from).path : from.path;
-    const toPath = typeof to === "string" ? URI.parse(to).path : to.path;
+    let fromPath = typeof from === "string" ? URI.parse(from).path : from.path;
+    let toPath = typeof to === "string" ? URI.parse(to).path : to.path;
+    fromPath = fromPath.replace(/\\/g, "/");
+    toPath = toPath.replace(/\\/g, "/");
```
montymxb: Can pull this pattern out into a little function to avoid the logic duplication.

montymxb (commented):
General feedback. Overall we should chat about this tomorrow with regards to the windows/unix behavior. It's separate from the changes here, but impacts them still, since we have to account for both cases in numerous spots.

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: Got it! Gonna work on the existing method on UriUtils and then apply it there ;)

Wagner-Laranjeiras: About the question from Steven, I'll answer in a second on the comment above :)

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: That's way cleaner indeed. thanks!

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: I thought that using `currentUri` would be more reliable, and for absolute paths, the results are almost the same. At this point in code `filePath` had a value like `file:///c%3A/...`, which would ask for some treatment before being used.

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: I honestly didn't spend much time at this snippet, because it was working. I just let what was already there. But I'll look into it and also apply the early return you suggested.

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: _Edit:_ I just remembered why that was there and the reason is that the `.relative` function returns `../` in front of the path, for unix absolute paths. I'll maintain this for a bit, because changing this at the function directly breaks other tests.

packages/language/src/utils/uri.ts
Wagner-Laranjeiras: I'm aware of this, but there are also problems when we don't incorporate this snippet. We can discuss at the meeting soon ;)

packages/language/test/preprocessor/set-file-path.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      "pli",
+      1,
+      "DECLARE VARC FIXED;",
+    );
+    context = { currentUri: URI.parse(doc.uri) };
+
+    setFilePath(item, doc.uri, context);
+
+    const expected = toCanonicalPath(URI.file(absolutePath).fsPath);
+    expect(item.relativeFilePath).toBe(expected);
```
Lotes: Expected should be a string literal. This makes the test easier to read.
You do not have to executed the test to find out what expected is.

packages/language/test/preprocessor/set-file-path.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    workspaceRoot = path.resolve("/workspace");
+    await pluginConfig.init(workspaceRoot);
+
+    context = null;
+  });
+
+  test("returns absolute path when target file resolves to absolute (outside workspace rules)", async () => {
+    const absolutePath = path.resolve("/absolute-folder/absolute.pli");
+    await vfs.writeFile(URI.file(absolutePath), "DECLARE VARC FIXED;");
```
Lotes: Is the file content important? I assume no, it is not so relevant for the test to work.
You could make a constant `const NoContent = "";` and replace that with your contents.

packages/language/test/preprocessor/set-file-path.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    const doc = TextDocument.create(
```
Lotes: Is it even necessary to create documents and a VFS?
The subject to test is `setFilePath` in my eyes. Which does not seem to require that stuff (doc.uri -> just use uri).

EDIT: I see you are referencing a singleton `PluginConfigurationProviderInstance`. I wish we could avoid this...

Lotes: Singletons can be an anti-pattern. They introduce dependencies that make tests more complex. As we can see here... you could reduce the complexity by providing the instance through the context object. But let's do this maybe in a different PR, if it is too expensive to do.

packages/language/test/preprocessor/set-file-path.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  setPluginConfigurationProvider,
+} from "../../src/workspace/plugin-configuration-provider";
+import { URI } from "../../src/utils/uri";
+import path from "path";
+
+function toCanonicalPath(filePath: string): string {
+  let normalized = path.normalize(filePath).replace(/\\/g, "/");
+  // Windows: ensure capital drive letter
+  if (/^[a-z]:\//.test(normalized)) {
+    normalized = normalized[0].toUpperCase() + normalized.slice(1);
```
Lotes: How about a helper function `capitalize(str)`? I think you were using capitalization also in a other function in this PR. Improves the readability a bit.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2041,6 +2041,15 @@ async function runIncludeInstruction(
   }
 }
 
+/**
+ * Removes leading character and capitalizes the first letter of the resulting string.
+ * Typically used to normalize Windows drive letters (e.g., "/c/path" → "C/path").
```
msujew: Minor: This comment doesn't seem right to me. There is a `:` missing after the `/c` and `C`

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2049,30 +2058,33 @@ async function runIncludeInstruction(
  * @param filePath File path to set, if null, no changes are made
  * @param context Used for currentUri to calculate relative paths
  */
-function setFilePath(
+export function setFilePath(
```
msujew: Suggestion: I wouldn't recommend exporting and testing this function directly due to all of its dependencies (`item`, `InterpreterContext`, `PluginConfigurationProviderInstance`). Instead, we could have some function in the `UriUtils` like `relativeDisplayPath(from, to)` that does of this work internally. I.e. compute a relative path if `to` is a subpath of `from`, and return an absolute path (formatted to the current OS) otherwise. This is cleaner and would make testing the code much easier.

Ideally, the function would also accept an optional `isWindows` argument (defaults to `UriUtils.isWindows`), which would make testing this even easier.

packages/language/test/preprocessor/set-file-path.test.ts
msujew, line 1: Suggestion: Please also test the Windows specific code.

packages/language/src/utils/uri.ts
```diff
@@ -19,43 +19,127 @@ export namespace UriUtils {
...
+  }
+
+  // In our test environment, sometimes the toPath is evaluated as a single slash ("/"),
+  // which can break the logic and return a relative path when an absolute should be given.
+  function normalizeForWindowsTests(path: string) {
+    if (path.startsWith("/")) {
+      path = path.substring(1);
+    }
+    return path;
+  }
```
msujew, line 113: Suggestion: This looks like a code smell. If we need specific code for our tests to work on Windows, that likely means that our virtual file system information implementation isn't giving back realistic data for Windows - which makes sense, since we're using Unix style paths on a Windows machine.

Ideally, this should be fixed by supplying the full information to the specific functions, i.e. whether we expect a Windows or a Unix style path and not let the system provide us with this information.

packages/language/test/preprocessor/set-file-path.test.ts
Wagner-Laranjeiras: Thanks Markus, I'm working on all suggestions you made regarding the tests. Still some work to do, but I was able to make it leaner. Indeed, a lot of what was there wasn't being used (I thought that I should make the testing environment as "real" as possible, but that turned out unnecessary). 😄

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: Got it! Thanks for the suggestion! I was able to work on it, it should be ok now.

packages/language/test/preprocessor/set-file-path.test.ts
Wagner-Laranjeiras: I'm trying to make the tests as agnostic as possible (although I'm implementing the changes you guys pointed out ;) ), and testing also manually on a Windows VM. But did you have another thing in mind regarding it?

packages/language/src/utils/uri.ts
Wagner-Laranjeiras: This is most certainly a code smell. I'll keep trying to solve this without the function. Thanks! 😄

packages/language/test/preprocessor/set-file-path.test.ts
msujew: If there is code that expects some `C:` prefix to be executed, we should have a test that includes a `C:` prefix to make sure that the code behaves as expected.

packages/language/src/utils/uri.ts
```diff
@@ -19,43 +20,163 @@ export namespace UriUtils {
...
```
msujew, line 113: Please add a `TODO` here and create a new issue to eventually refactor the URI handling in the project. I believe we've outgrown `vscode-uri`.

packages/language/test/preprocessor/set-file-path.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  let workspaceRoot: string;
+
+  beforeEach(async () => {
+    pluginConfig = new PluginConfigurationProvider();
+    setPluginConfigurationProvider(pluginConfig);
+
+    workspaceRoot = path.resolve("Users/mockUser/workspace");
+    await pluginConfig.init(workspaceRoot);
+  });
+
```
msujew: Suggestion: I don't think this setup code is required anymore?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2163,25 +2163,17 @@ function setFilePath(
...
+  item.filePath = filePath;
+  if (!context.currentUri) return;
+  const workspace = PluginConfigurationProviderInstance.getWorkspacePath();
+  const isWindows = UriUtils.isWindows;
+  item.relativeFilePath = UriUtils.relativeDisplayPath(
+    workspace,
+    filePath,
+    context.currentUri.path,
+    isWindows,
+  );
```
msujew: Minor: We can simply use the default argument here.

```suggestion
  item.relativeFilePath = UriUtils.relativeDisplayPath(
    workspace,
    filePath,
    context.currentUri.path,
  );
```

msujew (approved):
Alright, I think this is good to go. Please address the comments below before merging.

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: Thanks Mark! Gonna work on those and push again.

ssmifi (approved):

PR #507, opened 2025-12-03
Avoid infinite loop in tokenizer

msujew:
Improves our tokenizer to (a) prevent infinite loops (b) to improve lexing of unterminated string literals.

### DIFF
--- a/packages/language/src/parser/preprocessor-parser.ts
+++ b/packages/language/src/parser/preprocessor-parser.ts
@@ -1576,15 +1576,13 @@ function stringLiteral(state: ParserState): ast.Literal {
 }
 
 function unpackCharacterValue(literal: string): string {
-  let length = literal.length;
-  // Remove ending quotes if not escaped
-  if (
-    (literal.endsWith('"') || literal.endsWith("'")) &&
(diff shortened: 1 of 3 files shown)


packages/language/src/parser/preprocessor-parser.ts
```diff
@@ -1576,7 +1576,15 @@ function stringLiteral(state: ParserState): ast.Literal {
 }
 
 function unpackCharacterValue(literal: string): string {
-  return literal.substring(1, literal.length - 1).replace(/""/g, '"');
+  let length = literal.length;
+  // Remove ending quotes if not escaped
+  if (
+    (literal.endsWith('"') || literal.endsWith("'")) &&
+    literal.charAt(length - 2) !== "\\"
```
Lotes: Ah! Not sure about this one, some languages escape using backslash, some other languages make a double quote `""`. Was not sure if this is a thing in PL/I

msujew: Right, I've adjusted this again. We no longer actually generate a token for invalid strings (but still generate the error).

Lotes (approved):

PR #508, opened 2025-12-03
[Improvement] Config creation via Info dialogue

Wagner-Laranjeiras:
**Problem**

The problem starts when we change the plugin’s config folder. As expected, the includes break. When we then create a new config folder through the pop-up, it is created from a different entry point (the file we use as our program). After creating the extension via the quick fix (from the original entry point), the configuration is updated and the problem goes away.

**Changes**

1. Change listened event from `.workspace.onDidOpenTextDocument` to `.window.onDidChangeActiveTextEditor` (files open to change in focus)

2. Set up a "Don't Show Again" option for the Info pop-up.

3. More detailed Info pop-up message, which now specifies the current entry point.


packages/vscode-extension/src/extension/main.ts
```diff
@@ -47,73 +47,86 @@ export function activate(context: vscode.ExtensionContext): void {
...
-
-    // create the .pliplugin folder and files, using the current file as the entry point
-    fs.mkdirSync(plipluginPath);
-    fs.writeFileSync(
-      path.join(plipluginPath, "pgm_conf.json"),
-      JSON.stringify(
-        {
-          ...PluginConfiguration.DEFAULT_PROGRAM_FILE_CONTENT,
-          pgms: [
+      if (userResponse === "Don't show again") {
```
montymxb: We should at least pull this out as a constant so we don't add a typo into it later by accident.

packages/vscode-extension/src/extension/main.ts
```diff
@@ -47,73 +47,86 @@ export function activate(context: vscode.ExtensionContext): void {
...
 ) {
-  const listener = vscode.workspace.onDidOpenTextDocument(async (document) => {
-    // settle on the 1st workspace folder available
-    // TODO @montymxb May 15th, 2025: Support configs across multiple workspace folders
-    const workspaceFolder = vscode.workspace.workspaceFolders?.[0]?.uri.fsPath;
-    if (!workspaceFolder) {
-      return;
-    }
+  const listener = vscode.window.onDidChangeActiveTextEditor(
+    async (document) => {
```
montymxb: We should call this `editor`, `textEditor`, or something in that vein; but it's definitely not a document. I was wondering when I saw `document.document` below 🤔 .

packages/vscode-extension/src/extension/main.ts
```diff
@@ -47,73 +47,86 @@ export function activate(context: vscode.ExtensionContext): void {
 }
 
 /**
- * Listen for file open events, and prompt if we can create a .pliplugin folder
+ * Listen for changes on file activation, and prompt if we can create a .pliplugin folder
  * @returns Disposable listener
  */
+let shouldShowInfoMessage = true;
```
montymxb: This is just for as long as we're running this extension. I'm not sure if that's intended to be like this, or if we want to persist this across each instance of running the extension, otherwise the user will be prompted at the start each time. Maybe that's desired, just double checking.

montymxb (commented):
Changes look good initially but a couple of extra points.

- When the extension is started no notification is shown for the currently open file, not until we switch editors. 
- Until "don't show again" or "yes" is selected these messages tend to accumulate. It looks a bit clunky but makes sense given the user hasn't done anything with it yet. I don't know but is there a way to programatically dismiss any prior info message to avoid them stacking in that case?
- Preference doesn't persist across extension loads, it's tricky because it's necessary to handle pli files correctly, but it also feels like "don't show again" is more ephemeral than it implies it is.

packages/vscode-extension/src/extension/main.ts
Wagner-Laranjeiras: Got it!

packages/vscode-extension/src/extension/main.ts
Wagner-Laranjeiras: Okay, that makes sense. Gonna change to `textEditor` then. Thanks!

packages/vscode-extension/src/extension/main.ts
Wagner-Laranjeiras: Yep, that crossed my mind, but I let it like this, just because I can't think of an intuitive way for the user to get the pop-up again if they later change their mind, if they later want the pop-up. 

Would be great to offer a more severe "Don't show again" without ruining the possibility of having it back though.

Wagner-Laranjeiras:
Regarding the other comments:

> - When the extension is started no notification is shown for the currently open file, not until we switch editors.
> - Until "don't show again" or "yes" is selected these messages tend to accumulate. It looks a bit clunky but makes sense given the user hasn't done anything with it yet. I don't know but is there a way to programatically dismiss any prior info message to avoid them stacking in that case?

1. Thanks for pointing that out. I will bring back the listener for opening documents (I was just worried with the cost of it in terms of performance, so I just replaced it).

2. Yes, it's indeed a bit of annoying. I'll look into it and see if vscode offers a way of solving this. 

Thanks for the review Ben!

packages/vscode-extension/src/extension/main.ts
```diff
@@ -44,46 +44,73 @@ export function activate(context: vscode.ExtensionContext): void {
...
 
 /**
- * Listen for file open events, and prompt if we can create a .pliplugin folder
+ * Listen for changes on file activation, and prompt if we can create a .pliplugin folder
  * @returns Disposable listener
  */
-function registerOnDidOpenTextDocListener(
+let shouldShowInfoMessage = true;
+namespace PopUpOptions {
+  export const CONFIG_CREATION = {
```
ssmifi: _Minor_: If `CONFIG_CREATION` is only used in `handleMissingConfig`, you could declare it locally in that function and drop the namespace. If we end up with more popups later, we can always extract a dedicated `popups` file then.

packages/vscode-extension/src/extension/main.ts
```diff
@@ -44,46 +44,73 @@ export function activate(context: vscode.ExtensionContext): void {
...
+    textEditor.document.fileName,
+  );
 
-    // create the .pliplugin folder and files, using the current file as the entry point
+  // Not a Pli document or Config exists or Telemetry Event description not provided
+  if (!isPliDocument || fs.existsSync(plipluginPath) || !telemetryEvent) {
+    return;
+  }
+
+  sendTelemetryEvent(telemetryEvent, telemetryReporter);
```
ssmifi: Logically, the `sendTelemetryEvent` is not really connected to `handleMissingConfig`, correct? It would be nice if the functions actually only do their intended job and preferably only one job.
Also, changing the event to `onDidChangeActiveTextEditor` actually changes the semantics of the telemetry, right?

packages/vscode-extension/src/extension/main.ts
```diff
@@ -44,46 +44,73 @@ export function activate(context: vscode.ExtensionContext): void {
...
+
+  const userResponse = await vscode.window.showInformationMessage(
+    PopUpOptions.CONFIG_CREATION.infoTitle(currentFileRelativePath),
+    ...PopUpOptions.CONFIG_CREATION.userOptions,
+  );
+
+  // If user clicks "Don't show again"
+  if (userResponse === PopUpOptions.CONFIG_CREATION.userOptions[2]) {
```
ssmifi: _Minor_: Probably not a big concern here, but changing the order of the options will silently break the logic here. I believe you could just add the options independently:

>  YES: "Yes" as const,
  NO: "No" as const,
  DONT_SHOW_AGAIN: "Don't show again" as const,

> const userResponse = await vscode.window.showInformationMessage(
  PopUpOptions.CONFIG_CREATION.infoTitle(currentFileRelativePath),
  PopUpOptions.CONFIG_CREATION.YES,
  PopUpOptions.CONFIG_CREATION.NO,
  PopUpOptions.CONFIG_CREATION.DONT_SHOW_AGAIN,);

packages/vscode-extension/src/extension/main.ts
```diff
@@ -110,10 +138,24 @@ function registerOnDidOpenTextDocListener(
...
-      "'.pliplugin' folder and files created successfully.",
+      PopUpOptions.CONFIG_CREATION.successMessage,
+    );
+  }
+}
+
+function registerOnDidOpenTextDocListener(
+  telemetryReporter: TelemetryReporter | undefined,
+) {
+  const listener = vscode.window.onDidChangeActiveTextEditor(async (editor) => {
```
ssmifi: This is misleading: `registerOnDidOpenTextDocListener` actually registers `onDidChangeActiveTextEditor`.

ssmifi (changes requested):
I believe this unintentionally changes the meaning of the telemetry we emit. Since the listener now triggers on editor activation instead of document opening, the event no longer matches its previous semantics. 

However, I'm not quite convinced that the previous version is correct either: `pli.language.support.documentOpened` was only sent for PLI documents *without* a `.pliplugin` folder, so it was effectively tracking “PLI document opened with missing config”, not “PLI document opened” in general.

Could we:
- clarify what the intended semantics of this event are (document opened vs editor activated vs missing config), and
- adjust the listener, condition, and/or event name so they line up?

packages/vscode-extension/src/extension/main.ts
```diff
@@ -37,53 +37,81 @@ export function activate(context: vscode.ExtensionContext): void {
...
 
 /**
- * Listen for file open events, and prompt if we can create a .pliplugin folder
+ * Listen for changes on file activation, and prompt if we can create a .pliplugin folder
  * @returns Disposable listener
  */
-function registerOnDidOpenTextDocListener(
+let shouldShowInfoMessage = true;
+
+async function handleMissingConfig(
```
montymxb: Would be good to add a function header to document this one.

In addition this feels like it's handling two concerns at once, prompting the user for resolving the missing config + handling telemetry. It also seems like what @ssmifi pointed out below is valid, meaning we should reinstate the `onDidOpenTextDocument` listener to report the event separately there.

Assuming that's changed, we probably don't need any telemetry related stuff in this function, unless of course config related details are needed for reporting.

packages/vscode-extension/src/extension/main.ts
```diff
@@ -110,10 +139,22 @@ function registerOnDidOpenTextDocListener(
...
+}
+
+function registerOnDidChangeActiveTextEditor(
+  telemetryReporter: TelemetryReporter | undefined,
+) {
+  const listener = vscode.window.onDidChangeActiveTextEditor(async (editor) => {
+    await handleMissingConfig(
+      editor,
+      telemetryReporter,
+      "pli.language.support.documentAtivated",
```
montymxb: Typo on `documentAtivated`, and not sure if this one was requested specifically? If so then we should correct it, but if not then I would probably remove it.

montymxb (commented):
A couple more points here too. Feels like the telemetry should be restored & the new telemetry event dropped most likely (unless that was discussed separately, then feel free to disregard this).

Overall feels like we can discuss this tomorrow as well.

packages/vscode-extension/src/extension/main.ts
ssmifi: I checked the previous commits. The document opened telemetry event was guarded by the plugin logic since its addition. So, we should confirm the unintentional behavior with @asatklichov and then fix it. I opened https://github.com/zowe/zowe-pli-language-support/issues/521 to track it. 

For your PR @Wagner-Laranjeiras, please just handle it as before and leave it in the `onDidOpenTextDocument` event. Thanks.

ssmifi (commented):
Polished the last version a bit and restored the original semantics of the telemetry event (https://github.com/zowe/zowe-pli-language-support/issues/521). Please verify.

ssmifi (commented):
Other changes LGTM. In a follow up, I suggest to persist the user's choice. The popup could be activated again via a command/setting, after switching it off .

packages/vscode-extension/src/extension/main.ts
```diff
@@ -37,19 +37,25 @@ export function activate(context: vscode.ExtensionContext): void {
...
   const workspaceFolder = vscode.workspace.workspaceFolders?.[0]?.uri.fsPath;
   if (workspaceFolder) {
     watchPlipluginFolder(client, workspaceFolder, context, telemetryReporter);
   }
+
+  void handleMissingConfig(vscode.window.activeTextEditor);
+}
+
+function registerOnDidChangeActiveTextEditor() {
+  const listener = vscode.window.onDidChangeActiveTextEditor(async (editor) => {
```
Lotes: Instead: `return vscode.window. ...`.

I associate the callback here as a listener (the first argument). Your variable is a subscription. A subscription has the power to unregister the listener again :) .

packages/vscode-extension/src/extension/main.ts
```diff
@@ -63,27 +69,69 @@ function registerOnDidOpenTextDocListener(
...
+/**
+ * Listen for changes on file activation, and prompt if we can create a .pliplugin folder
+ * @returns Disposable listener
+ */
+let shouldShowInfoMessage = true;
 
-    if (userResponse !== "Yes") {
-      return;
-    }
+async function handleMissingConfig(textEditor: vscode.TextEditor | undefined) {
```
Lotes: I would prefer to have this one responsibility of asking and creating a config folder into a own file like `missing-config-handler.ts`. As far as possible.
The main file might get more tasks of registering and doing stuff, so it would be an opportunity to keep this file slim and readable through proper chosen names.

packages/vscode-extension/src/extension/main.ts
```diff
@@ -63,27 +69,69 @@ function registerOnDidOpenTextDocListener(
...
+  if (!isPliDocument || fs.existsSync(plipluginPath)) {
+    return;
+  }
+
+  const currentFileRelativePath = path.relative(
+    workspaceFolder,
+    textEditor.document.fileName,
+  );
+
+  const options = {
```
Lotes: Nice options record.

packages/vscode-extension/src/extension/main.ts
```diff
@@ -63,27 +69,69 @@ function registerOnDidOpenTextDocListener(
...
+    DONT_SHOW_AGAIN: "Don't show again",
+    YES: "Yes",
+    NO: "No",
+  } as const;
+
+  const userResponse = await vscode.window.showInformationMessage(
+    `Create a '.pliplugin' folder in the project root using '${currentFileRelativePath}' as the entry point in 'pgm_conf.json'?`,
```
Lotes: I would guess the user needs to be a kind of PL/I expert to understand this message. Or is this config setup only specific for our IDE/Extension pair?

Maybe a more readable message could be: "There is no startup configuration. Want to create one with ${...} as entry point?"

packages/vscode-extension/src/extension/main.ts
Lotes: *Reopening for a moment*

Meanwhile: was there not the proposal by Azat to add a Quickfix for the missing folder? That could be the intuitive path. Then you could persist this flag in our extension config. WDYT?

packages/vscode-extension/src/extension/main.ts
Wagner-Laranjeiras: Good point. Done it!

packages/vscode-extension/src/extension/main.ts
Wagner-Laranjeiras: That's another good point. Was also what Steven suggested. Gonna work on this for this PR.

Wagner-Laranjeiras:
Please do not review it yet - will implement users choice persistence

packages/vscode-extension/src/common/missing-config-handler.ts
Lotes, line 1: Thanks for making this file.

The current filename starts with a verb. Which breaks an implicit convention (we use nouns normally, I think). I would suggest `missing-config-handler.ts`.
But I would not insist on this change. It is a minor "mimimi" ^^* ...

Lotes (approved):

ssmifi (approved):

Issue #333, opened 2025-08-14
Builtins for CICS


asatklichov:
Currently we are only supporting core-PL/I-language built-ins. 
Let's add support for DB2 and CICS built-ins. 

e.g.  DFHRESP(NORMAL) or DFHEIBLK  

https://www.ibm.com/docs/en/cics-ts/6.x?topic=components-built-in-functions 

<img width="461" height="131" alt="Image" src="https://github.com/user-attachments/assets/ddc7ea01-bd53-4cf6-8fad-441c62485102" />

For more info 
https://www.ibm.com/docs/en/cics-ts/6.x?topic=available-documentation-in-pdf
https://www.ibm.com/docs/en/cics-ts/5.6.0?topic=code-how-use-resp-resp2-options
https://www.ibm.com/docs/en/epfz/6.1.0?topic=functions-built-in

PR #509, opened 2025-12-03
Support `DFHRESP` statement

msujew:
Related to https://github.com/zowe/zowe-pli-language-support/issues/333

Does not implement the full suite of CICS builtins. I have something separate planned for that.

### DIFF
--- a/packages/language/test/fourslash/preprocessor/cics/dfhresp-disabled.ts
+++ b/packages/language/test/fourslash/preprocessor/cics/dfhresp-disabled.ts
@@ -13,4 +13,6 @@
 
 //// DFHRESP(DISABLED)
 
+// Expect that the DISABLED code map to the correct number
+// Note: Applies to all other valid codes as well
 preprocessor.expectTokens("84");
--- a/packages/language/test/fourslash/preprocessor/cics/dfhresp-disabled.ts
+++ b/packages/language/test/fourslash/preprocessor/cics/dfhresp-disabled.ts
@@ -1,0 +1,16 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../../framework.ts" />
+
+//// DFHRESP(DISABLED)
+
+preprocessor.expectTokens("84");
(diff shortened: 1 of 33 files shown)


packages/language/test/fourslash/preprocessor/cics/dfhresp-disabled.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../../framework.ts" />
+
+//// DFHRESP(DISABLED)
+
+preprocessor.expectTokens("84");
```
ssmifi, line 18: _Suggestion_: The expected tokens are not obvious. Could we make the `CicsResponseCode` accessible via the fourslash constants?

packages/language/test/fourslash/preprocessor/cics/parse-fail-dfhresp-only-keyword.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../../framework.ts" />
+
+//// DFHRESP
+
+// The DFHRESP keyword on its own should be treated as a normal token
+preprocessor.expectTokens("DFHRESP");
```
ssmifi, line 17: _Question_: This would result in a parser error later on, right, which would be ok in the missing close case, I suppose? However, in this case, shouldn't we provide a more concrete diagnostic? Same for "wrong code"?

ssmifi (commented):
Implementation looks good, thanks. I have two small questions about the testing/diagnostics.

packages/language/test/fourslash/preprocessor/cics/dfhresp-disabled.ts
msujew: We could, but then we don't really test the codes themselves. I.e. we (or an AI agent, looking at you GH Copilot, which constantly tried to change the enum values) could accidentally change the value of one `CicsResponseCode` and we wouldn't notice it anymore. This way, the test asserts that the exact expected string value is produced when executing the statement.

packages/language/test/fourslash/preprocessor/cics/parse-fail-dfhresp-only-keyword.ts
msujew: Good point. I'll add a parser error diagnostic for incorrect codes. Note that the compiler emits something like `DFHRESP` as a token if simply used as a variable. Only in the `DFHRESP(<code>)` syntax it will actually emit the respective code string.

packages/language/test/fourslash/preprocessor/cics/dfhresp-disabled.ts
ssmifi: I see. If you're nice, you could leave a comment that states that this is testing the actual cics code, but not necessary. :)

packages/language/test/fourslash/preprocessor/cics/parse-fail-dfhresp-only-keyword.ts
ssmifi: Ok, as this is behaving like the real compile, I suppose it is ok. Maybe we should keep in mind (or track) that we potentially could provide better diagnostics than the real compiler here.

ssmifi (approved):
Thank you!

Issue #311, opened 2025-07-29
[Discrepancy] % in do-types


ssmifi:
In 

```
 RGT005: PROCEDURE() OPTIONS(MAIN); 
   %DCL I FIXED;
   %DCL X CHAR;
   %I = 0;
   %X = "";
   %DO WHILE(I <= 3); 
     %X = X || "A";
     %I = I + 1;
   %END;
   DCL Variable%;X FIXED;
 END RGT005;    
```  
`%DO WHILE(I <= 3);` is accepted but `%DO %WHILE(I <= 3);` is not.

_Note: I will check the other do-types as well_

ssmifi:
Former result regarding the requirement of `%` are also true for
`%DO FOREVER;` vs `%DO %FOREVER;`
`%DO UNTIL;` vs `%DO %UNTIL;`
`%DO WHILE(I > 0) UNTIL(I > 3);` vs `%DO %WHILE(I > 0) %UNTIL(I > 3);`
In each case, the latter is not valid syntax.

However, with `%loop: DO;`
`%IF I < 3 THEN GO TO loop;`
`%IF I < 3 %THEN GO TO loop;`
`%IF I < 3 %THEN %GO TO loop;`
`%IF I < 3 %THEN %GOTO loop;`
`%IF I < 3 THEN %GOTO loop;`
`%IF I < 3 THEN %GO TO loop;`
all work.

Furthermore,
`%loop: DO;` and `%loop:; %DO;` both produce `DCL VariableAAA FIXED;` 
but our implementation produces `VariableAAA` and `VariableA`, respectively.


#### Further Tests regarding `skip` in do loops:

All 
`%DO SKIP; WHILE(I <= 3);` 
`%DO SKIP; FOREVER;` 
`%DO SKIP; UNTIL(I > 3);` 
`%DO SKIP; WHILE(I >= 0) UNTIL(I > 3);` 
`%DO SKIP; UNTIL(I > 3) WHILE(I >= 0);` 
are valid and produce `DCL Variable FIXED;`

Furthermore,
`%DO WHILE(I > 0); SKIP; UNTIL(I > 3);` does also produce `DCL Variable FIXED;`
but 
`%DO WHILE(I >= 0); SKIP; UNTIL(I > 3);` is invalid syntax. 

This suggests that the pre-processor code is interpreted as it goes and non-pre-processor code is ignored. 
`%DO SKIP; xyz;` is actually valid,
[...]

PR #514, opened 2025-12-08
Fix IF and SELECT pp parser

### LINKED ISSUE #513: [Parser] `%` are sometimes optional in preprocessor
When the compiler explicitly expects another preprocessor unit, the `%` token becomes optional. Examples:

```
%IF 0 = 1 %THEN DO; // <- Would usually require a %
  PUT("HELLO");
%ELSE DO; // <- Would usually require a %
  PUT("WORLD");
%END;
```

This seemingly only applies to the true/false branches of if statements and the when/otherwise cases of the select statement.

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/513

Fixes two main points of our preprocessor parser:
1. "simple" if statements such as `%IF 1 %THEN; <emit some code>` had the wrong semantics. Essentially the `;` after `%THEN` acts as a null-statement, after which normal PLI code is supposed to go. This change adjusts the behavior to only allow "real" preprocessor statements to follow the `%THEN` keyword in the context of the if statement. (this one is difficult to explain)
2. The compiler makes the **first** `%` after the `%THEN` optional. Meaning that `%IF ... %THEN DO; ... %END;` and `%IF ... %THEN %DO; ... %END;` are the same. Adding the `%` is fine with the compiler, but it is not required.

### DIFF
--- a/packages/language/src/parser/preprocessor-parser.ts
+++ b/packages/language/src/parser/preprocessor-parser.ts
@@ -69,7 +69,7 @@ export function statement(
       return commonStatement(state, {
         withEnd: end,
         endPercent: endP,
-        startPercent: false,
+        startPercent: false, // Percent token already consumed
         labels: true,
       });
     } else {
@@ -80,7 +80,7 @@ export function statement(
     return commonStatement(state, {
       withEnd: end,
       endPercent: endP,
-      startPercent: false,
+      startPercent: false, // Inside of a procedure, percent not required
       labels: true,
     });
   }
@@ -1061,10 +1061,10 @@ function selectStatement(state: ParserState): ast.SelectStatement {
     );
   }
   state.consume(statement, CstNodeKind.SelectStatement_Semicolon0, t.Semicolon);
-  while (state.canConsumeKeyword(t.WHEN)) {
+  while (state.canPercentConsume(t.WHEN)) {
     statement.cases.push(whenStatement(state));
   }
-  if (state.canConsumeKeyword(t.OTHERWISE)) {
+  if (state.canPercentConsume(t.OTHERWISE)) {
     statement.cases.push(otherwiseStatement(state));
   }
   // END statement is preceded by a percent
@@ -1075,7 +1075,7 @@ function selectStatement(state: ParserState): ast.SelectStatement {
 
 function whenStatement(state: ParserState): ast.WhenStatement {
   const when = ast.createWhenStatement();
-  state.consumeKeyword(when, CstNodeKind.WhenStatement_WHEN, t.WHEN);
+  state.percentConsume(when, CstNodeKind.WhenStatement_WHEN, t.WHEN);
   state.consume(when, CstNodeKind.WhenStatement_OpenParen, t.OpenParen);
   let exp = expression(state);
   if (exp) {
@@ -1101,7 +1101,7 @@ function whenStatement(state: ParserState): ast.WhenStatement {
 
 function otherwiseStatement(state: ParserState): ast.OtherwiseStatement {
   const otherwise = ast.createOtherwiseStatement();
-  state.consumeKeyword(
+  state.percentConsume(
     otherwise,
     CstNodeKind.OtherwiseStatement_OTHERWISE,
     t.OTHERWISE,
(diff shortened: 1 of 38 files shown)


ssmifi:
I remember testing % earlier this year w.r.t. the do loops and logged some findings regarding %IF here https://github.com/zowe/zowe-pli-language-support/issues/311, esp. % being optional in several cases. Did you see/check these cases as well?
In this case, probably mainly if the % before the THEN is optional or not.

msujew:
> I remember testing % earlier this year w.r.t. the do loops and logged some findings regarding %IF here https://github.com/zowe/zowe-pli-language-support/issues/311, esp. % being optional in several cases.

Ah, I see, that's good to know.

> In this case, probably mainly if the % before the THEN is optional or not.

Thanks for pointing that out. You're right, it looks like the `%` is optional.

msujew:
@ssmifi I've also made the `%` for `THEN` optional. Thanks for the hint!

packages/language/src/parser/preprocessor-parser.ts
```diff
@@ -97,14 +126,25 @@ function labels(state: ParserState): ast.LabelPrefix[] {
...
+  options: StatementParseOptions & { withEnd: false },
+): ast.Statement | null;
+export function commonStatement(
+  state: ParserState,
+  options: StatementParseOptions,
+): ast.Statement | ast.EndStatement | null;
+export function commonStatement(
+  state: ParserState,
+  options: StatementParseOptions,
 ): ast.Statement | ast.EndStatement | null {
```
ssmifi, line 132: _Minor_: The JSDoc seems to be outdated.

packages/language/test/fourslash-harness/harness-parser.ts
```diff
@@ -71,7 +71,7 @@ class HarnessTestParser {
     private context: Context,
   ) {
     // Reverse it for performance reasons, faster to pop than shift
-    this.lines = text.split("\n").toReversed();
+    this.lines = text.split(/\r?\n/).toReversed();
```
ssmifi, line 74: _Question_: Is there a specific reason for this change? Would this allow all fourslash tests to be in CRLF format? There are specific tests that test for windows line endings with a \r escape character, e.g., in the margin preprocessor. Wouldn't it be less confusing if the tests stay in one format and explicitly test for different line endings if necessary (or automatically test for all line endings)?

packages/language/test/fourslash/preprocessor/select-statement-multiple-conditions.ts
```diff
@@ -12,8 +12,8 @@
 /// <reference path="../framework.ts" />
 
 //// %SELECT;
-//// %WHEN (0, 1, 2) X
-//// %OTHERWISE Y
+//// %WHEN (0, 1, 2) DO; X %END;
+//// %OTHERWISE DO Y %END;
```
ssmifi: _Question_: Is the semicolon at the DO missing on purpose? Does this work because of the recovery?

packages/language/test/fourslash/preprocessor/simple-if-else.ts
```diff
@@ -11,9 +11,11 @@
...
-//// %IF 0 %THEN
+//// %IF 0 %THEN DO;
 ////   dcl X fixed;
-//// %ELSE
+//// %END;
+//// %ELSE DO;
 ////   dcl Y fixed;
+//// %END;
 
 preprocessor.expectTokens(" dcl Y fixed;");
```
ssmifi: _Tiny_: We can remove the leading space here, can't we? (Same for -then -with-scan)

ssmifi (commented):
Thanks, looks good to me. I just have two questions about the testing.

packages/language/test/fourslash/preprocessor/select-statement-multiple-conditions.ts
msujew: No, not missing on purpose. Good catch :)

packages/language/test/fourslash-harness/harness-parser.ts
msujew: IMO, I don't think that the line ending should matter for running the test. I've noticed that if I create a new file on windows (which automatically gets CRLF in VSCode), my tests start failing unexpectedly, because the lines aren't split correctly. Note that special characters such as `\r` in the test code will actually be stored as `\\r` and not be subject to changes due to this change.

ssmifi (approved):
👍

Issue #511, opened 2025-12-05
Findings related with parser


asatklichov:
- [x] (1) Resolve false-positives:  green  code in editor: SQL builtins, host variables, CICS, etc 
 
- [x] (2) Host variables bug fixes related to (paranthesis, dimension range, )
 
- [x] (3) Missing implementation for host variables 'Result set locators' and 'Result set Table locators'

- [x] (4) Missing validations for (all host-variable arrays in PL/I) host variables  in dimension range. 
Will be done LATER.   

     ` DCL (c7(100), c8(400000)) Character(15); /*dimension range 1–32767*/`

- [x] (5) Grammar error in binary host-variable arrays 
               ` DCL b_ARRAY(100) SQL TYPE IS bin; /*ERR*/`
 
- [x] (6) Hover overs support (for SQLCA and SQLDA, struct variables)

- [x] (7) Bug fixes related to quick fix 

   a. once renaming .plugin folder quick fix does not resolve,)
   b. if copybook is not found in workspace, or defined absolute path error message should be 'NOT FOUND'  not 'Can not be opened' 
   c. if program uses variable from other copybooks, it resolves it as implicit variable,  but after a while, or after reloading it then recognizes it as declared variable 

- [x] (8) Large file

 
             a) include not processed as PL/I [single threaded issue] 
             b) slow has performance issue 

- [x] (9) Pop-up dialogue on copybooks should have a) correct PATH name b) column should be 72 character, lines are OK

 

- [x] (10) Issue with compound dataset resolution e.g. %include "DIM1" is not resolved same as %include DIM1 

- [x] ( 11) Code suggestions and completions does not work properly 

- [x] (12) Vertical Ruler at column 73 is LOST sometimes 11. 

- [x] (13) MSG: Duplicate compiler option found for <TOKEN>  causes/hides many other issues

- [x] (14) MSG: String options is not recognized as a valid option keyword and is ignored. Reproduced in multiple forms.

- [x] (15) Resolve below sql-includes as normal build-ins SQLCA, SQLDA in code.  
 
EXEC SQL INCLUDE SQLCA;
EXEC SQL INCLUDE SQLDA;

- [x]  (16) INPUT & OUTPUT attributes of dataset declaration produce improper errors

msujew:
@asatklichov Can you clarify point 5? I've simply followed the exact behavior of the compiler during the implementation:

Compiler input:

```
DCL b_ARRAY(100) SQL TYPE IS bin;
```

[...]

PR #515, opened 2025-12-10
Stabilize Include Resolution Ordering

montymxb:
Fixes some recently reported issues with the way we resolve member & file includes:

- update include resolution to resolve in a bfs fashion w/ alphabetical sorting per level
    - ex. `%INCLUDE "a.pli";` should resolve `cpy/a.pli` over `cpy/nested/a.pli`
    - also `%INCLUDE "a.pli";` should resolve `x/a.pli` over `y/a.pli`
    - works for members & files
    - refactored search for the vfs to leverage a sorted files list, maintaining our suffix matching approaches
- resolve members before files for file-based includes (resolves point 10 of #511)
    - ex. `%INCLUDE "m1";` will now work similarly to `%INCLUDE m1;`, favoring matching members first
- improve an existing test case to ensure recursive ddname resolution works as expected

There's some tests I've tossed in for these as well. The vfs was particularly problematic given the order of insertion changes the testing results here, which was the justification for the sorted files list. The Array sort operation should be using something like Tim sort or Quick sort internally, but crawling the array for suffix matches is still unchanged. So it may be an improvement to use a suffix tree later on for these cases.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2238,7 +2238,7 @@ function getFileNameOrPartialName(item: IncludeItem): string | undefined {
  * Attempts to resolve the URI of an include file factoring in process group libs, relative & absolute paths
  *
  * @param item Include item to resolve a URI for
- * @param state Current PP state, used to resolve relative paths, program configs, and report errors
+ * @param context Interpreter context used for resolution & diagnostic collection
  * @returns URI of the included file if found, otherwise undefined
  */
 async function resolveIncludeFileUri(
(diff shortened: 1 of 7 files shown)


ssmifi (approved):
LGTM, while you are on it, you could tidy up `resolveIncludeFileUri` a bit :)
  * The JSDoc param is wrong 
  * `@param` in `resolveLibFileUri` JSDoc is missing. 
  * Is your June TODO still valid? 
  * Exit early for `!pgroup`.

montymxb:
@ssmifi thanks for the input! I'll take care of that plus those other points too. Not sure about the June point however, I'll re-review to check if it's still relevant.

montymxb:
Updating with review points (just didn't push yet).

@ssmifi Not sure exactly which June todo you're referring to, but I believe it's `Disabled relative & absolute pathing per request...` in resolveIncludeFileUri? That one is still relevant given the behavior on the mainframe does work in this way (ex. using `%INCLUDE "./myfile.pli"` will bring up that file from the cwd. But it was removed by request as noted in the comment. We were waiting to hear back more concretely from upstream users as to whether this was going to be desired. It could probably be removed going forward, but I get the feeling that as soon as we do that somebody will ask for that very feature, just a feeling however.

PR #516, opened 2025-12-10
Generate variables for EXEC CICS statements

msujew:
As part of an effort to improve the PL/I tolerance, we need to ensure that the expected declaration statements are generated when running into an `EXEC CICS...` statement. The CICS preprocessor will then generate a bunch of declarations after the last `PROC` statement encountered. Similar to how the `SQL TYPE` attribute does it.

These variables are required for any PL/I code that has any interaction with CICS. Related to point 1 of https://github.com/zowe/zowe-pli-language-support/issues/511. This code simply mirrors the behavior of the CICS preprocessor when encountering a `EXEC CICS...` statement.

### DIFF
--- a/packages/language/src/syntax-tree/ast.ts
+++ b/packages/language/src/syntax-tree/ast.ts
@@ -225,6 +225,7 @@ export enum SyntaxKind {
   WriteStatementOption,
   XFormatItem,
   CicsResponseStatement,
+  CicsExecStatement,
 }
 
 export enum KeywordConditions {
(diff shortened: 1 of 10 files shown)


packages/language/src/syntax-tree/ast.ts
```diff
@@ -842,8 +843,9 @@ export type SyntaxNode =
   | Statement
   | StopStatement
   | StringLiteral
-  | SqlAttributeStatement
   | CicsResponseStatement
+  | CicsExecStatement
```
ssmifi: _Minor_: Order alphabetically. (Also Environment, Procedure... and CommonDeclarationAttribute, Unit from // Exclusive to preprocessor)

ssmifi (approved):
LGTM

PR #517, opened 2025-12-11
Improve parser error recovery

msujew:
I've noticed that we don't recover correctly if the parser fails during the lookahead of a new statement. I.e. with the input:

```
DEFINE TEST;
```

The parser expects either `ORDINAL`, `STRUCTURE` or `ALIAS`. If none of them appear, the parser does not advance, and instantly recovers, which leads to an infinite loop in the parser.

This change adjusts the recovery process to always skip at least one token, which ensures that we don't get caught in a loop. It also moves the recovery logic outside of the parser state.

### DIFF
--- a/packages/language/src/parser/parser-state.ts
+++ b/packages/language/src/parser/parser-state.ts
@@ -30,31 +30,33 @@ import {
 import { tokenIdxToClass } from "./token-type-factory";
 import * as environment from "../workspace/environment";
 
-export function preprocessorParserState(tokens: t.Token[]): ParserState {
-  return new ParserState(tokens, ParserStateMode.Preprocessor);
+export enum RecoveryResult {
+  /**
+   * Recover at the current token
+   */
+  Recover,
+  /**
+   * Recover after the current token
+   */
+  RecoverNext,
+  /**
+   * Continue the recovery process (the parser state will go on to the next token)
+   */
+  Continue,
 }
 
-export function finalParserState(tokens: t.Token[]): ParserState {
-  return new ParserState(tokens, ParserStateMode.Final);
-}
-
-export enum ParserStateMode {
-  Preprocessor,
-  Final,
-}
+export type RecoveryFunction = () => RecoveryResult;
 
 export class ParserState {
   readonly tokens: t.Token[];
-  readonly mode: ParserStateMode;
   readonly diagnostics: Diagnostic[];
   public index: number;
   public inError = false;
 
   private inProcedure = false;
 
-  constructor(tokens: t.Token[], mode: ParserStateMode) {
+  constructor(tokens: t.Token[]) {
     this.tokens = tokens;
-    this.mode = mode;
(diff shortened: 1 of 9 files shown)


packages/language/src/typesystem/primitive-type-builder.ts
```diff
@@ -740,41 +741,6 @@ export class DefaultPrimitiveTypeBuilder implements PrimitiveTypeBuilder {
       case ast.DefaultAttribute.RESERVED: //@see https://www.ibm.com/docs/en/epfz/6.1.0?topic=declarations-reserved-attribute
       case ast.DefaultAttribute.STRUCTURE: //@see https://www.ibm.com/docs/en/epfz/6.1.0?topic=definitions-defining-typed-structures-unions
       case ast.DefaultAttribute.VARIABLE: //@see https://www.ibm.com/docs/en/epfz/6.1.0?topic=attributes-variable-attribute
-      case ast.DefaultAttribute.BACKWARDS:
```
Lotes, line 743: Why the removal?

packages/language/src/parser/parser.ts
```diff
@@ -670,41 +670,26 @@ const statement = rule(
       element.value = unit.rule(state);
     }
 
-    recover(
-      state,
-      () =>
+    state.recover(() => {
```
Lotes: Make the lambda as an own function

packages/language/src/parser/parser-state.ts
```diff
@@ -182,14 +184,22 @@ export class ParserState {
     this.inError = true;
   }
 
-  recover(): void {
+  recover(fn: RecoveryFunction): void {
```
Lotes, line 187: Nice idea with the custom recovery function.

packages/language/src/parser/parser.ts
Lotes: It is quite long, the preprocessor implementation looks more compact, because of the separation.

packages/language/test/rules.test.ts
Lotes, line 1: Can we simulate the endless loop in the tests, here?

Lotes (commented):
I also like the cleaner implementation of the parser state, so it fits preprocessor and main parser :).

packages/language/test/rules.test.ts
Lotes: Or is the other Fourslash test enough?

packages/language/src/typesystem/primitive-type-builder.ts
msujew: Because they were duplicated. ESBuild actually complains about them, if you use the "Run selected harness test file" launch config.

packages/language/test/rules.test.ts
msujew: The other fourslash test is probably enough. I'm not sure how much benefit a more dedicated test has.

Lotes (approved):

PR #522, opened 2025-12-18
Support short builtin functions

msujew:
Does as the title says. We had some missing builtins, which this PR attempts to fix.


Lotes (approved):
Just an idea: We could add JSDocs with a "see" link pointing to the language spec. This way, we can track were the requirement is coming from.

PR #523, opened 2026-01-06
[Typesystem] Implement TYPE and LIKE attributes

Lotes:
* add support for `TYPE`and `LIKE` attributes
* add `UNION` data type
* add `parentType` as back reference to parent `STRUCTURE` and `UNION`
* add cycle detection during type inference
* add tests for different `STRUCTURE`/`UNION` scenarios with `LIKE` and `TYPE` attributes


packages/language/src/typesystem/composite-type-builder.ts
```diff
@@ -23,8 +23,20 @@ export class DefaultCompositeTypeBuilder implements CompositeTypeBuilder {
...
+  ): TypeDescriptions.Structure | TypeDescriptions.Union {
+    if (
+      declaredItem.attributes.some(
+        (attr) =>
+          attr.kind === ast.SyntaxKind.ComputationDataAttribute &&
+          attr.type === ast.DefaultAttribute.UNION,
+      )
+    ) {
+      return TypeDescriptions.Union({
+        level: declaredItem.level!,
```
msujew: Suggestion: I don't think we can enforce this. Return `1` if not set? We do this in a lot of linking code as well.
```suggestion
        level: declaredItem.level ?? 1,
```

packages/language/src/typesystem/descriptions.ts
```diff
@@ -1055,6 +1076,37 @@ function createStructureTypeDescription({
...
 }
 
+//--- Union ---
+const UnionType = DataType.Union;
+type UnionType = typeof UnionType;
+
+interface UnionTypeDescriptionProps {
+  level: number;
+  members: Record<string, TypeDescriptions.Any>;
+  membersMetadata: Record<string, BuilderDeclareItem>;
```
msujew: Question: What happens when encountering invalid types with duplicate names? Does it make sense to still generate the correct type for this?

```
DCL
  1 A,
    2 B,
      3 C,
    2 B,
      3 D;
```

Or to phrase my question differently: Can we model this as an array? Would this fix this issue?

packages/language/src/typesystem/descriptions.ts
```diff
@@ -328,9 +342,14 @@ interface WithTypeDescriminator {
   type: DataType;
 }
 
+interface WithParentType {
+  parentType?: TypeDescriptions.Structure | TypeDescriptions.Union; //TODO: TypeDescriptions.Union;
```
msujew: Is the TODO still relevant?

```suggestion
  parentType?: TypeDescriptions.Structure | TypeDescriptions.Union;
```

packages/language/src/typesystem/descriptions.ts
```diff
@@ -114,6 +116,10 @@ export enum AttributeKind {
   Scope,
   /** @see https://www.ibm.com/docs/en/epfz/6.1.0?topic=facilities-preprocessor-scan */
   ScanMode,
+  /** @see https://www.ibm.com/docs/en/epfz/6.1.0?topic=unions-like-attribute */
+  SetLike,
```
msujew, line 120: I've taken a look at this link and noticed a test case we might be missing:

> The following example is valid, but only because the LIKE references are expanded after they are all resolved, otherwise the reference aa3_array would be ambiguous:

```
  dcl 1 aa(30)                           
     ,5 aa1               char( 5)       
     ,5 aa2               fixed bin(31)  
     ,5 aa3_array(30)                    
       ,7 aa3_1           fixed dec(15,2)
       ,7 aa3_2           fixed dec(15,2)
       ,7 aa3_3           fixed dec(11,4)
       ,7 aa3_4           fixed dec(7,3) 
    ;  

  dcl bb                  like aa;          
  dcl cc                  like aa3_array;     
```

Can you add a test file for this that ensures that the `aa3_array` resolves correctly (and without an error)? Feel free to give the variables better names/simplify the test. Note that this might not work, in which case you can use the `*.fail.ts` syntax to mark the test file as failing.

msujew (commented):
Looking pretty good, though I didn't have time to thouroughly test this. Will do tomorrow :)

packages/language/src/typesystem/descriptions.ts
Lotes: Question: Do you want to query those duplicated members during the testing phase? Then I really have to use arrays.

Lotes: I used Maps from DeclaredVariable to a value. Is this enough?

packages/language/test/fourslash/validate/typesystem/like/like-structure-array.ts
```diff
@@ -0,0 +1,114 @@
...
+
+verify.noDiagnostics(undefined, ...code.TypeSystem);
+types.expectTypeAt("bb", {
+  type: types.dataTypes.Structure,
+  dimension: [
+    {
+      lowerBound: { value: 1 },
+      upperBound: { value: 30 },
+    },
+  ],
```
msujew: Suggestion: I don't think that this is actually correct. The documentation has this piece of information:

> The following declarations yield the same structure for X. 

```
  dcl
    1 A(10) aligned static,
      2 B    bit(4),
      2 C    bit(4),
    1 X like A;

 
  dcl
    1 X,
      2 B bit(4),
      2 C bit(4);
```

> Notice that the dimension (DIM(10)), ALIGNED, and STATIC attributes are not copied as part of the LIKE expansion.

Meaning that the dimension should not be copied.

packages/language/test/fourslash/validate/typesystem/typedefs/ordinal-definition.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../../../framework.ts" />
+
+// @wrap: main
+//// define ordinal Color ( Red, Green, Blue );
+//// dcl <|col|> type Color;
```
msujew, line 16: Suggestion: Also add a test for a `DEFINE STRUCTURE` statement:

```
DEFINE STRUCTURE 1 A, 2 B, 3 C CHAR(5);
DCL X TYPE A;
```

packages/language/test/fourslash/validate/typesystem/like/like-structure-array.ts
```diff
@@ -0,0 +1,114 @@
...
+/// <reference path="../../../framework.ts" />
+
+////  dcl 1 aa(30)
+////     ,5 aa1               char( 5)
+////     ,5 aa2               fixed bin(31)
+////     ,5 aa3_array(30)
+////       ,7 aa3_1           fixed dec(15,2)
+////       ,7 aa3_2           fixed dec(15,2)
+////       ,7 aa3_3           fixed dec(11,4)
+////       ,7 aa3_4           fixed dec(7,3)
```
msujew: Suggestion: Can we reduce the amount of variables in here? I think this leads to a bit of clutter in the test.

packages/language/src/typesystem/descriptions.ts
msujew: Yeah, I think that works well. Thank you!

Lotes:
There are two things open in my eyes:
```
DCL XXX TYPE A B.C.D; //b.c.d is a qualify block, but I could not find any examples
```

And:

```
DEFINE STRUCTURE 1 XXX, ...;
DCL YYY LIKE XXX.A.B.C;
```

But currently I could not connect to the mainframe to test the behavior.

Lotes:
> 
> ```
> DEFINE STRUCTURE 1 XXX, ...;
> DCL YYY LIKE XXX.A.B.C;
> ```
> 
> But currently I could not connect to the mainframe to test the behavior.

This example is not a valid program. A LIKE into a DEFINE STRUCTURE is not allowed.

packages/language/test/fourslash/linker/type-attribute/structure-items.ts
```diff
@@ -13,7 +13,7 @@
 
 // @wrap: main
 //// DEFINE STRUCTURE 1 A, 2 <|b:B|> FIXED(31);
-//// DCL MY_STRUCT TYPE A;
+//// DCL <|MY_STRUCT|> TYPE A;
```
msujew: Suggestion: I don't think this does anything?

msujew (approved):
Looks good to me, thanks 👍

PR #526, opened 2026-01-07
Improve caching of program configs

msujew:
This change caches the program config and process group config in the compiler unit to ensure that we don't have to look it up all the time. Especially during validation, which has to access those configs quite often. This should speed up performance for all use cases, but especially when dealing with large files.


ssmifi (approved):
LGTM

PR #527, opened 2026-01-07
Optimize tokenizer to use array

msujew:
Replaces the previous `Map` based implementation with an array. This provides a meaningful speed improvement to the lexer (roughly 10% in my tests) without changing the actual behavior.


ssmifi (approved):
Nice.

PR #529, opened 2026-01-08
Fix misplaced diagnostics in compiler options with plugin configs

### LINKED ISSUE #524: Findings  related to preprocessors
- [x] (1) compiler option validation is missing for missing parenthesis case for  PP( 

*PROCESS pp(include   /*Error4: no validation */

Image

 

 
- [x] (2) Below code misses validation  on %NOTE('aHOJ ', NAME);

```

 REPL: PROC;
   %REPLACE A BY 67;
   %REPLACE B BY 34;

   /*
   page 623 - Lang.Ref
   If code has a value other than those listed above, a diagnostic 
   message is produced and a default
  value is taken. If the value is less than 0 or greater than 16, severity
   U is the default. Otherwise, the next lower severity is the default.
   */
   %NOTE('A+B = ', (A+B));

   %REPLACE NAME WITH 'ENE';
   %NOTE('aHOJ ', NAME);
 END;

```

- [x] (3)  Analyze the feasibility for 

```
%process pp(macro('INCONLY')); 
 and 
 %process pp(macro('FIXED(DECIMAL)'));
```

ssmifi:
Fixes https://github.com/zowe/zowe-pli-language-support/issues/524 1

This error only occurs if there are diagnostics originating in the plugin configs. In this case, the `issueCount` in the programConfig might override the properties in the diagnostics of the source file. Hence, the source file diagnostics are misplaced and the config diagnostics are lost. 

The fix
1. copies the diagnostics from the config because we are changing their properties,
2. shows the config diagnostics at the first `*PROCESS` again,
3. and sets the correct severity for parser errors in the compiler options.

Also added a test to check correct placement of the diagnostics in this constellation.

On a side note: Is it (still) necessary to maintain both `compiler-options` and `pli-options` in the plugin config? For me, this was just another layer of confusion. If not, we should maybe settle on the more convenient syntax and drop the other one.

### DIFF
--- a/packages/language/src/workspace/plugin-configuration-provider.ts
+++ b/packages/language/src/workspace/plugin-configuration-provider.ts
@@ -11,6 +11,7 @@
 
 import { minimatch } from "minimatch";
 import { Diagnostic } from "vscode-languageserver-types";
+import { Diagnostic as CompilationUnitDiagnostic } from "../language-server/types";
 import { CompilerOptionResult } from "../preprocessor/compiler-options/options";
 import {
   AbstractCompilerOptions,
@@ -53,7 +54,7 @@ export interface ProgramConfig {
    * Number of issues found in the pli-options for this program config (which generate compiler options)
    * Used to avoid duplicate issue reporting later on when running translation in a program context
    */
-  issueCount?: number;
+  issues?: CompilationUnitDiagnostic[];
 }
 
 interface SerializedProgramConfig {
@@ -564,7 +565,7 @@ export class PluginConfigurationProvider {
         );
       }
       programConfig.abstractOptions = abstractOptions;
-      programConfig.issueCount = translatedOptions.issues.length;
+      programConfig.issues = translatedOptions.issues;
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -11,6 +11,7 @@
 
 import { minimatch } from "minimatch";
 import { Diagnostic } from "vscode-languageserver-types";
+import { Diagnostic as CompilationUnitDiagnostic } from "../language-server/types";
```
msujew, line 14: Minor: Can we instead do this?
```suggestion
import { Diagnostic as LspDiagnostic } from "vscode-languageserver-types";
import { Diagnostic } from "../language-server/types";
```

msujew (approved):
Looks good to me, just a minor issue, see below. Ideally, any future bug fixes that are related to this topic should instead refactor this part of the diagnostics generation in favor of https://github.com/zowe/zowe-pli-language-support/issues/535.

PR #531, opened 2026-01-08
Fix various parser issues

msujew:
Does as the title says:

1. There was an issue in the `canConsumeFirstItem` method: If we already were in an error, it didn't automatically return false, which led to infinite loops.
2. Contrary to the documentation, empty `ENTRY()` attributes (i.e. with empty `()`) are allowed by the compiler.
3. There is another documentation issue related to the descriptors within the `ENTRY` attribute, which has been fixed.
4. The `RESERVED` attribute has an optional `(IMPORTED)` suffix, which was previously ignored.
5. The preprocessor parser now properly supports unary expressions.
6. The preprocessor interpreter now supports all assignment operators (i.e. `+=`, `*=`, `||=`, etc.).


packages/language/src/parser/abstract-parser.ts
```diff
@@ -316,7 +316,6 @@ const binaryPrecedence = buildPrecendenceMap([
     BinaryOperator.LessThanEquals,
     BinaryOperator.Equals,
     BinaryOperator.NotEquals,
-    BinaryOperator.LessThanGreaterThan,
```
Lotes, line 319: Why was this removed? Is it not a binary op? Do we have to add it again with a different priority? If not, remove it also from the comment above this line.

Lotes: Or was it duplicated? Looks like it…

Lotes (approved):
LGTM

packages/language/src/parser/abstract-parser.ts
msujew: It's duplicated. `NotEquals` (`^=` in PLI) is the same as `LessThanGreaterThan` (`<>` in PLI). Has the exact same semantic meaning, it's just a different token.

PR #533, opened 2026-01-12
Fix preprocessor builtins path

msujew:
Fixes a minor bug that prevented opening the macro builtins file.


ssmifi (approved):
Fixed. Thanks.

PR #537, opened 2026-01-15
Hover support for builtins

Lotes:
* especially for builtins like `SQLCA` and `SQLDA`
* uses Type system information and convert it back to a string
* converting from the type information has the advantage that the user can reflect about the actual type information, since conflicting attributes get ignored


packages/language/test/fourslash/validate/typesystem/declare/dcl-bit-bit-triggers-error.ts
```diff
@@ -14,8 +14,11 @@
 // @wrap: main
 //// DCL <|1:ANYTHING|> BIT(10) <|2:BIT|>(11);
 
-verify.expectDiagnosticsAt(2, code.Error.IBM1309I);
+verify.expectDiagnosticsAt(2, code.Error.IBM2462I);
```
Lotes, line 17: Really?

packages/language/src/language-server/hover-request.ts
```diff
@@ -50,6 +49,37 @@ interface MarkupGenerator {
   (context: MarkupGeneratorContext): MarkupResponse;
 }
 
+function compositeToString(
```
Lotes: I would like to have those parts closer to the descriptions file as part of the toString method…

msujew (commented):
I thinks that mostly looks good - however, I'm also concerned about the `TODO` code. Haven't tested it, but please add automated for this case:

```
DCL 1 A, 2 B FIXED;
DCL 1 X, 2 Y LIKE A;
PUT(X.Y.<|>B);
```

I would expect that this shows the whole structure, i.e. `1 X, 2 Y, 3 B FIXED`? Or how does this work?

Generally, some additional tests using `LIKE` and `TYPE` + `STRUCT` would be appreciated!

packages/language/test/fourslash/hover/type-related/like-put-on-leaf-type-level-3.ts
```diff
@@ -0,0 +1,24 @@
...
+//// DCL 1 X, 2 Y, 3 Z LIKE A;
+//// PUT(X.Y.Z.<|1>B);
+
+hover.expectMarkdownAt(
+  1,
+  hover.codeBlock(`DCL 1 X,
+      2 Y,
+        3 Z,
+          4 B FIXED;`),
+);
```
msujew, line 24: Sorry to ask for another one :)

```
DCL 1 A, 2 B, 3 C FIXED;
DCL 1 X, 2 Y LIKE A;
PUT(X.Y.B.<|1>C);

Lotes: Thanks, will add this to the test suite. I already have a bunch of failing tests, but more can not hurt :D...

Lotes:
To answer your question, how does it work: Each type has a `variableNode` and a `parentType`... In situations with LIKE and TYPE I can not just assign the same type: I have to copy it, changing the variable node and  the parent chain.
But currently the code for the TODO is acting only for one specific case. I will find a general solution and add more test cases.

msujew (approved):
Looks good to me 👍

PR #539, opened 2026-01-19
Use one mutex per compilation unit

msujew:
Fixes an issue that pops up quite consistently when opening a workspace that has multiple files open on startup. Only one of the files will receive their semantic highlighting. Currently, this can also be observed on the playground for the same reason:

Since we only have one global mutex, opening multiple files at the same time will abort all build processes except for the last one that was triggered. Since each compilation unit can be built independently from one another, the obvious solution is to have one mutex per compilation unit.

This PR adapts the codebase to do exactly that.

Testing this isn't exactly straightforward - one easy test case is that the playground should be immediately highlighted. Do a bit of manual smoke testing to confirm that this change didn't break other behavior.

### DIFF
--- a/packages/language/src/language-server/connection-handler.ts
+++ b/packages/language/src/language-server/connection-handler.ts
@@ -9,14 +9,17 @@
  *
  */
 
-import { CompilationUnitHandler } from "../workspace/compilation-unit";
+import {
+  CompilationUnit,
+  CompilationUnitHandler,
+} from "../workspace/compilation-unit";
 import {
   CancellationToken,
   Connection,
   DocumentHighlight,
   TextDocumentSyncKind,
 } from "vscode-languageserver";
-import { URI, UriUtils } from "../utils/uri";
+import { URI } from "../utils/uri";
 import { definitionRequest } from "./definition-request";
 import { referencesRequest } from "./references-request";
 import { semanticTokenLegend, semanticTokens } from "./semantic-tokens";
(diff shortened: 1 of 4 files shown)


Lotes:
Only thing I saw was: You are using `mutex.read` everywhere, but no `mutex.write` has encountered. I would expect such a thing, when removing a "global" mutex and introducing a local one.

EDIT: Nevermind, saw the `mutex.run`.

EDIT: Works great in the playground. But I would not say, that I smoke tested it enough.

packages/language/src/language-server/connection-handler.ts
```diff
@@ -200,14 +208,17 @@ export function startLanguageServer(connection: Connection): void {
...
-      const compilationUnit =
-        compilationUnitHandler.getCompilationUnit(parsedUri);
-      const textDocument = compilationUnit?.services.files.getDocument(uri);
-      if (textDocument && compilationUnit) {
+    const uri = params.textDocument.uri;
+    const position = params.position;
+    const parsedUri = URI.parse(uri);
+    const compilationUnit =
+      compilationUnitHandler.getCompilationUnit(parsedUri);
+    if (!compilationUnit) {
```
Lotes: Minor: These lines are repeating. Maybe it makes sense to have a central function that does the `!unit` and `textDocument` checks and embeds a return when unit is undefined and also embeds a `mutex acquires` branch.

It might be too much action, since everything is in one file. It would be only one place to change on the other hand.

msujew: Good point, I've consolidated them in the latest commit.

packages/language/src/language-server/connection-handler.ts
```diff
@@ -314,57 +339,48 @@ export function startLanguageServer(connection: Connection): void {
     });
   });
   connection.onWorkspaceSymbol(async (params) => {
```
Lotes, line 318: KnowledgeGap: what is "WorkspaceSymbol" doing? Why has it no mutex anymore?

packages/language/src/language-server/connection-handler.ts
```diff
@@ -314,57 +339,48 @@ export function startLanguageServer(connection: Connection): void {
...
-          diagnostics,
-        });
-
-        // reindex reachable compilation units
-        await compilationUnitHandler.reindex(
-          connection,
-          CancellationToken.None,
-        );
+    async () => {
+      // handle changes to the .pliplugin config folder's contents
```
Lotes, line 327: This one has also no mutex, but it is about the plugin config, which is operating before any mutex should be acquired. Right?

Lotes (approved):
I was not able to break it by querying the affected LSP requests. LGTM.

packages/language/src/language-server/connection-handler.ts
Lotes: ChatGPT: It is looking up the entire symbol index. Symbol search.

OK!

packages/language/src/language-server/connection-handler.ts
msujew: Updating/changing the config async is fine, because the compilation unit caches the config internally. The `reindex` call does the mutex stuff then.

PR #540, opened 2026-01-19
Add completion for LIKE/TYPE attributes

msujew:
Does as the title says, adds completion for the `:<type>:` style type reference and also via `TYPE <type>` and `TYPE (<type>)`. Also adds completion for the `LIKE <variable>` attribute.

Removes a bit of code in the completion provider for the special `case CstNodeKind.*_Semicolon:` cases, these are now just funneled through the `token === Semicolon` check.


packages/language/test/fourslash/completion/type-reference-ordinal.ts
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+//// DEFINE ORDINAL TEST(A, B);
+//// DCL A TYPE <|1>TEST;
```
Lotes, line 15: The keyword ORDINAL can also be used here I believe. Do we make any distinction between ORDINAL and TYPE?

Lotes: I meant `DCL A ORDINAL <|1>TEST;`...
Example from the language reference page 143: `dcl Wall_color ordinal Color;`

packages/language/test/fourslash/completion/like-attribute.ts
```diff
@@ -0,0 +1,23 @@
...
+
+//// DCL 1 A, 2 B FIXED BIN(31);
+//// DCL 1 X, 2 Y FIXED BIN(31);
+//// DCL Z <|1>LIKE <|2>A;
+
+completion.expectAt(1, {
+  includes: ["LIKE"],
+});
+completion.expectAt(2, {
+  includes: ["A", "X"],
```
Lotes, line 22: excludes B and Y

packages/language/test/fourslash/completion/type-reference-ordinal.ts
msujew: This already works, but I'll add a test case for it.

packages/language/test/fourslash/completion/like-attribute.ts
msujew: Currently, we cannot really do this kind of fine grained completion - it's only (1) references (2) member calls or (3) keywords.

Lotes (approved):

msujew:
I'll postpone merging this after https://github.com/zowe/zowe-pli-language-support/pull/546 is reviewed & merged, since it refactors some parser stuff related to parsing of the `:<type>:` references.

PR #541, opened 2026-01-19
Fix macro variable completion

msujew:
Fixes an issue that proposed non-pp variables in macro context. This was because the root of the pp-statements was never set correctly. This change fixes the issue and adds a test.


Lotes (approved):
Nice.

You could think about moving the container assignment (`stmt.container = compilationunit.preProcessorAst`) to the lexer. Then the responsibility is better assigned, I think.

msujew:
> You could think about moving the container assignment (stmt.container = compilationunit.preProcessorAst) to the lexer. Then the responsibility is better assigned, I think.

Good point, might be something for a refactoring down the line. I would keep it as it is for now, since we also assign the `preprocessorAst` in the same method.

PR #542, opened 2026-01-19
Remove outdated scripts

msujew:
We had some outdated scripts that no longer worked after #494. I don't think there is a need (or even a way) to adapt these for the new infrastructure. This PR simply removes the scripts.


scripts/generate-railroad-diagram.mts
Lotes, line 1: Has a reference in the ARCHITECTURE.md

Lotes (commented):
Weird that they were not linked in the `package.json`.
I have found one reference in the markdown.

Lotes (approved):

PR #543, opened 2026-01-20
Improve code completion (statements/expressions)

msujew:
Based on top of https://github.com/zowe/zowe-pli-language-support/pull/541.

Improves the completion further for start of statements (especially for the preprocessor use case) and when we expect an expression next.

We now maintain a list of cst nodes that are supposed to precede an expression (or a general reference).

### DIFF
--- a/packages/language/test/fourslash/completion/declaration-dimensions.ts
+++ b/packages/language/test/fourslash/completion/declaration-dimensions.ts
@@ -11,20 +11,9 @@
 
 /// <reference path="../framework.ts" />
 
-//// DCL 1 A, 2 B, 3 C, 4 D char(10);
-//// <|1>
+//// DCL A FIXED BIN INIT(10);
+//// DCL B(<|1>);
 
 completion.expectAt(1, {
-  includes: [
-    "A",
-    "B",
-    "C",
-    "D",
-    ...constants.CompletionKeywords.StatementStart.values().map(
-      (keyword) => keyword.label,
-    ),
-    ...constants.CompletionKeywords.StatementStartPreprocessorWithPercent.values().map(
-      (keyword) => keyword.label,
-    ),
-  ],
+  includes: ["A"],
 });
--- a/packages/language/src/language-server/completion/completion-generator.ts
+++ b/packages/language/src/language-server/completion/completion-generator.ts
@@ -10,11 +10,15 @@
  */
 
 import { CompletionItemKind } from "vscode-languageserver-types";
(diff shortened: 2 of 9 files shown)


packages/language/test/fourslash/completion/declaration-dimensions.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+//// DCL A FIXED BIN INIT(10);
+//// DCL B(<|1>);
+
+completion.expectAt(1, {
```
Lotes, line 17: Should this exclude non-arithmetic variables later?

packages/language/test/fourslash/completion/pp/assignment.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/// <reference path="../../framework.ts" />
+
+//// DCL X FIXED BIN(31) INIT(1);
+//// %DCL A FIXED INIT(10);
+//// %A = <|1>20;
+//// X = <|2>42;
+
+// Completion at 1 should suggest pp variables only
+completion.expectAt(1, {
```
Lotes, line 20: I think this test was already part of a different PR. Be aware of that for merging.

packages/language/src/workspace/lifecycle.ts
```diff
@@ -61,6 +61,9 @@ export async function tokenize(
   );
   compilationUnit.tokens = result.all;
   compilationUnit.preprocessorAst.statements = result.statements;
+  result.statements.forEach((stmt) => {
```
Lotes, line 64: This was also part of a different PR.

packages/language/test/fourslash/completion/declaration-dimensions.ts
msujew: Maybe, though I'm generally not a big fan of performing type based completion filtering. It usually leads to a lot of unwanted negative side effects. I'd be in favor of sorting though.

packages/language/test/fourslash/completion/pp/assignment.ts
msujew: Yes, as indicated by the PR description, this PR is based on top of the other PR to correctly build the preprocessor completion.

Lotes: I cannot read ^^

Lotes (approved):

PR #546, opened 2026-01-21
Fix type function references

msujew:
Fixes an issue that prevented functions like `BIND(:t, p:)` from parsing correctly.

Introduces a surprising amount of changes, as we now need pass everywhere in the parser whether a reference can accept types (or only normal variables). Enhances the parser infrastructure to support parser rule arguments.


packages/language/src/parser/parser.ts
```diff
@@ -6030,13 +6071,17 @@ const referenceItem = rule(
       element.ref = ast.createReference(
         element,
         idToken,
-        ast.ReferenceType.Variable,
+        refType ?? ast.ReferenceType.Variable,
```
Lotes, line 6074: Interesting, all these changes in here for this line.
I was wondering if there was a way via parser state. Can the `refType` change within an expression?

Lotes: Like using push, pop, peek where it is needed: Entering and leaving an expression (?). But maybe this would confuse more.

With your solution you have opened a path for other missing data, like a parametrized rule.

Lotes (approved):

packages/language/src/parser/parser.ts
msujew: It can change:

```
BIND(:TYPE, SOME_PROC(...):);
```

The `...` part should not reference a type, but the `TYPE` part should.

Issue #545, opened 2026-01-21
PL/I Green code


asatklichov:
Currently PL/I extension is supporting core features of language and with compiler and preprocessor support, and 
support for CICS built-ins and host variables. 

Below items are needed to be implemented to make it more complete, or make error-tolerable     

- [x] (1)    label  name with # 

- [x] (2)   parser errors 
e.g. `IF TEST & Q(118) THEN PUT SKIP DATA;`

e.g. 

```
 L5120:  IF (LIQ(0) = K & HERE(BOTTLE))                                 13800000
                | K = LIQLOC(LOC) THEN GOTO L5010;   
```
e.g.

  `DCL SYSIS    FILE RECORD INPUT KEYED ENV(ORGANIZATION(INDEXED));`

e.g.

```

 SELECT(DSORG);                                                    03690007
      WHEN('PS') OPEN FILE(SYSXX) TITLE(INDD);                          03700007
      OTHERWISE  OPEN FILE(SYSXX) SEQL TITLE(INDD);                     03710007
      END;                                                              03720007
      CALL SETREC;                                                      03730001
      ON ENDFILE(SYSXX) GOTO EOJ;                                       03740007
      R='END OF FILE IN INPUT DATASET   INDD='||INDD;   
```

Lotes:
I will take the parser error around `ENV(ORGANIZATION(INDEXED))`.

EDIT: SEQL is also working now in #554

msujew:
@asatklichov This is ready to test. These should be fixed now.

asatklichov:
@msujew thanks yes it is working, but still IF statement has small caveat still.  
e.g.   Below statements are correct codeyng style. 

```
IF FFLC=QSAM|FFLC=BSAM|**FFLC**=BSAML THEN DO;  

or 

 IF FROM<1|**FROM**>LEN THEN F=1; ELSE F=FROM; 
```
Above statements give Unknown identifier 'FLC' or Unknown identifier 'ROM' error, because of not white space after '|' . 

If we put white space as below , it is resolved.
```
IF FFLC=QSAM| FFLC=BSAM| FFLC=BSAML THEN DO; 

or 

IF FROM<1| FROM>LEN THEN F=1; ELSE F=FROM;  
```

msujew:
@asatklichov I see, it seems to affect all `OR` operators. See https://github.com/zowe/zowe-pli-language-support/pull/562.

asatklichov:
[...]

PR #547, opened 2026-01-21
Fix IF <> assignment ambiguity

msujew:
Based on https://github.com/zowe/zowe-pli-language-support/pull/546 (I needed the `verify.noParserDiagnostics()` test function).

Related to an error in https://github.com/zowe/zowe-pli-language-support/issues/545.

Whenever we encountered an IF that looked like it could be an assignment, but that was actually an `IF` (including the `THEN`), we still thought that it'd be an assignment. For example: `IF (2 * 3) = 6 THEN;...`. Without the `THEN`, this is actually an assignment:

```
// IF statement
IF (2 * 3) = 6 THEN; ...
// Assignment
IF (2 * 3) = 6;
```

Previously, our code assumed that both of these would be assignments. This change fixes the ambiguity in the lookahead and adds a bunch of new tests for the different cases.

### DIFF
--- a/packages/language/src/parser/parser.ts
+++ b/packages/language/src/parser/parser.ts
@@ -28,6 +28,10 @@ import {
 } from "./abstract-parser";
 import { Severe } from "../validation/pli-codes";
 import { Diagnostic, Severity } from "../language-server/types";
+import {
+  performAssignmentLookahead,
+  performEndStatementLookahead,
+} from "./parser-lookahead";
 
 export function parsePli(input: tokens.Token[]): {
   tree: ast.Program;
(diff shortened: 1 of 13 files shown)


packages/language/test/fourslash/parser/assignment-with-array-named-if.ts
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+// @wrap: main
+//// DCL 1 IF(10) FIXED BIN(31);
```
Lotes, line 15: "DCL >1<", 1 is not needed. Is this even valid to have a 1 when it is not a structure?

packages/language/test/fourslash/parser/assignment-multi-variable.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+// @wrap: main
+//// DCL (A, B) FIXED BIN(31);
+//// A, B = 123;
```
Lotes, line 16: Question: Is this valid syntax? What is the Value of A and B?

packages/language/test/fourslash/parser/assignment-with-array-named-if.ts
msujew: As far as I know, it's valid, but I'll confirm with the compiler.

packages/language/test/fourslash/parser/assignment-multi-variable.ts
msujew: It's the same as writing:

```
DCL A FIXED BIN(31);
DCL B FIXED BIN(31);
A = 123;
B = 123;
```

packages/language/test/fourslash/parser/assignment-with-array-named-if.ts
msujew: Just confirmed with the compiler. Completely fine 👍

packages/language/src/parser/parser.ts
```diff
@@ -6499,7 +6499,21 @@ const numberLiteral = rule(
   },
 );
 
+function indicatesAssignment(token: tokens.Token): boolean {
```
Lotes: Since the lookahead performing checks and helper functions are used in both parsers it would make sense to move them to the parser state or an own file `parser-lookaheads.ts`, I think.

Lotes (approved):
Thanks. LGTM :)

PR #548, opened 2026-01-22
Implement deprecate validation

### LINKED ISSUE #398: Add deprecation options
Currently we are not warning if we use deprecated options in compile-time options. 

e.g. What about  using `INCLUDE(‘’)` - empty not error in mainframe 

See the Specification and follow the places with identified as deprecated 

Image

ssmifi:
Closes https://github.com/zowe/zowe-pli-language-support/issues/398

This one looks bigger than it is because of all the statement tests. Functional changes are:
- Implement the deprecate validation for https://github.com/zowe/zowe-pli-language-support/issues/398
Therefore, 
  - ~Add token to statements~ (obsolete after rebase)
  - Use sets for deprecate/deprecatenext compiler options
  - Set deprecate/deprecatenext to undefined as default to avoid validations with empty lists
- Add a flag to compiler option rules that allows duplicates
- BUGFIX: Actually request a new default object when compiler options are reset. This fixes a bug where nested compiler options are not reset when changed

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/options-pli.ts
+++ b/packages/language/src/preprocessor/compiler-options/options-pli.ts
@@ -990,14 +990,9 @@ export namespace CompilerOptions {
       format?: DefaultFormat;
     };
   }
-  export interface Deprecate {
-    // TODO ssmifi: check if these should really be accumulated or be overwritten per type.
-    items: DeprecateItem[];
-  }
-  export interface DeprecateItem {
-    type: DeprecateItemType;
-    value: string;
-  }
+  export type Deprecate = {
+    [K in keyof typeof DeprecateItemType]: Set<string>;
+  };
   export interface Display {
     std?: boolean;
     wto?: boolean;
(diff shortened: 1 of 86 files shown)


packages/language/src/parser/parser.ts
```diff
@@ -664,6 +664,8 @@ const statement = rule(
       label && element.labels.push(label);
     }
 
+    element.token = state.token ?? null;
+
```
msujew: Question: Should the deprecated statement validation also apply for preprocessor statements?

msujew (approved):
Looks good, just one question below.

packages/language/src/parser/parser.ts
ssmifi: I would say no, since the pp has an own deprecate directive that specifically only targets pp procedures. Then again the pli deprecate also covers includes. We should verify this behavior on the mainframe to be sure. Will do.

ssmifi: Preprocessor statements are not affected. I've added a test case to validate this.

Issue #544, opened 2026-01-21
Findings 21-01-2026


asatklichov:
- [x] (1) Provide support for 'SYSPRINT' and ''SYSIN' grammar 

- [x] (2) Support for  Types System - BIND (both case) and RESPEC  
E.g 
```
  h = BIND(:T, P:);  
  h = BIND(: T, ALLOC( SIZE(:T:) ); 
   h = RESPEC(:T, P:);
```
 
- [x] (3)  Reference ambiguity - is shadowed by '"case-upper-validation": true' flag set. 
It does not show ambiguity again, once flag is removed. 

"compiler-options": [
        "PP(MACRO('CASE(UPPER) RESCAN(UPPER)'))"
]

Simply it is a compiler option side affect. If you define below option in compiler options, 
 it  hides errors in declarations, and if you remove it back, it does show it errors back again  

```
 %DCL X FIXED;
 %X = 1;   

 %dcl x(2) fixed entry; 
 %dcl y(2) fixed builtin; 
```
 
         
- [x] (4)  Macro preprocessors
```

Just show not supported  warning - IGNORE, EOLCOM, DBCS
Completion of CASE(UPPER) and INCONLY 

``` 

- [x] (5)  Provide quick fix support as below sample example. 

This example (customer uses their editor) has a quick fix suggestions. 

<img width="382" height="337" alt="Image" src="https://github.com/user-attachments/assets/8af1116e-799d-410b-9a19-1431f1762730" />

 

I tried below code in our PL/I extension,  and  it gives error message, but lacks quick fix suggestion 
or helpful message. 

<img width="900" height="177" alt="Image" src="https://github.com/user-attachments/assets/5080cbd4-87cf-4e02-9a73-b414fa4b75a3" />

```

 GERT1: PROCEDURE OPTIONS (MAIN);

  DCL X CHAR(100); 
  DCL CT BIN FIXED(31);
  
  DCL 1 A BASED(ADDR(X)),
        5 B BIN FIXED(31), 
        5 C (CT REFER(B)),  /*lacks quick fix */
           10* CHAR(1);  


 DCL 1 AA BASED(ADDR(X)),
        5 B BIN FIXED(31), 
        5 C (CT REFER(B));  /*lacks quick fix */  

 END GERT1;
```

asatklichov:
@msujew  items 1,2 and 5
@ssmifi 3 and 4

PR #550, opened 2026-01-22
Fix re-resolution of cached references

ssmifi:
Fix for https://github.com/zowe/zowe-pli-language-support/issues/544 3

If the lifecycle is triggered by a change that uses the instruction cache, e.g. due to changes to plugin config, the symbol table does not resolve the references again and no linker diagnostics are generated. 
The fix invalidates the references that the symbol table wants to build before resolution.


msujew (approved):
LGTM

PR #551, opened 2026-01-26
[Improvement] Handle ddname validation

Wagner-Laranjeiras:
Issue #399 

This PR is a follow-up to #478. 
It
- creates explicitly a "member-name-validation" field on the config and set it to false as default
- adds diagnostics for members using the ABC.DEF(filename) syntax
- extends testing
- improves handling of URI differences between VFS and extension files, including proper encoding and decoding.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2230,11 +2230,9 @@ function isMemberIncludeItem(obj: any): obj is MemberIncludeItem {
 
 /**
  * Encodes each segment of a URI path to ensure special characters are properly escaped.
- * Accepts either a string or URI object and returns the encoded URI as a string.
  */
-function encodeUri(input: string | URI): string {
-  const uri = typeof input === "string" ? URI.parse(input) : input;
-  const encodedPath = uri.path.split("/").map(encodeURIComponent).join("/");
+function encodeUri(input: URI): string {
+  const encodedPath = input.path.split("/").map(encodeURIComponent).join("/");
   return URI.parse(encodedPath).toString();
 }
 
@@ -2297,7 +2295,7 @@ async function runInclude(
   }
 
   try {
-    const document = await TextDocuments.get(encodeUri(uri.path));
+    const document = await TextDocuments.get(encodeUri(uri));
 
     if (!document) {
       throw new Error("Document not found after URI resolution.");
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2509,10 +2509,8 @@ async function resolveIncludeFileUri(
 
   let libMatch: URI | undefined;
   // whether a match was found for a member include
-  const TEST1 = isMemberWithoutDDName(item);
-  const TEST2 = isMemberWithDDName(item);
   let needsMemberValidation =
-    TEST1 || TEST2;
+    isMemberWithoutDDName(item) || isMemberWithDDName(item);
 
   for (const lib of computedLibs) {
     if (isLibsDir(lib)) {
@@ -2538,7 +2536,7 @@ async function resolveIncludeFileUri(
         path: libFileUri,
         extensions: pgroup.includeExtensions,
       });
-      if (libMatch && !isMemberWithDDName) {
+      if (libMatch && !isMemberWithDDName(item)) {
         // regular file match, no member to validate
         needsMemberValidation = false;
         break;
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2508,9 +2508,6 @@ async function resolveIncludeFileUri(
     isMemberWithoutDDName(item) || isMemberWithDDName(item);
 
   for (const lib of computedLibs) {
-    if (libMatch) {
-      break;
-    }
     if (isLibsDir(lib)) {
       const libFileUri = resolveLibFileUri(lib.dir, fileNameOrPartial);
 
@@ -2534,9 +2531,9 @@ async function resolveIncludeFileUri(
         path: libFileUri,
         extensions: pgroup.includeExtensions,
       });
-      if (libMatch && !isMemberWithDDName(item)) {
+      if (libMatch) {
         // regular file match, no member to validate
-        needsMemberValidation = false;
+        needsMemberValidation = isMemberWithDDName(item);
         break;
       }
     } else if (isMemberWithoutDDName(item)) {
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2228,13 +2228,14 @@ function isMemberIncludeItem(obj: any): obj is MemberIncludeItem {
   );
 }
 
-function handleUri(input: string | URI): string {
+/**
+ * Encodes each segment of a URI path to ensure special characters are properly escaped.
+ * Accepts either a string or URI object and returns the encoded URI as a string.
+ */
+function encodeUri(input: string | URI): string {
   const uri = typeof input === "string" ? URI.parse(input) : input;
-  const key = URI.parse(
-    uri.path.split("/").map(encodeURIComponent).join("/"),
-  ).toString();
-
-  return key;
+  const encodedPath = uri.path.split("/").map(encodeURIComponent).join("/");
+  return URI.parse(encodedPath).toString();
 }
 
 async function runInclude(
@@ -2296,7 +2297,7 @@ async function runInclude(
   }
 
   try {
-    const document = await TextDocuments.get(handleUri(uri));
+    const document = await TextDocuments.get(encodeUri(uri));
 
     if (!document) {
       throw new Error("Document not found after URI resolution.");
@@ -2416,16 +2417,8 @@ async function resolveIncludeFileUri(
     return undefined;
   }
 
-  // whether the include item is a standalone member, no ddname specified
-  // in such cases this member may be the suffix of an a ddname entry in the libs, (ex. `A.B.C(member)`)
-  // corresponding to mainframe behavior, if `cpy/A.B.C` or `cpy` is in libs, we should be able to resolve `member`
-  const isMemberWithoutDDName = isMemberIncludeItem(item) && !item.ddname;
-  // Check if it's member wish specified ddname
-  const isMemberWithDDName = !!(
-    isMemberIncludeItem(item) &&
-    item.ddname &&
-    item.memberName
-  );
+  const isMemberWithoutDDName2 = isMemberIncludeItem(item) && !item.ddname;
+  console.log(isMemberWithoutDDName2);
 
   /**
    * Computes the URI for a lib file based on whether the path is absolute or relative.
@@ -2484,9 +2477,42 @@ async function resolveIncludeFileUri(
     }
   }
 
+  /**
+   * Type guard to check if an include item is a standalone member without a DDName specified.
+   * In such cases, this member may be the suffix of a DDName entry in the libs (e.g., `A.B.C(member)`).
+   * Following mainframe behavior, if `cpy/A.B.C` or `cpy` is in libs, we should be able to resolve `member`.
+   *
+   * @param item - The include item to check
+   * @returns True if the item is a member include without a ddname
+   */
+  function isMemberWithoutDDName(
+    item: IncludeItem,
+  ): item is MemberIncludeItem & { ddname: undefined } {
+    return isMemberIncludeItem(item) && !item.ddname;
+  }
+
+  /**
+   * Type guard to check if an include item is a member with an explicitly specified DDName (ex. `ABC.DEF(MEMBER)`).
+   *
+   * @param item - The include item to check
+   * @returns True if the item is a member include with both ddname and memberName populated
+   */
+  function isMemberWithDDName(
+    item: IncludeItem,
+  ): item is MemberIncludeItem & { ddname: string; memberName: string } {
+    return (
+      isMemberIncludeItem(item) &&
+      Boolean(item.ddname) &&
+      Boolean(item.memberName)
+    );
+  }
+
   let libMatch: URI | undefined;
   // whether a match was found for a member include
-  let needsMemberValidation = isMemberWithoutDDName || isMemberWithDDName;
+  const TEST1 = isMemberWithoutDDName(item);
+  const TEST2 = isMemberWithDDName(item);
+  let needsMemberValidation =
+    TEST1 || TEST2;
 
   for (const lib of computedLibs) {
     if (isLibsDir(lib)) {
@@ -2517,7 +2543,7 @@ async function resolveIncludeFileUri(
         needsMemberValidation = false;
         break;
       }
-    } else if (isMemberWithoutDDName) {
+    } else if (isMemberWithoutDDName(item)) {
       // standalone member w/out an explicit ddname, search within ddlib for a match
       const ddLibUri = resolveLibFileUri(lib.ddLib);
       libMatch = await FileSystemProviderInstance.search({
@@ -2528,11 +2554,11 @@ async function resolveIncludeFileUri(
         break;
       }
     } else if (
-      isMemberWithDDName &&
-      lib.ddLib.toLowerCase().endsWith(item.ddname!.toLowerCase())
+      isMemberWithDDName(item) &&
+      lib.ddLib.toLowerCase().endsWith(item.ddname.toLowerCase())
     ) {
       // member w/ ddname, search for an exact match using ddlib
-      const end = lib.ddLib.length - item.ddname!.length;
+      const end = lib.ddLib.length - item.ddname.length;
       const libPath = lib.ddLib.substring(0, end);
       const ddLibUri = resolveLibFileUri(libPath, fileNameOrPartial);
       libMatch = await FileSystemProviderInstance.search({
@@ -2547,9 +2573,8 @@ async function resolveIncludeFileUri(
 
   // depending on whether we matched a member, check to apply validation on the name
   if (needsMemberValidation) {
-    const memberToValidate = isMemberWithDDName
-      ? // Extract the member name from the parenthesis
-        fileNameOrPartial.split("(")[1].split(")")[0]
+    const memberToValidate = isMemberWithDDName(item)
+      ? item.memberName
       : fileNameOrPartial;
 
     checkToValidateMember(memberToValidate);
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2228,14 +2228,6 @@ function isMemberIncludeItem(obj: any): obj is MemberIncludeItem {
   );
 }
 
-/**
- * Encodes each segment of a URI path to ensure special characters are properly escaped.
- */
-function encodeUri(input: URI): string {
-  const encodedPath = input.path.split("/").map(encodeURIComponent).join("/");
-  return URI.parse(encodedPath).toString();
-}
-
 async function runInclude(
   item: IncludeItem,
   context: InterpreterContext,
@@ -2295,7 +2287,7 @@ async function runInclude(
   }
 
   try {
-    const document = await TextDocuments.get(encodeUri(uri));
+    const document = await TextDocuments.get(uri);
 
     if (!document) {
       throw new Error("Document not found after URI resolution.");
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2488,7 +2488,7 @@ async function resolveIncludeFileUri(
     return (
       isMemberIncludeItem(item) &&
       item.ddname !== null &&
-      item.memberName !== null
+      item.memberName.length > 0
     );
   }
 
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -2411,6 +2411,9 @@ async function resolveIncludeFileUri(
   // in such cases this member may be the suffix of an a ddname entry in the libs, (ex. `A.B.C(member)`)
   // corresponding to mainframe behavior, if `cpy/A.B.C` or `cpy` is in libs, we should be able to resolve `member`
   const isMemberWithoutDDName = isMemberIncludeItem(item) && !item.ddname;
+  // Check if it's member wish specified ddname
+  const isMemberWithDDName =
+    !!(isMemberIncludeItem(item) && item.ddname && item.memberName);
 
   /**
    * Computes the URI for a lib file based on whether the path is absolute or relative.
@@ -2471,7 +2474,7 @@ async function resolveIncludeFileUri(
 
   let libMatch: URI | undefined;
   // whether a match was found for a member include
-  let needsMemberValidation = isMemberWithoutDDName;
+  let needsMemberValidation = isMemberWithoutDDName || isMemberWithDDName;
 
   for (const lib of computedLibs) {
     if (isLibsDir(lib)) {
@@ -2497,7 +2500,7 @@ async function resolveIncludeFileUri(
         path: libFileUri,
         extensions: pgroup.includeExtensions,
       });
-      if (libMatch) {
+      if (libMatch && !isMemberWithDDName) {
         // regular file match, no member to validate
         needsMemberValidation = false;
         break;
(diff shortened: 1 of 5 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2228,6 +2228,15 @@ function isMemberIncludeItem(obj: any): obj is MemberIncludeItem {
   );
 }
 
+function handleUri(input: string | URI): string {
```
ssmifi: `handleUri` is quite generic. Please choose a more descriptive name.
The same applies to `key`. This name suggests a value used as a key in a corresponding data structure.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2411,6 +2420,12 @@ async function resolveIncludeFileUri(
   // in such cases this member may be the suffix of an a ddname entry in the libs, (ex. `A.B.C(member)`)
   // corresponding to mainframe behavior, if `cpy/A.B.C` or `cpy` is in libs, we should be able to resolve `member`
   const isMemberWithoutDDName = isMemberIncludeItem(item) && !item.ddname;
+  // Check if it's member wish specified ddname
+  const isMemberWithDDName = !!(
```
ssmifi: I think the `!!` is unnecessary here. Please also make the comment more specific. If the consts are not needed in the local functions, you could move them closer to `needsMemberValidation` to avoid the code fragmentation. Also, `isMemberIncludeItem` is evaluated twice. Which could be fine, but maybe you want to restructure a bit.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2513,12 +2528,11 @@ async function resolveIncludeFileUri(
         break;
       }
     } else if (
-      isMemberIncludeItem(item) &&
-      item.ddname &&
-      lib.ddLib.toLowerCase().endsWith(item.ddname.toLowerCase())
+      isMemberWithDDName &&
+      lib.ddLib.toLowerCase().endsWith(item.ddname!.toLowerCase())
```
ssmifi: _Suggestion_: If you refactor this slightly, you could consider using a type guard to ensure ddname is set and avoid the non-null assertion (`!`). For example:
```typescript
function isMemberWithDDName(item: IncludeItem): item is MemberIncludeItem & { ddname: string; memberName: string } {
  return isMemberIncludeItem(item) && !!item.ddname && !!item.memberName;
}
```

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2533,7 +2547,12 @@ async function resolveIncludeFileUri(
 
   // depending on whether we matched a member, check to apply validation on the name
   if (needsMemberValidation) {
-    checkToValidateMember(fileNameOrPartial);
+    const memberToValidate = isMemberWithDDName
+      ? // Extract the member name from the parenthesis
+        fileNameOrPartial.split("(")[1].split(")")[0]
```
ssmifi: _Question_: This looks a bit brittle. Could we use memberName here instead?

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -152,12 +171,23 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
     this.sortedFilesCache = undefined;
   }
 
+  /**
+   * Converts a URI object to a normalized file:// URL string.
+   * Ensures proper file:// protocol prefix and includes fragment if present.
+   * Returns lowercase normalized path for case-insensitive comparison.
+   */
+  buildUri(uri: URI): string {
```
ssmifi: The name is misleading. There is no Uri being built. Could you clarify in the comment why this is needed (as opposed to the standard uri.toString())? I assume this is the inverse of the ddname issue with the normal uri handling, but that is not clear. Also, the string construction is not clear to me. Why do we need to differentiate between prefixes? Is the lowercase an issue for case-sensitive file systems?

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -240,9 +270,10 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
    */
   async search(options: SearchOptions): Promise<URI | undefined> {
     if (isPathSearch(options)) {
+      // const hasParenthesis = options.path.toString(true).includes("(") && options.path.toString(true).includes(")");
```
ssmifi: Please remove the commented code. But you can comment on the use of decodeURI and why it is necessary for ddnames for example.

ssmifi (commented):
Thanks for taking this on. I’d suggest adding a bit more concrete documentation around the ddname handling and why the default URI functions aren’t sufficient in this case. That extra context would make the intent clearer for future readers.

Wagner-Laranjeiras:
Thanks for the review @ssmifi ! I'll work on it ;)

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: Got it, gonna improve it

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: Gonna work on the comment! I had to force a boolean here, because otherwise, if all conditions were met, the variable would return the value from `item.memberName` and ending up being a string (which for me was a surprise behaviorwise).

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: Got it. That crossed my mind, which brings me to the follow up question: by doing that, should I also do it for the `isMemberWithoutDDName`, for consistency sake?

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: Not sure, but I would say yes. Thanks for catching this

packages/language/src/workspace/file-system-provider.ts
Wagner-Laranjeiras: You're right, that does sound misleading. Gonna change that. I added this prefix handling because in other occasions I worked with paths that started without a slash (ex.: in files seating directly under the workspace root). The lowercase was there from before, and we do that in order parts of the code. I think it's important to deal with URIs in a standard manner, since the file list in which we search is all lowercase.

packages/language/src/workspace/file-system-provider.ts
Wagner-Laranjeiras: Ops, that escaped my eyes. Thanks for pointing out

packages/language/test/fourslash/preprocessor/include/with-member-validation.ts
```diff
@@ -43,6 +43,26 @@
...
+
+// @filename: cpy/ABC.DEF(GHIJKLMNO)
+//// // expecting to be invalid
+//// DECLARE V7 FIXED;
+
+// @filename: cpy/ABC.DEF(#GH@#_$I)
+//// // expecting to be invalid
+//// DECLARE V8 FIXED;
+
+// @filename: cpy/ABC.DEF(_GH@#_$IJKLM)
```
Lotes, line 62: Nice test extension.

Minor: It is a bit confusing to read, because label `N` has the content `DECLARE V<N-1> FIXED;`. I was searching for `V10`. I think you just continued this pattern. Could you align the labels? I am not sure if this is so important haha, but I spotted it...

Lotes: Mark has also built a new naming strategy for labels some months ago.
When you do `<|m12345678|>` instead of `<|1:m12345678|>`, the label is `m12345678`.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2228,6 +2228,14 @@ function isMemberIncludeItem(obj: any): obj is MemberIncludeItem {
   );
 }
 
+/**
+ * Encodes each segment of a URI path to ensure special characters are properly escaped.
+ */
+function encodeUri(input: URI): string {
+  const encodedPath = input.path.split("/").map(encodeURIComponent).join("/");
```
Lotes: This split-join I also saw a bit down a other file. Any chance to reduce this duplication :)?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2287,7 +2295,8 @@ async function runInclude(
   }
 
   try {
-    const document = await TextDocuments.get(uri);
+    const document = await TextDocuments.get(encodeUri(uri));
```
Lotes: Is this ok? I mean are the URI of these documents were created with this encoded URI?

Lotes: I would expect that the same function to normalize this URI should be reused at each place dealing with URIs...

I do not feel secure enough in this part of the code to give an approval :-( ...

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -51,6 +51,25 @@ export function isPathSearch(obj: any): obj is PathSearch {
...
+ * - Unencoded `#` characters being interpreted as fragment delimiters
+ * - Over-encoding `/` and `:` (which would break the URI structure)
+ * - Using `encodeURIComponent` on the entire path, which is incorrect
+ */
+function filePathToFileURI(path: string): URI {
+  return URI.parse(
+    "file://" + path.split("/").map(encodeURIComponent).join("/"),
+  );
+}
+
```
msujew: Suggestion: Isn't there already `URI.file` for this exact use case? Or does it also suffer from the `#` issue?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2228,6 +2228,14 @@ function isMemberIncludeItem(obj: any): obj is MemberIncludeItem {
   );
 }
 
+/**
+ * Encodes each segment of a URI path to ensure special characters are properly escaped.
+ */
+function encodeUri(input: URI): string {
+  const encodedPath = input.path.split("/").map(encodeURIComponent).join("/");
+  return URI.parse(encodedPath).toString();
+}
```
msujew: Suggestion: This function removes the URI schema. I would like to have a `UriUtils.parseWithoutFragment` function instead, that mirrors the [original behavior](https://github.com/microsoft/vscode-uri/blob/65786c7aef8aa1d142fedfde76073cc3549736d2/src/uri.ts#L266-L279). We should only have a single function for parsing URI strings into `URI` objects. Otherwise we will run into cache misses on documents with the same URI.

packages/language/src/preprocessor/instruction-interpreter.ts
msujew: I agree - even if it works fine for now, it's rather brittle.

packages/language/test/fourslash/preprocessor/include/with-member-validation.ts
Wagner-Laranjeiras: Yep, I also caught that haha the reason is because we are testing also for a file that doesn't exist ("_A1@#"). That's why we have 10 includes/tests but only 9 variables. 

That said, I didn't know about the naming strategy from Mark. Thanks for mentioning it, I applied to the test now! ;)

packages/language/src/workspace/file-system-provider.ts
Wagner-Laranjeiras: We could use it, the only difference I could catch in this case is that `URI.parse()` returns a `.path` already without the `file://` on the string, while `URI.file()` maintains it. The following would work though:

```
  const stringPath = path.toLowerCase().includes("file:")
    ? path.replace("file:", "")
    : path;
  return URI.file(stringPath);
```

Wagner-Laranjeiras:
Thank you for the review guys!

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: At the end of the day, I was able to get the document passing just the regular URI (and tested both manually as well with written tests). I can still implement this function, but it currently wouldn't be used.

Even though the function might not be implemented right now, I learned from you guys comments, so thanks again for the reviews.

packages/language/test/fourslash/preprocessor/include/with-member-validation.ts
Lotes: Thanks, in my eyes it is more readable now :)

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2469,9 +2464,40 @@ async function resolveIncludeFileUri(
...
+   * In such cases, this member may be the suffix of a DDName entry in the libs (e.g., `A.B.C(member)`).
+   * Following mainframe behavior, if `cpy/A.B.C` or `cpy` is in libs, we should be able to resolve `member`.
+   *
+   * @param item - The include item to check
+   * @returns True if the item is a member include without a ddname
+   */
+  function isMemberWithoutDDName(
+    item: IncludeItem,
+  ): item is MemberIncludeItem & { ddname: undefined } {
+    return isMemberIncludeItem(item) && !item.ddname;
```
ssmifi, line 2476: _Minor_: `ddname` is already defined as `string | null`. Unless there is a particular reason, there is no need to "recast" it to `undefined`.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2469,9 +2464,40 @@ async function resolveIncludeFileUri(
...
+   * @returns True if the item is a member include with both ddname and memberName populated
+   */
+  function isMemberWithDDName(
+    item: IncludeItem,
+  ): item is MemberIncludeItem & { ddname: string; memberName: string } {
+    return (
+      isMemberIncludeItem(item) &&
+      Boolean(item.ddname) &&
+      Boolean(item.memberName)
+    );
```
ssmifi, line 2492: I would not use a boolean constructor here. Booleans usually can be tested easily: `item.ddname !== null`. `item.memberName` is always set, isn't it?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2499,10 +2525,10 @@ async function resolveIncludeFileUri(
       });
       if (libMatch) {
         // regular file match, no member to validate
-        needsMemberValidation = false;
+        needsMemberValidation = isMemberWithDDName(item);
```
ssmifi, line 2526: _Question_: I don't get this assignment. `isMemberWithDDName(item)` would have already set `needsMemberValidation` to true before the loop, wouldn't it?
_If_ this is not needed here, the `needsMemberValidation` at the bottom could be moved before the loop, and the loop cases could just return as soon as they find the correct `libMatch`.

packages/language/src/workspace/file-system-provider.ts
ssmifi: _Question_: Could we just use `URI.file` and drop `filePathToFileURI`? Do we need to strip the `file:`?

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -152,12 +164,27 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
...
+   * Returns lowercase normalized path for case-insensitive comparison.
+   */
+  reconstructFileUri(uri: URI): string {
+    const prefix = uri.path.startsWith("/") ? "file://" : "file:///";
+    // In some cases, filenames containing `#` (e.g. `A1@#_$`) are split by URI parsing:
+    // everything after `#` is interpreted as a URI fragment. As a result, the path
+    // becomes `A1@` and `_$` is treated as the fragment, causing part of the filename
+    // to be lost if not handled explicitly.
+    const fragment = uri.fragment ? `#${uri.fragment}` : "";
+    return `${prefix}${uri.path}${fragment}`.toLowerCase();
```
ssmifi, line 178: _Question_: Do we have to reconstruct this by ourserlves? This looks a bit artificial to me. Why can't we use something like `uri.toString(true).toLowerCase()`?

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -240,9 +267,7 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
    */
   async search(options: SearchOptions): Promise<URI | undefined> {
     if (isPathSearch(options)) {
-      // perform a search with a uri
-      const searchPath = options.path
-        .toString(true)
+      let searchPath = decodeURIComponent(options.path.toString(true))
```
ssmifi, line 270: If URI.file is working for dd members, you could probably use `URI.file(uri.fsPath).toString(true).toLowerCase()` and stay on URI territory.

ssmifi (commented):
Thanks. File handling can definitely be exhausting. :)
That said, I’d feel more comfortable if we could reuse the existing URI capabilities and minimize any reconstruction on our side. Could you please review the open questions and test whether `URI.file` can handle ddnames?

Wagner-Laranjeiras:
Thanks for the reviews Steven, I'm working on it right now

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: Got it, thanks ;)

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: Thanks! When I check for the `.length`, it could return the last item size. I used boolean to be more explicit, but `item.ddname !== null`should also work well 😄

packages/language/src/preprocessor/instruction-interpreter.ts
Wagner-Laranjeiras: I would say we need this in here, because we use `libMatch` for catching both regular files and `ddmembers`. So we need to make sure that we are pointing out if this match is indeed a member with the `ABC.DEF(MEMBER)` syntax or a regular file that doesn't need the validation.

packages/language/src/workspace/file-system-provider.ts
Wagner-Laranjeiras: _Question 1:_  It also bothers me because it feels “hacky”. But at this point we get the path as a string, that usually contains the prefix (ex.: `file:///cpy/ddname(m12345678)`).    If we don’t strip the “file:” from it and pass it into the `URI.file()`, then the output is the following object:  
```
{
  scheme: "file",
  authority: "",
  path: "/file:///cpy/ddname(m12345678)",
  query: "",
  fragment: "",
  _formatted: "file:///file%3A///cpy/ddname%28m12345678%29",
  _fsPath: null,
}
```
  instead of the current

```
    {
  scheme: "file",
  authority: "",
  path: "/cpy/ddname(m12345678)",
  query: "",
  fragment: "",
  _formatted: "file:///cpy/ddname%28m12345678%29",
  _fsPath: null,
}

```
_________________________________

  _Question 2:_  This one also bothers me and it’s there to handle really specific cases. The `uri.toString(true).toLowerCase()` works perfectly for most case scenarios.  But when we have to handle a path that contains a “#” (ex.: `"/cpy/ddname(a1@#_$)”`), then the string we use to get the file is `"file:///cpy/ddname(a1@%23_$)"` instead of `"file:///cpy/ddname(a1@#_$)”`. I also tried using `URI.file(uri.toString(true).toLowerCase()).path` but it ends up being the same. I can't use without the `true` because it ends up even worse. 
  
  I really wish there were some better solutions for those problems (and probably there are, but I couldn't find so far).

packages/language/src/workspace/file-system-provider.ts
Wagner-Laranjeiras: Same problem here 🥲 an input like the following:
```
{
  scheme: "file",
  authority: "",
  path: "/cpy/ABC.DEF(GH@#_$I)",
  query: "",
  fragment: "",
  _formatted: "file:///cpy/ABC.DEF%28GH%40%23_%24I%29",
  _fsPath: "/cpy/ABC.DEF(GH@#_$I)",
}
```

would return `"file:///cpy/abc.def(gh@%23_$i)"` instead of `"file:///cpy/abc.def(gh@#_$i)"`, which messes with the search that happens right after.

msujew: Are you using `toString(true)`? That is supposed to skip the encoding of the URI.

Wagner-Laranjeiras: We need the file names decoded for the searches that happens later to match (ex.: the `this.files` holds the decoded names of the files in the vsf)

Wagner-Laranjeiras:
Guys, if you have time later this week, could you please test this code on your machine?

The full environment (files and setup) is available at `investigation/wagner/include-behavior-fix`. It's just build and run the extension, which will provide all info you need.

I’m seeing different results compared to Azat’s machine.

packages/language/src/preprocessor/instruction-interpreter.ts
ssmifi: `item.memberName.length !== false` should always be true. How about `item.memberName.length > 0` if you want to check for not empty?

packages/language/src/preprocessor/instruction-interpreter.ts
ssmifi: Ah, ok, i see. This actually can set the membername back to false if `isMemberWithoutDDName` set it to true before. Got it. Thanks.

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -286,7 +311,7 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
           fpl.endsWith(memberPart) &&
           fpl.startsWith(options.dirPath.toString(true).toLowerCase())
         ) {
-          return URI.parse(filePath);
+          return filePathToFileURI(filePath);
```
ssmifi: I looked more into the URI issue. As Wagner pointed out, we have a mismatch in the handling of URIs here. Example: 
```
file:///cpy/ddname(a1@#_$)
d {
  scheme: 'file',
  authority: '',
  path: '/cpy/ddname(a1@#_$)',
  query: '',
  fragment: '',
  _formatted: null,
  _fsPath: null
}
d {
  scheme: 'file',
  authority: '',
  path: '/file:///cpy/ddname(a1@#_$)',
  query: '',
  fragment: '',
  _formatted: null,
  _fsPath: null
}
```

The issue is that URI.file expects an os file path and constructs a URI. The string is kept as path, which would be ideal for the file names containing special chars (#). However, we are already providing a complete URI in the string and that is also kept as it is. So, IMHO, the real issue here is providing the correct value to URI.file, which should do the job. 

A linked question is if we need to retrieve the full URI as strings in `getSortedFiles`. If so, I think the string should be normalized between the calls of `getSortedFiles` and `filePathToFileURI`, e.g., by adding a helper to the uri tools like `stripSchemaFromURIString` (filePathToFileURI) or similar. But I think it would be better to avoid this if possible.
Anyway, we can untangle this even more later, but we should try to keep the confusion small. If there is normalization necessary, please comment clearly. 

Furthermore, if we decode manually `filePathToFileURI(filePath)` can be replaced by `URI.parse(filePath)` and we can drop `filePathToFileURI`. At least the provided tests are fine.

Wagner-Laranjeiras:
Hey guys, thanks again for the reviews!

@ssmifi , I changed the boolean checks like you suggested. I couldn't check via the `length` though, because the value could be `null` for the `ddname`, but as you pointed out, `memberName` could benefit from it.

I also added the `stripSchemaFromURIString` directly to the value returned from `getSortedFiles`, but that also forced me to apply it to the conditional we have under the search, later in the code. Please let me know if you prefer the other way.

In any case, I agree that we need to revisit it and refactor, so we deal more consistently with URI paths. That would be the follow up task.

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -229,7 +255,8 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
         return aSlashes - bSlashes;
       });
     }
-    return this.sortedFilesCache;
+
+    return this.sortedFilesCache.map(stripSchemaFromURIString);
```
ssmifi: I wouldn't strip it every time in the get. If possible, get should give the files back as they were put it. But you can strip the schema in the search where you retrieve with `getSortedFiles` and strip the results then.

ssmifi:
I left one further comment about the placement of your strip schema. I've seen that you created https://github.com/zowe/zowe-pli-language-support/issues/568 and I agree that we should streamline the uri handling here in the future. Apart from that, we can probably pull this in for the release if @msujew agrees.

packages/language/src/workspace/file-system-provider.ts
Wagner-Laranjeiras: Got it, thanks! Gonna work on this change now and push it. ;)

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -279,12 +304,16 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
...
+      const sortedFiles = this.getSortedFiles().map(stripSchemaFromURIString);
       for (const filePath of sortedFiles) {
         const fpl = filePath.toLowerCase();
         if (
           fpl.endsWith(memberPart) &&
-          fpl.startsWith(options.dirPath.toString(true).toLowerCase())
+          fpl.startsWith(
+            stripSchemaFromURIString(
+              options.dirPath.toString(true),
+            ).toLowerCase(),
```
msujew: Suggestion: I believe this code is calling `stripSchemaFromURIString(...).toLowerCase()` on the same content on each iteration of `sortedFiles`, right? We should compute this only once and then reuse it.

msujew (approved):
Looks good to me, just one more comment.

packages/language/src/workspace/file-system-provider.ts
Wagner-Laranjeiras: Yes, you're completely right. Oversaw that, thanks for catching!

packages/language/src/preprocessor/instruction-interpreter.ts
ssmifi: Maybe I was confused here. If you just want to test for empty, you can use ! ofc.

PR #552, opened 2026-01-26
Note validation

Lotes:
Fix REPLACE instruction to be included into scope computation.


packages/language/src/linking/symbol-table.ts
```diff
@@ -480,6 +475,14 @@ function handleNode(
...
   // This switch statement handles special cases for certain syntax nodes.
   switch (node.kind) {
+    case SyntaxKind.ReplaceStatement:
+      if (node.name && node.nameToken) {
+        parentScope.symbolTable.addSymbolDeclaration(
+          node.name,
+          QualifiedSyntaxNode.createExplicit(node.nameToken, node),
+        );
+      }
+      break;
```
msujew, line 496: Minor: Can you move these below the `ProcedureStatement` case? That way we don't mix name-defining elements and those that need special handling.

packages/language/src/validation/pli-codes.ts
```diff
@@ -77,6 +77,14 @@ export function isSimplePLICode(code: unknown): code is SimplePLICode {
   return isPLICode(code) && typeof code.message === "string";
 }
 
+export const CustomCodes = {
+  UnknownIdentifier: {
+    code: "UNKNOWN_IDENTIFIER",
+    severity: Severity.E,
+    message: (name: string) => `Unknown identifier '${name}'`,
+  },
+};
```
msujew: Suggestion: This file is exclusively for compiler error codes. Please move this into `lsp-codes.ts` or `internal-codes.ts`.

code_samples/preprocessor/note.pli
msujew, line 1: Suggestion: I don't think this file belongs here? Refer to the other files in the directory. You can move this into the `code_samples/preprocessor` folder.

msujew (commented):
Looks mostly good 👍

msujew (approved):
Thanks 👍

PR #553, opened 2026-01-26
Improve symbol table generation performance

msujew:
Improves the performance of the symbol table generation by ~40% by improving on four main points:

1. Only set the scope on the procedure/package level and walk up the parent chain if necessary. Theoretically, this is computationally the same, but reduces the size of the scope maps massively, and therefore reduces memory pressure.
2. Cache the IDs of qualified syntax nodes, which was expensive to compute.
3. Move the decision whether something is a priority reference into the recursion algorithm. This ensures that we don't have to walk the expression tree multiple times.
4. Massively reduces the garbage we produce during the symbol table generation by replacing the multi-map function results with iterators instead of arrays. This massively cuts down on the produced garbage. Previously, the symbol table generation spent ~30% just in the garbage collector, which this PR reduces to <10%.

### DIFF
--- a/packages/language/src/linking/symbol-table.ts
+++ b/packages/language/src/linking/symbol-table.ts
@@ -39,10 +39,6 @@ import { LinkerErrorReporter } from "./error";
 import { DiagnosticCategory } from "../validation/diagnostics-store";
 import { getFirstStructureVariable } from "../syntax-tree/ast-utils";
 
-function nonNull<T>(value: T | null | undefined): value is T {
-  return value !== null && value !== undefined;
-}
-
 /**
  * Terminology:
  * - Redeclaration: A symbol that is declared with the same name in the same scope.
@@ -290,7 +286,13 @@ function getQualifiedSymbols(
   }
 }
 
-const RedeclarationPriority = [SyntaxKind.LabelPrefix];
+function redeclarationPriority(kind: SyntaxKind): number {
+  if (kind === SyntaxKind.LabelPrefix) {
+    return 1;
+  } else {
+    return 0;
+  }
+}
 
 /**
  * Assigns the `isRedeclared` property to all symbols that are redeclared.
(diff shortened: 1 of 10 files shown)


packages/language/src/utils/collections.ts
```diff
@@ -145,8 +147,20 @@ export class MultiMap<K, V> {
...
+  *values(): IterableIterator<V> {
+    for (const values of this.map.values()) {
+      for (const value of values) {
+        yield value;
+      }
+    }
+  }
+
+  valuesArray(): V[] {
+    const result: V[] = [];
```
Lotes, line 159: Can be done by `this.map.values().flatMap(v => v)`, I think.

packages/language/src/validation/validator.ts
```diff
@@ -137,29 +137,24 @@ function linkingRedeclarationErrorsToDiagnostics(
...
-      .flatMap((scope) => scope.symbolTable.symbols.values())
-      .filter((symbol) => symbol.isRedeclared),
-  );
-
-  for (const symbol of symbols) {
-    const nameToken = symbol.token;
-    if (!nameToken) {
-      continue;
-    }
+  for (const scope of regularScopeCache.values()) {
```
Lotes, line 140: I am wondering if specifically this part is really faster. There is no break or early exit. I would assume that - only this part - is slower using generators instead of using arrays.
But I can be wrong... I just made this experience from a hobby project 10 years ago or so: Same code with generator instead of arrays was way slower. The only advantage I see is when you have conditions that allow to early exit.

packages/language/src/linking/symbol-table.ts
```diff
@@ -443,28 +458,35 @@ function iterateSymbolTable(
...
       context.referencesCache.add(ref);
     }
   }
 
-  // We connect the current node to its scope for usage in the linking phase.
-  context.scopeCache.add(node, parentScope);
-
   const continueRunning = handleNode(node, parentScope, context);
-  if (continueRunning) {
+  if (continueRunning === true) {
```
Lotes: Is " === true" really necessary?

packages/language/src/linking/symbol-table.ts
```diff
@@ -536,14 +558,14 @@ function handleNode(
...
     case SyntaxKind.AssignmentStatement:
-      const refs = node.refs
-        .map((ref) => ref.element?.element?.ref)
-        .filter(nonNull);
-      for (const ref of refs) {
-        parentScope.symbolTable.addImplicitDeclarationByReference(
-          ref,
-          context.reporter,
-        );
+      for (const nodeRef of node.refs) {
```
Lotes, line 568: For this change I have mixed feelings ^^... hopefully this change will not cycle back. haha

packages/language/src/utils/collections.ts
msujew: Flatmap returns an `IteratorObject`. We could do `Array.from`, but I feel like this is the most efficient and readable we can get here.

packages/language/src/validation/validator.ts
msujew: The slow part wasn't the iteration itself, but rather that calling `regularScopeCache.values().flatMap().filter()` generated two *gigantic* arrays. This was the slowest part and also led to a massive slowdown in the garbage collection. Using an iterator completely eliminates the array allocations.

packages/language/src/linking/symbol-table.ts
msujew: The previous implementation needed to do three loops over every array. First to map, then filter and then a final loop to generate the table entry. This only requires to loop once, but does the exact same internally.

packages/language/src/linking/symbol-table.ts
msujew: Not really, I'll drop that.

Lotes (approved):

PR #554, opened 2026-01-27
Add environment options

Lotes:
See [here](https://www.ibm.com/docs/en/epfz/6.1.0?topic=transmission-specifying-environment-options), [here](https://www.ibm.com/docs/en/epfz/6.1.0?topic=sets-specifying-environment-options) and [here](https://www.ibm.com/docs/en/epfz/6.1.0?topic=attribute-environment).

I am not so sure about `CONSECUTIVE` and `INDEXED`, do they only come with `ORGANIZATION(...)`?


packages/language/test/fourslash/parser/environment-attribute/env-organization.ts
```diff
@@ -9,10 +9,9 @@
...
-/// <reference path="../framework.ts" />
+/// <reference path="../../framework.ts" />
 
 // @wrap: main
-//// DCL (A, B) FIXED BIN(31);
-//// A, B = 123;
+//// DCL SYSIS FILE RECORD INPUT KEYED ENV(ORGANIZATION(INDEXED));
 
-verify.noParserDiagnostics();
+verify.noDiagnostics();
```
msujew: Suggestion: Just to make sure we don't introduce any additional validations in the future that break this test. Please also apply this to the other tests.

```suggestion
verify.noParserDiagnostics();
```

code_samples/environment-variables.pli
msujew, line 1: I don't think we need this file.

msujew (approved):
Looks good, just a suggestion for the tests, see below.

PR #555, opened 2026-01-27
Quickfix for ambiguous references

Lotes:
For #544 part (5)


packages/language/src/linking/error.ts
```diff
@@ -135,7 +135,21 @@ export class LinkerErrorReporter {
...
+  ) {
+    function fullName(symbol: QualifiedSyntaxNode): string[] {
+      let current: QualifiedSyntaxNode | null = symbol;
+      const parts: string[] = [];
+      while (current) {
+        parts.unshift(current.name);
+        current = current.parent;
+      }
+      return parts;
+    }
```
msujew, line 152: Suggestion: We should ensure here that the quick fix actually helps resolve the ambiguity. For example, in the following, the quick fix will not help:

```
DCL 1 A, 2 B FIXED;
DCL 1 A, 2 B FIXED;
B = 1; <-- B cannot be "fixed" by changing to A.B
```

packages/language/src/linking/error.ts
```diff
@@ -135,7 +135,21 @@ export class LinkerErrorReporter {
   /**
    * S IBM1881I
    */
-  reportAmbiguousReference(reference: Reference, name: string) {
+  reportAmbiguousReference(
+    reference: Reference,
+    name: string,
+    symbols: QualifiedSyntaxNode[],
+  ) {
+    function fullName(symbol: QualifiedSyntaxNode): string[] {
```
msujew, line 144: Suggestion: Can we make this function return the `AmbiguousReferenceData` type? This would ensure that we get a compilation error here in case the type changes.

packages/language/src/linking/error.ts
```diff
@@ -135,7 +135,21 @@ export class LinkerErrorReporter {
...
+      let current: QualifiedSyntaxNode | null = symbol;
+      const parts: string[] = [];
+      while (current) {
+        parts.unshift(current.name);
+        current = current.parent;
+      }
+      return parts;
```
msujew: Minor: `unshift` is a rather expensive operation (since it has to move the whole array every time, can we do `push` and `reverse` instead?

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -141,11 +141,72 @@ export function quickFixUppercaseText(
...
+  if (names.size !== current.length) {
+    return [];
+  } else {
+    const actions: CodeAction[] = [];
+    for (const [, fullName] of current.filter(
+      ([, name], index) =>
+        !current.some(
+          ([_, n], i) => i !== index && (n.endsWith("." + name) || n === name),
+        ),
+    )) {
```
msujew: Suggestion: Can we do something to improve the readability? Ideally throughout the whole method, but this part is particularly difficult to understand.

msujew (commented):
Cool, good to see that we're able to do this with so little code :)

packages/language/src/linking/error.ts
Lotes: That is already working. You will get no quick fix in this case

packages/language/src/linking/error.ts
Lotes: The function is called per element. But I put the type info somewhere close to this

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Lotes: Yes sure. It is really a pain to read.
I added a comment and better naming of the partial functions.
It filters out all names that are a suffix of different member.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -141,11 +141,76 @@ export function quickFixUppercaseText(
...
+    //We filter out all real suffixes. If all members are suffixes of each other, there are no suggestions.
+    const isSuffixOf = (suffix: string, str: string) =>
+      str.endsWith("." + suffix) || str === suffix;
+    const whereNameIsNotASuffix = (name: string, index: number): boolean =>
+      !current.some(([_, n], i) => i !== index && isSuffixOf(name, n));
+    for (const [, fullName] of current.filter(([, name], index) =>
+      whereNameIsNotASuffix(name, index),
+    )) {
+      const action: CodeAction = {
+        title: `Change "${fullName}"`,
```
msujew: Suggestion: The original text from the issue "Change <text>" implies to me that we change something from \<text\> to \<some-other-text\>. But we specifically want to change to the new text value.

```suggestion
        title: `Change to "${fullName}"`,
```

msujew (approved):
Thanks looks good to me, just one minor issue, see below.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Lotes: Yes, agree. I had it like this before but VS Code is cutting off the last option label, so I tried a bit. But with actually no success. I can change it back.

PR #556, opened 2026-01-28
Update case upper validation logic

ssmifi:
Solution for https://github.com/zowe/zowe-pli-language-support/issues/544 4 CASE(UPPER)

- As discussed, the `MACRO('CASE(UPPER)')` validation is now triggered when the corresponding directive is present in the source file. For this, the compiler option translator had to be extended to recognize the origin of the options. Before the change, it was more or less agnostic about whether it originated from the plugin configuration or the source files.
- The whole validation can still be turned off by setting `case-upper-validation` to false in the `pli-options`.
- I took the opportunity to clean up `extractCompilerOptions` a bit.
  - The top level translator itself if now wrapped in a class so that different sender can just incrementally translate rules. Note that the single translator rules for pli, macro, and sql are still residing in single objects each. 
  - The diagnostics are handled by the translator itself and it provides functions to anchor them to the source file instead of the post-processing.
  - The input text processing only uses a single substring for the text after the directives, which should improve performance when there are multiple directives.
  - Removed the translation in the `plugin-configuration-provider` as it was not really used anyways. It is enough when the configuration rules are parsed here and translated later. 
- The `test-builder` uses the config's defaults by default now. If a test needs a particular set of options, it should set them in the test files.

_edit_: As requested: The macro case upper option considers all input text, hence also affecting non-preprocessor code.

### DIFF
--- a/packages/language/src/validation/macro/case.ts
+++ b/packages/language/src/validation/macro/case.ts
@@ -14,14 +14,9 @@ import * as AST from "../../syntax-tree/ast";
 import { ValidationAcceptor } from "../validator";
 import { CompilationUnit } from "../../workspace/compilation-unit";
 import { CompilerOptions as MacroCompilerOptions } from "../../preprocessor/compiler-options/options-macro";
-import { Token } from "../../parser/tokens";
-import {
-  TraversalState,
-  traverseAllNodes,
-} from "../../syntax-tree/ast-iterator";
-import { PreprocessorTokens } from "../../preprocessor/pli-preprocessor-tokens";
-import { MultiMap } from "../../utils/collections";
 import { LspCodes } from "../lsp-codes";
+import { PreprocessorTokens } from "../../preprocessor/pli-preprocessor-tokens";
+import { stringToken } from "../../preprocessor/compiler-options/parser";
 
 export function MACRO_Case(
   node: AST.Program,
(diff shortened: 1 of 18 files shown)


packages/language/src/validation/macro/case.ts
```diff
@@ -48,7 +49,7 @@ export function MACRO_Case(
 
   const processedTokens = new Set<Token>();
 
-  traverseAllNodes(compilationUnit.preprocessorAst, (child) => {
+  traverseAllNodes(node, (child) => {
```
msujew: Question: It makes sense to fix it that way, but I'm wondering why we need `traverseAllNodes` in the first place. AFAIK we are supposed to iterate through all tokens anyway, right? What is the difference between doing this and simply iterate over `getAllTokens()`?

msujew (commented):
Works well, thank you 👍 Just one question, see below.

packages/language/src/validation/macro/case.ts
ssmifi: Good point. It used to traverse only over specific ast nodes, but we are visiting all tokens now. I noticed that the tokens from the compiler options are still the original chevrotain tokens without set `originalImage` field and different token identifiers. Maybe we should overhaul this in the future, but I wouldn't want to refactor that now before the release. Therefore, I added a special case in the validation to recognize this. wdyt?

msujew (approved):
Thanks, looks good to me 👍

packages/language/src/validation/macro/case.ts
msujew: Yep, seems reasonable. Thank you!

Issue #535, opened 2026-01-13
[Improvement] Refactor plugin config parsing


msujew:
Currently, we're simply using `JSON.parse` to parse the plugin config values. This works well, but prevents us from displaying the issues of the plugin config directly within the plugin config. We should use something like [`jsonc-parser`](https://www.npmjs.com/package/jsonc-parser) to properly retrieve the position information for diagnostics display.

ssmifi:
I leave this question here. Perhaps we can agree on this before the refactoring.

> Is it (still) necessary to maintain both compiler-options and pli-options in the plugin config? For me, this was just another layer of confusion. If not, we should maybe settle on the more convenient syntax and drop the other one.

msujew:
@ssmifi We agreed to drop `pli-options` some time ago.

PR #557, opened 2026-01-30
Remove pli-options from plugin config

ssmifi:
As confirmed earlier (see https://github.com/zowe/zowe-pli-language-support/issues/535), this one removes the old `pli-options` from the plugin config.


msujew (approved):
Thanks 👍

PR #558, opened 2026-02-01
Implement inconly

ssmifi:
Implement `incOnly` as requested in https://github.com/zowe/zowe-pli-language-support/issues/544 4. The normal preprocessor parsing should be disrupted as little as possible.

### DIFF
--- a/packages/language/src/parser/parser-entry.ts
+++ b/packages/language/src/parser/parser-entry.ts
@@ -25,6 +25,7 @@ import {
   isCicsExecStatement,
   isCicsResponseStatement,
 } from "./cics-parser";
+import { CompilerOptions } from "../preprocessor/compiler-options/options";
 
 export type PreprocessorParserResult = {
   statements: ast.Statement[];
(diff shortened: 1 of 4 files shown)


packages/language/src/parser/parser-entry.ts
```diff
@@ -34,20 +35,39 @@ export type PreprocessorParserResult = {
...
   state: ParserState,
+  compilerOptions?: CompilerOptions,
 ): PreprocessorParserResult {
   const statements: ast.Statement[] = [];
 
+  // Select the appropriate token type index for preprocessor statements
+  // depending on the INCONLY option.
+  const incOnly = compilerOptions?.macroOptions?.incOnly;
+  const parseTokenTypeIdx = incOnly ? NaN : t.Percent.tokenTypeIdx;
+  const parseTokenTypeIdxWithIncOnly = incOnly ? t.Percent.tokenTypeIdx : NaN;
```
msujew: Suggestion: I believe this kind of implementation can become pretty complicated, pretty quickly if there are more conditional preprocessor parser branches. Especially because the `SQL` and `CICS` stuff is supposed to be controlled by compiler options as well. What do you think about the following?

```ts
// null indicates a failure to parse, undefined that this should parse a token statement instead
type StatementParser = (state: ParserState) => ast.Statement | null | undefined;

function generateStatementParser(compilerOptions: CompilerOptions): StatementParser {
  const branches: StatementParser[] = [];
  // ...conditionally add branches here based on compiler options
  return (state) => {
    for (const branch of branches) {
       const result = branch(state);
       if (result !== undefined) {
         return result;
       }
     }
     return undefined;
  };
}
```

And then we can simply generate the statement parser at this point in the function and call it within the loop. Seems much cleaner to me.

msujew (commented):
Works well, but I believe there's a better way to do this.

msujew (approved):
Thanks, works well and is much cleaner!

PR #561, opened 2026-02-02
Support builtin files

msujew:
Addresses (1) of https://github.com/zowe/zowe-pli-language-support/issues/544.

Adds the two builtin files to the list of builtins.


ssmifi (approved):
Thanks! In general, I would be happy if we could avoid noDiagnostics at all, but I suppose it is reasonable here.

PR #562, opened 2026-02-02
Fix OR operator lexing

msujew:
Removes the erroneous `context.index++` and lets the `advance` handle everything.

### DIFF
--- a/packages/language/test/compiler/lexing.test.ts
+++ b/packages/language/test/compiler/lexing.test.ts
@@ -43,7 +43,7 @@ describe("Lexer uses compiler options", () => {
   test("Should lex correct OR without compiler option", async () => {
     const doc = await parseStmts(`
   DECLARE VAR FIXED;
-  VAR = 1 | 2;
+  VAR = 1|2;
   VAR |= 2;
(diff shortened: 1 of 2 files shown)


packages/language/test/compiler/lexing.test.ts
```diff
@@ -43,7 +43,7 @@ describe("Lexer uses compiler options", () => {
   test("Should lex correct OR without compiler option", async () => {
     const doc = await parseStmts(`
   DECLARE VAR FIXED;
-  VAR = 1 | 2;
+  VAR = 1|2;
```
ssmifi, line 46: Did you remove the spaces on purpose? Should work either way, right? Should we test both ways?

ssmifi (approved):
Works, thanks. Just one minor question.

packages/language/test/compiler/lexing.test.ts
msujew: Funnily enough, removing the spaces is the test. Previously, this would have resulted in a parser error, since the `|` "swallowed" the `2`. Now, it is working as expected.

Issue #520, opened 2025-12-15
[Parser] Clarify `PUT EDIT` arguments


msujew:
In our example we have seen a `PUT` statement like this:

```
PUT SKIP(2) EDIT
     ('---------- ENTRY TABLE ----------')
     (A)
     ('   I ', CHAR('NAME', 33),
      'TYPE       DCL  USE1  USE2 CALL1 CALL2')
     (SKIP, A, A, A)
     ((I, ENTRY(I) DO I = 0 TO ENTRY_TABLE_COUNT))
     (SKIP, F(4), X(1), A(33), A(8), (5) F(6), X(1), A, X(1), A);
```

Take special notice of the second-to-last line:

```
((I, ENTRY(I) DO I = 0 TO ENTRY_TABLE_COUNT))
```

As per the language reference manual, each entry in the outer `()` can be either a `data-list item` (meaning something like a variable - an expression from the parser perspective) or a `data-list type-3-do` wrapped inside of another set of `()`:

<img width="299" height="128" alt="Image" src="https://github.com/user-attachments/assets/f3645abd-f158-4cf4-bc28-1ce04661c6b4" />

This implies that the line above should actually be:

```
(I, (ENTRY(I) DO I = 0 TO ENTRY_TABLE_COUNT))
```

Notice that the `(` has moved from before the `I` to the `ENTRY` variable. However, the compiler has absolutely no issues compiling both of these variants. Our parser will currently stop at the `,`, which it claims to be an unexpected symbol - correct as per the language reference.

### Other Examples:

The language reference gives further examples, but they all follow the expected format and are parsed correctly by our implementation:

```
put list (Card_Rate, Dynamic_Flow);
put list ((Thickness(Distance) do Distance = 1 to 1000));
put list (P, Z, M, R);
put list (A*B/C, (X+Y)**2);
```

(from page 314 of the language reference manual)

### Question:

What is the `((I, ENTRY(I) DO I = 0 TO ENTRY_TABLE_COUNT))` line supposed to express. It is semantically the same as the other one, or does it hold special meaning?

msujew:
cc @asatklichov @VitGottwald this is one of the last missing pieces for full tolerance of our examples. It would be great if we can clarify this somehow :)

asatklichov:
resolved

PR #563, opened 2026-02-02
Fix parsing of PUT data specifications

msujew:
Related to https://github.com/zowe/zowe-pli-language-support/issues/520. It is still not clear what exactly something like `((I, ENTRY(I) DO I = 0 TO ENTRY_TABLE_COUNT))` means as a semantic construct, but this change adjusts the parser to support the syntax.

I'll keep the issue open so that we can clarify the semantics of this syntax in a separate change.


ssmifi (approved):
LGTM

PR #564, opened 2026-02-02
Fix locator call in assignment lookahead

msujew:
I noticed during testing of some other parser issues that statements such as `A->B = C` don't parse correctly due to missing handling of the `->` operator (locator call) in the lookahead.

This change adds operator to the lookahead logic and improves handling of the lookahead in general by combining multiple tokens into a single *lookahead table*.


ssmifi (approved):
LGTM

PR #565, opened 2026-02-03
Support optional package end statement

### LINKED ISSUE #560: PACKAGE statement need not be terminated by END
Even though this is not documented in the Language Reference diagram for the PACKAGE statement the END is optional.

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/560.

Simply makes the `END` statement for package statements optional.

Also refactors our `END` statement parsing a bit; We previously parsed the semicolon afterwards in the calling rule, instead of in the `END` statement itself. This didn't really make a lot of sense and this change simply moves the parsing into the statement. This simplifies some of the parsing code but required a minor change in the document symbol builder.


ssmifi (approved):
LGTM

PR #569, opened 2026-02-04
[Improvement] Fix include resolutions on files outside the workspace

Wagner-Laranjeiras:
`getProcessGroupConfigFromLib` does not correctly handle libraries provided as absolute paths, causing valid files to be missed.

ex.: For example, the URI path `"C:/pli/app-ext"` does not match the `"app-ext"` lib.

`getProgramConfig` suffers from a similar issue, where, in it's case, the program configuration can't be computed.

ex.: the program URI `"file:///Users/mockUser/projects/app/bin/app-ext.pli"` would never match, because it's stored in the configuration such as: `"file:///workspace/Users/mockUser/projects/app/bin/app-ext.pli"`

This PR adds explicit absolute-path detection and updates the lookup logic to correctly resolve such libraries and programs.

### DIFF
--- a/packages/language/src/workspace/plugin-configuration-provider.ts
+++ b/packages/language/src/workspace/plugin-configuration-provider.ts
@@ -696,10 +696,16 @@ export class PluginConfigurationProvider {
     //   config key:   file:///workspace/Users/alice/projects/app/bin/app-ext.pli
     // Without this fallback, the config would not be found.
     for (const [pattern, config] of this.programConfigs.entries()) {
-      const progConfigPath = uri.toLowerCase().includes("file:///")
+      let encodedProgConfigPath = uri.toLowerCase().includes("file:///")
         ? uri.replace("file:///", "")
         : uri;
-      if (pattern.endsWith(progConfigPath)) return config;
+      encodedProgConfigPath = encodedProgConfigPath.replace(
+        /^([a-zA-Z]):/,
+        (_, drive) => encodeURIComponent(drive + ":"),
+      );
+      encodedProgConfigPath = encodedProgConfigPath.toLowerCase();
+      const newPattern = pattern.replace(/%5C/gi, "/").toLowerCase();
+      if (newPattern.endsWith(encodedProgConfigPath)) return config;
     }
     // no match
     return undefined;
--- a/packages/language/src/workspace/plugin-configuration-provider.ts
+++ b/packages/language/src/workspace/plugin-configuration-provider.ts
@@ -595,9 +595,9 @@ export class PluginConfigurationProvider {
 
   /**
    * Sets the program configs of this plugin configuration provider, overwriting any existing configs.
-   * The config key is set as the full path relative to the workspace.
+   * Program paths are normalized and resolved relative to the workspace (unless absolute).
    * Post-processes the program configs after setting them, to ensure abstract options are built.
-   * @param workspacePath The full workspace path (used as a prefix for program config keys)
+   * @param workspacePath The full workspace path (used as base for resolving relative program paths)
    * @param programConfigs Program configs loaded from .pliplugin/pgm_conf.json (when present)
    */
   public setProgramConfigs(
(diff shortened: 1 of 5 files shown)


packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -721,8 +721,18 @@ export class PluginConfigurationProvider {
   public getProcessGroupConfigFromLib(libUri: URI): ProcessGroup | undefined {
     const dirname = UriUtils.basename(UriUtils.dirname(libUri));
     for (const config of this.processGroupConfigs.values()) {
-      if (config.$computedLibsSet.has(dirname)) {
-        return config;
+      for (const lib of config.$computedLibsSet) {
+        const isPathAbsolute = UriUtils.isWindows
```
ssmifi: Can we test this? Preferably os-independent? If we really need the `isWindows` check here test here, please provide a mock for the test to also test the windows paths on different os.

ssmifi (commented):
Thanks. Can we test this?

_edit_: Sorry, the draft update didn't reach me in time. So, still the same question for the PR then. :)

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: We could not use the `isWindows`, but the drive check would have to run even for Unix environments, which was not the original idea for it. Please let me know what do you think. 😉

packages/language/test/lsp/get-prog-config.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+describe("Check if `getProgramConfig` inside `PluginConfigurationProvider` retrieve proper values.", () => {
+  (test("atches workspace-relative program - most common use case", async () => {
+    await setupConfig([{ program: "*.pli", pgroup: "default" }]);
+    const testUri = URI.parse("workspace/a.pli");
+    const config =
+      PluginConfigurationProviderInstance.getProgramConfig(testUri);
+
+    expect(config).toBeDefined();
+  }),
```
msujew: Formatting:
```suggestion
  test("Matches workspace-relative program - most common use case", async () => {
    await setupConfig([{ program: "*.pli", pgroup: "default" }]);
    const testUri = URI.parse("workspace/a.pli");
    const config =
      PluginConfigurationProviderInstance.getProgramConfig(testUri);

    expect(config).toBeDefined();
  }),
```

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -684,10 +687,49 @@ export class PluginConfigurationProvider {
...
+   * @param searchUri - The file URI to normalize
+   * @param pattern - The search pattern to normalize
+   * @returns Object containing normalized URI and pattern
+   */
+  private normalizeSearchPatterns(searchUri: string, pattern: string) {
+    const normalizedPattern = pattern.replace(/%5C/gi, "/").toLowerCase();
+
+    const uriWithoutProtocol = searchUri.toLowerCase().includes("file:///")
+      ? searchUri.replace("file:///", "")
+      : searchUri;
```
msujew: Question: This will strip the leading `/` of unix-style paths. Is this what we want? Why? Note that the schema is only the `file://` part.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -684,10 +687,49 @@ export class PluginConfigurationProvider {
...
+    //   program URI:  file:///Users/mockUser/projects/app/bin/app-ext.pli
+    //   config pattern:   file:///workspace/Users/mockUser/projects/app/bin/app-ext.pli
+    // Without this fallback, the config would not be found.
+    for (const [pattern, config] of this.programConfigs.entries()) {
+      const { normalizedUri, normalizedPattern } = this.normalizeSearchPatterns(
+        uri,
+        pattern,
+      );
+      if (normalizedPattern.endsWith(normalizedUri)) return config;
+    }
```
msujew: Suggestion: Instead of trying to work around the fact that we store program configs in a rather weird format, wouldn't it be better to fix the format instead of adding more workarounds? What is the reason for this format?

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -721,8 +763,19 @@ export class PluginConfigurationProvider {
...
+          UriUtils.processDriveLetter(normalizedLib).drive ||
+          !UriUtils.isPathRelative(normalizedLib);
+
+        const matches = isPathAbsolute
+          ? normalizedLib.endsWith(dirname)
+          : normalizedLib === dirname;
+
+        if (matches) {
+          return config;
+        }
```
msujew: Suggestion: Doing this defeats the purpose of using a `Set` in the first place (constant time lookups). We should probably normalize the `$computedLibsSet` before storing it instead.

packages/language/test/lsp/get-prog-config.test.ts
```diff
@@ -0,0 +1,78 @@
...
+async function setupConfig(
+  programConfig: { program: string; pgroup: string }[],
+) {
+  const pluginConfig = new PluginConfigurationProvider();
+  setPluginConfigurationProvider(pluginConfig);
+  pluginConfig.setProgramConfigs("/workspace", programConfig);
+}
+
+describe("Check if `getProgramConfig` inside `PluginConfigurationProvider` retrieve proper values.", () => {
+  (test("Matches workspace-relative program - most common use case", async () => {
```
msujew: Formatting:

```suggestion
  test("Matches workspace-relative program - most common use case", async () => {
```

msujew (approved):
Thanks, looks good to me 👍

PR #572, opened 2026-02-05
[Bugfix] Fix search for DD members without DD names

Wagner-Laranjeiras:
Related to PR #551 

**Problem:** Currently on Windows machines, members without DD names (ex.: ddname(member)) can't be resolved. This happens because the path we feed to the glob search.

**Fix:** Replacing the double slashes ("\\") with one forward slash in a Unix like path ("/") gives the glob search the means to return a match.

### DIFF
--- a/packages/vscode-extension/src/language/main.ts
+++ b/packages/vscode-extension/src/language/main.ts
@@ -73,9 +73,12 @@ class NodeFileSystemProvider implements FileSystemProvider {
         nocase: true,
       });
     } else {
+      // Windows paths may contain double slashes, which can prevent blob search from finding matches.
+      // This mainly affects INCLUDEs for members without DD names (ddname(member)).
+      const forwardSlashPath = options.dirPath.fsPath.replace(/\\/g, "/");
       // member lookup w/ dir path
       files = await glob.glob(
-        `${options.dirPath.fsPath}**/*\\(${options.member}\\)`,
+        `${forwardSlashPath}**/*\\(${options.member}\\)`,
         {
           nodir: true,
           absolute: true,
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/language/main.ts
```diff
@@ -73,9 +73,12 @@ class NodeFileSystemProvider implements FileSystemProvider {
         nocase: true,
       });
     } else {
+      // Windows paths may contain double slashes, which can prevent blob search from finding matches.
+      // This mainly affects INCLUDEs for members without DD names (ddname(member)).
+      const forwardSlashPath = options.dirPath.fsPath.replace(/\\/g, "/");
```
msujew: Suggestion: The comment isn't quite correct. `\\` is not a double slash, it's just a way to escape the `\` character. Please adjust the comment to and the variable name, since what is essentially happening here is *normalization*:
```suggestion
      // Windows paths contain backslashes that need to be normalized before using them as glob patterns.
      // This mainly affects INCLUDEs for members without DD names (ddname(member)).
      const normalizedPath = options.dirPath.fsPath.replace(/\\/g, "/");
```

Wagner-Laranjeiras: Yep, I was just having this conversation with Steven. I messed up the nomenclature there 😅 Thanks for clarifying, I'll update it.

Wagner-Laranjeiras: I'll let the "may" there though, as in some cases, the libs can be written on the configuration already with the slash (ex.: C:/User/example).

msujew (approved):
Thanks, LGTM 👍

Issue #570, opened 2026-02-04
Findings 02-02-26


asatklichov:
- [x] (1) Outline support is not provided on opened file once we open VSCode first time or re-load it. We need to either close and re-open file or change the tab.  See the provided recording 

- [x] (2) Errors from previously closed program are appearing when we do changes in proc_grps.json, see the provided recording

- [x] (3) Errors or warnings are hidden once we end up values with "," as in the snapshot.
```

%process aggregate;  
  HELLO:
    PROCEDURE OPTIONS(MAIN);
    /*A PROGRAM TO OUTPUT HELLO WORLD*/  ;
    PUT LIST ('HELLO, WELCOME TO PLI/1 WORLD ');

  END HELLO;
``` 

<img width="399" height="156" alt="Image" src="https://github.com/user-attachments/assets/ec37f7aa-90e6-41c6-be8f-c1fbf68a0987" />

 
- [x] (4)   Continuation of Green code support, see attached files

 
- [x] (5)   Regression issue regarding environment variables 

```
 HELLO:
   PROCEDURE OPTIONS(MAIN);

   dcl repdsn file record output env(fb, recsize(80), total); 

   dcl sysprint file output;   
   
 END HELLO;
```

or 
```

 DCL PARTS          FILE RECORD INPUT ENV(FB,RECSIZE(180),TOTAL);
 DCL SYSPRINT       FILE OUTPUT;    
```

- [x] (6) If we remove label from end statement then it does not show warning 'The PROCEDURE NAMEZ is not referenced.' which was shown correctly before. 

```

 HELLO:
   PROCEDURE OPTIONS(MAIN);
   /*A PROGRAM TO OUTPUT HELLO WORLD*/  ;
   PUT LIST ('HELLO, WELCOME TO PLI/1 WORLD '); 
 END HELLO;

 Namez: procedure;
 end Namez; 
```

when we remove label warning is not shown. 

<img width="249" height="59" alt="Image" src="https://github.com/user-attachments/assets/86e29a68-123f-4df5-9b2d-308eede35ac6" />

   - [x] (7)  Regression error regarding [type changes ]([url](https://github.com/zowe/zowe-pli-language-support/issues/392)),  
 see the below error message on variable 'b'

```
 MYPROC: PROCEDURE OPTIONS (MAIN); 
    dcl a char(10);
    dcl b like a;/*The LIKE reference is neither a structure nor a union*/
    a = 5; 
    b = 3;   
  END MYPROC;
[...]

PR #573, opened 2026-02-05
Remove completely closed compilation units

msujew:
Fixes (2) of #570. The fix is rather trivial: If a file closes, check whether all files of that compilation unit are now closed - if they are, completely discard the compilation unit. This removes the whole compilation unit from memory and prevents the error seen in the issue.

This can be manually tested by opening a workspace with a `.pliplugin` config directory:
1. Open, edit (making sure that the file has an error diagnostic) and then close a `.pli` file associated to a config group in the `.pliplugin` directory.
2. The errors should disappear from the problem view.
3. Edit the program config - the errors should not appear again in the view.

### DIFF
--- a/packages/language/src/workspace/compilation-unit.ts
+++ b/packages/language/src/workspace/compilation-unit.ts
@@ -310,17 +310,13 @@ export class CompilationUnitHandler {
       // Nothing to close
       return false;
     }
-    let isOpen = false;
     for (const file of unit.services.files.keys()) {
       if (EditorDocuments.has(file)) {
-        isOpen = true;
-        break;
+        // Return early if any file is still open
+        return false;
       }
     }
-    if (!isOpen) {
-      return this.deleteCompilationUnit(uri);
-    }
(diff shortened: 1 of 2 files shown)


packages/language/src/workspace/compilation-unit.ts
```diff
@@ -288,13 +288,41 @@ export class CompilationUnitHandler {
...
+    const unit = this.compilationUnits.get(uri.toString());
+    if (!unit) {
+      // Nothing to close
+      return false;
+    }
+    let isOpen = false;
+    for (const file of unit.services.files.keys()) {
+      if (EditorDocuments.has(file)) {
+        isOpen = true;
+        break;
```
ssmifi: _Minor_: Couldn't we return here immediately and remove the `isOpen` alltogether?

ssmifi (approved):
Looks good. I just have a minor remark to simplify `tryCloseCompilationUnit`.

PR #576, opened 2026-02-09
Add TOTAL in ENVIRONMENT attribute 

Lotes:


code_samples/environment.pli
```diff
@@ -0,0 +1,11 @@
...
+   PROCEDURE OPTIONS(MAIN);
+
+   dcl repdsn file record output env(fb, recsize(80), total); 
+
+   dcl sysprint file output;   
+   
+ END HELLO;
+
+  DCL PARTS          FILE RECORD INPUT ENV(FB,RECSIZE(180),TOTAL);
+  DCL SYSPRINT       FILE OUTPUT;
```
msujew: Suggestion: I don't think we need a new sample file for every token/parser change. If the tests are not enough, then we should improve the test suite.

msujew (approved):
Looks good to me, just one suggestion, see below.

PR #581, opened 2026-02-09
[Improvement] Launch config public service

Wagner-Laranjeiras:
While we don't deal with multiple `.pliplugin` folders (which I understand we are going to tackle soon), let's consider doing this, even if for just a few weeks.

Jokes aside, this simple PR save us from adding `"/plugin-example"` to the launch config every single day. 😄


msujew (approved):

ssmifi:
Please set a reasonable commit message when merging.

PR #582, opened 2026-02-09
Add diagnostic when procedure is not used

Lotes:
Before this fix only cross-referenced symbols were checked. The result was that symbols without any usage were hidden - so no diagnostic.
Now the code checks all symbols.

### DIFF
--- a/packages/language/src/validation/validator.ts
+++ b/packages/language/src/validation/validator.ts
@@ -182,7 +182,6 @@ export function linkingErrorsToDiagnostics(
   // Warn if a label is never referenced
   for (const scope of scopeCaches.regular.values()) {
     nodeLoop: for (const node of scope.symbolTable.nodeLookup.keys()) {
-      const nodeReferences = references.findReferences(node);
       if (node.kind !== SyntaxKind.LabelPrefix) {
         continue;
       }
@@ -203,6 +202,7 @@ export function linkingErrorsToDiagnostics(
       }
 
       // Ignore all `END` nodes, since a procedure will always have one `END` node referencing itself
+      const nodeReferences = references.findReferences(node);
       for (const reference of nodeReferences) {
         if (reference.owner.container?.kind !== SyntaxKind.EndStatement) {
           // The label is referenced, so we don't generate a warning
--- a/packages/language/src/validation/validator.ts
+++ b/packages/language/src/validation/validator.ts
@@ -180,7 +180,10 @@ export function linkingErrorsToDiagnostics(
   }
 
   // Warn if a label is never referenced
-  for (const [node, nodeReferences] of references.allReverseReferences()) {
+  const symbolReferences = [...scopeCaches.regular.values()]
+    .flatMap((scope) => [...scope.symbolTable.nodeLookup.keys()])
+    .map(sym => [sym, references.findReferences(sym)] as const);
+  for (const [node, nodeReferences] of symbolReferences) {
     if (node.kind !== SyntaxKind.LabelPrefix) {
       continue;
     }
--- a/packages/language/src/validation/validator.ts
+++ b/packages/language/src/validation/validator.ts
@@ -182,7 +182,7 @@ export function linkingErrorsToDiagnostics(
   // Warn if a label is never referenced
   const symbolReferences = [...scopeCaches.regular.values()]
     .flatMap((scope) => [...scope.symbolTable.nodeLookup.keys()])
-    .map(sym => [sym, references.findReferences(sym)] as const);
+    .map((sym) => [sym, references.findReferences(sym)] as const);
   for (const [node, nodeReferences] of symbolReferences) {
     if (node.kind !== SyntaxKind.LabelPrefix) {
       continue;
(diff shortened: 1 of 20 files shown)


packages/language/src/validation/validator.ts
```diff
@@ -180,7 +180,10 @@ export function linkingErrorsToDiagnostics(
   }
 
   // Warn if a label is never referenced
-  for (const [node, nodeReferences] of references.allReverseReferences()) {
+  const symbolReferences = [...scopeCaches.regular.values()]
+    .flatMap((scope) => [...scope.symbolTable.nodeLookup.keys()])
+    .map((sym) => [sym, references.findReferences(sym)] as const);
```
msujew: Suggestion: This looks **very** expensive. Please don't manifest the values from the scope in an array. This will reintroduce the issue fixed in https://github.com/zowe/zowe-pli-language-support/pull/553.

packages/language/test/fourslash/preprocessor/procedures/answer/answer-page.ts
```diff
@@ -15,9 +15,9 @@
...
 //// %END;
 //// %ACTIVATE MYMACRO;
-//// ppp: PROC;
+//// ppp: PROC OPTIONS(MAIN);
 ////   MYMACRO
 //// END;
 
-preprocessor.expectTokens("ppp: PROC; END;");
+preprocessor.expectTokens("ppp: PROC OPTIONS(MAIN); END;");
 verify.noDiagnostics();
```
msujew, line 21: Suggestion: I didn't notice it back then, but can we simply remove the whole `ppp` procedure and call `MYMACRO` on the root level? This makes the test much more readable and less prone to issues in the future. This applies to the other test files.

packages/language/src/validation/validator.ts
Lotes: Do we have a cheaper source of syntax nodes for one compilation unit? Traversing the AST sounds more painful.
Now I replaced the arrays by for loops.

Lotes: A different perspective could be: The mainframe does not show any diagnostics for the current scenario. Why should we?

packages/language/src/validation/validator.ts
```diff
@@ -37,8 +37,8 @@ export type ValidationFunction<T extends SyntaxNode> = (
 
 export type ValidationChecks = Partial<{
   [K in keyof typeof SyntaxKind as (typeof SyntaxKind)[K] extends SyntaxNode["kind"]
-    ? K
-    : never]: ValidationFunction<
+  ? K
+  : never]: ValidationFunction<
```
msujew: I believe this is failing the linter?

packages/language/src/validation/validator.ts
msujew: > Now I replaced the arrays by for loops.

Yes, that's much better, thank you. The issue is the array allocation (creating and then disposing the memory) and not the iteration.

> The mainframe does not show any diagnostics for the current scenario.

What do you mean with the "current scenario"? AFAIK the compiler will complain about declared variables and procedures that aren't referenced anywhere.

packages/language/src/validation/validator.ts
```diff
@@ -180,37 +180,40 @@ export function linkingErrorsToDiagnostics(
...
-    // Ignore all `END` nodes, since a procedure will always have one `END` node referencing itself
-    const actualReferences = nodeReferences.filter(
-      (reference) =>
-        reference.owner.container?.kind !== SyntaxKind.EndStatement,
-    );
+      // Ignore all `END` nodes, since a procedure will always have one `END` node referencing itself
+      const actualReferences = nodeReferences.filter(
+        (reference) =>
+          reference.owner.container?.kind !== SyntaxKind.EndStatement,
+      );
```
msujew: While you're already at this: Can you improve this so we don't allocate a new array every time? Instead, I would like to see a loop that continues the outer loop if it finds a non-end statement reference.

msujew (commented):
Thanks, much better.

packages/language/src/validation/validator.ts
Lotes: > What do you mean with the "current scenario"? AFAIK the compiler will complain about declared variables and procedures that aren't referenced anywhere.

I was checking the snippet in the mainframe and got no feedback about unused symbols. Let me have another look... *lookup* 

```
 Namez1: procedure;
 end Namez1; 

 Namez2: procedure;
 end;
```

The mainframe will not complain for this snippet.
But it could be true for other references, that are not procedure labels.

packages/language/src/validation/validator.ts
```diff
@@ -180,37 +180,37 @@ export function linkingErrorsToDiagnostics(
...
-  for (const [node, nodeReferences] of references.allReverseReferences()) {
-    if (node.kind !== SyntaxKind.LabelPrefix) {
-      continue;
-    }
+  for (const scope of scopeCaches.regular.values()) {
+    nodeLoop: for (const node of scope.symbolTable.nodeLookup.keys()) {
+      const nodeReferences = references.findReferences(node);
+      if (node.kind !== SyntaxKind.LabelPrefix) {
+        continue;
+      }
```
msujew: Minor: Reverse this to prevent the lookup:
```suggestion
      if (node.kind !== SyntaxKind.LabelPrefix) {
        continue;
      }
      const nodeReferences = references.findReferences(node);
```

msujew (approved):
Perfect, thank you!

PR #586, opened 2026-02-09
Provide semantic highlighting for all tokens

### LINKED ISSUE #579: Publish all semantic tokens
The textmate grammar we have does not address all PL/I syntax rules. To deal with that we use semantic tokens to fix certain situations. However, there will most likely always be some corner case that is properly handled in the parser proper but not in the textmate grammar.

How about publishing everything as a semantic token to make sure the syntax coloring reflects the parser's understanding of the code rather than a mixture of the texamate grammar + select tweaks?

Any performance concerns?

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/579

Attempts to provide highlighting for all text in the document. Can be theoretically tested on VS Code by removing the `pli.merged.json` after building the project - however, this seems to disable the `modifier` type highlighting (for some reason?).

Also fixes a minor issue in the compiler options lexer that led to highlighting bleeding into the second editor line.

### DIFF
--- a/packages/language/src/language-server/semantic-tokens.ts
+++ b/packages/language/src/language-server/semantic-tokens.ts
@@ -13,10 +13,7 @@ import { SemanticTokensBuilder } from "vscode-languageserver";
 import { CompilationUnit } from "../workspace/compilation-unit";
 import { Range } from "./types";
 import { TextDocument } from "vscode-languageserver-textdocument";
-import {
-  SemanticTokensLegend,
-  SemanticTokenTypes,
-} from "vscode-languageserver-types";
+import { SemanticTokensLegend } from "vscode-languageserver-types";
 import {
   DeclaredVariable,
   getContainer,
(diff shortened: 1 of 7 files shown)


msujew:
@VitGottwald Can you try testing this in neovim? It would be good to know how well this performs there.

packages/language/src/language-server/semantic-tokens.ts
```diff
@@ -114,40 +117,50 @@ function tokenType(token: Token): string | undefined {
...
-    token.element?.kind === SyntaxKind.TypeAttribute
-  ) {
-    return token.element.type?.node ?? undefined;
-  } else if (
-    (token.kind === CstNodeKind.HandleAttribute_TypeId0 ||
-      token.kind === CstNodeKind.HandleAttribute_TypeId1) &&
-    token.element?.kind === SyntaxKind.HandleAttribute
-  ) {
-    return token.element.type?.node ?? undefined;
+  switch (token.kind) {
```
ssmifi, line 145: The `cases` can fall through, which is probably not intended.

ssmifi (approved):
Looks nice. I think there is one fall through that is not intended though.

packages/language/src/language-server/semantic-tokens.ts
msujew: You're right - good catch!

PR #587, opened 2026-02-10
Add `GlobalMutex` for LSP operations

msujew:
Fixes issue (1) of https://github.com/zowe/zowe-pli-language-support/issues/570.

Re-Introduces a `GlobalMutex` to ensure that we do not accidentally respond to LSP requests before the compilation unit is ready to accept requests. This `GlobalMutex` does not actually wait until the lifecycle of specific compilation unit is done, but instead just ensures that LSP requests are not performed on non-existing compilation units that are still in the process of being created.

The original issue is consistently reproducible by applying the following patch to the `NodeFileSystemProvider`:

```diff
class NodeFileSystemProvider implements FileSystemProvider {
  async readFile(uri: URI): Promise<string> {
+   await new Promise((resolve) => setTimeout(resolve, 500));
    return fs.promises.readFile(uri.fsPath, "utf8");
  }
```

### DIFF
--- a/packages/language/src/workspace/compilation-unit.ts
+++ b/packages/language/src/workspace/compilation-unit.ts
@@ -41,7 +41,7 @@ import {
   ProgramConfig,
 } from "./plugin-configuration-provider.js";
 import { EvaluationResults } from "../preprocessor/instruction-interpreter.js";
-import { createMutex, Mutex } from "./mutex.js";
+import { createMutex, GlobalMutex, Mutex } from "./mutex.js";
 import { Deferred, isOperationCancelled } from "../utils/promises.js";
 import { InstructionCache } from "../preprocessor/instruction-cache.js";
 import { TextDocument } from "vscode-languageserver-textdocument";
@@ -220,7 +220,11 @@ export async function createCompilationUnit(
 export class CompilationUnitHandler {
   private compilationUnits: Map<string, CompilationUnit> = new Map();
   private connection!: Connection;
-  private ready = new Deferred();
+  private readyDeferred = new Deferred();
+
+  get ready(): Promise<void> {
+    return this.readyDeferred.promise;
+  }
 
   getCompilationUnit(uri: URI): CompilationUnit | undefined {
     return this.compilationUnits.get(uri.toString());
@@ -276,7 +280,7 @@ export class CompilationUnitHandler {
    * **Must** be called if `listen` has been called previously.
    */
   finalize(): void {
-    this.ready.resolve();
+    this.readyDeferred.resolve();
   }
 
   listen(connection: Connection): void {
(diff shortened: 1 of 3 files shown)


packages/language/src/workspace/compilation-unit.ts
```diff
@@ -384,16 +395,18 @@ export class CompilationUnitHandler {
...
   ): Promise<void> {
-    for (const unit of this.getAllCompilationUnits()) {
-      if (cancellationToken.isCancellationRequested) {
-        return;
-      }
-      unit.mutex.run(async (cancellationToken) => {
-        const textDocument = await TextDocuments.get(unit.uri.toString());
-        if (textDocument) {
-          this.process(unit, textDocument, connection, cancellationToken);
+    GlobalMutex.run(async () => {
```
ssmifi: _Question_: Can we safely fire-and-forget here? Or should we wait?

packages/language/src/workspace/compilation-unit.ts
```diff
@@ -276,7 +280,7 @@ export class CompilationUnitHandler {
    * **Must** be called if `listen` has been called previously.
    */
   finalize(): void {
-    this.ready.resolve();
+    this.readyDeferred.resolve();
```
ssmifi, line 288: _Suggestion_: Should we name the method more accurately? E.g., `markReady`?

packages/language/src/workspace/mutex.ts
```diff
@@ -71,3 +71,5 @@ class MutexImpl implements Mutex {
 export function createMutex(): Mutex {
   return new MutexImpl();
 }
+
+export const GlobalMutex = createMutex();
```
ssmifi: _Question_: Does the `GlobalMutex` basically protect the compilation unit map in the handler? Would it make sense to move the mutex into the handler?

packages/language/src/workspace/compilation-unit.ts
```diff
@@ -320,21 +324,28 @@ export class CompilationUnitHandler {
...
-    await this.ready.promise;
-    const unit = await this.getOrCreateCompilationUnit(uri);
-    if (!unit) {
-      // standalone library files do not synthesize new compilation units
-      return;
-    }
-    unit.mutex.run(async (cancellationToken) => {
-      const document = await EditorDocuments.get(unit.uri);
-      if (!document) {
+    await GlobalMutex.run(async () => {
```
ssmifi: _Question_: Should `tryCloseCompilationUnit` also be protected by the mutex? E.g., could a unit be added again through a delayed `process` after it was deleted?

ssmifi (commented):
Looks good. I just have a few questions to understand the locking better.

packages/language/src/workspace/compilation-unit.ts
msujew: Good point. I'm not entirely sure, but I'll add a mutex call anyway. It shouldn't hurt.

ssmifi (approved):
LGTM, Thanks!

PR #588, opened 2026-02-10
Support forward declarations using `ENTRY`

### LINKED ISSUE #534: [Linking] Respect `ENTRY` attribute on variable declarations
Right now, when you declare an entry variable and a corresponding procedure like this:

```
DCL TEST ENTRY (CHAR(10));
TEST: PROC (VAR);
  DCL VAR CHAR(10);
  PUT(VAR);
END;
```

You receive an `IBM1306IE` error:

```
Repeated declaration of TEST is invalid and will be ignored.
```

However, on the mainframe, as long as the entry declaration and the procedure do not live inside of the same local procedure, they do not produce any errors. Instead `DCL TEST ENTRY` acts as a "forward declaration" of the corresponding procedure.

As a result, during the linking process, we should check whether these two declarations can match, and in that case not show the diagnostic.

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/534

Adds two features to the language server:
1. Ambiguous references and duplicate declarations are no longer reported when encountering forward declarations with `ENTRY`. Meaning that a variable with `ENTRY` and a procedure with the same name are semantically the same element.
2. In addition to the `node` of a `Reference` object, the reference resolver can also now populate the `nodes` field (only in addition - not as an alternative). This field is used for LSP features to point to duplicate/ambiguous references to improve the user experience. This feature is also used for the feature in point 1, where we want to reference both the procedure as well as the `ENTRY` declaration.

Updates our test engine to support testing ambiguous references.

### DIFF
--- a/packages/language/src/syntax-tree/ast.ts
+++ b/packages/language/src/syntax-tree/ast.ts
@@ -657,11 +657,23 @@ export interface Reference<T extends SyntaxNode = SyntaxNode> {
   owner: SyntaxNode;
   text: string;
   token: Token;
+  /**
+   * This is the main target of the reference.
+   *
+   * If the reference hasn't been resolved yet, this will be `undefined`.
+   * If the reference resolution has failed (i.e. the reference doesn't actually point to any element in the code), this will be `null`.
+   * In all other cases, this will be the resolved target of the reference.
+   */
   node: T | null | undefined;
   /**
-   * PLI generally only allows a reference to resolve to a single element.
+   * PL/I generally only allows a reference to resolve to a single element.
    * However, for the purpose of the language server, we want to be able to reference multiple elements at once.
    * For example, in the case of ambiguous reference or when encountering a procedure and its forward declaration.
+   * This field is only really useful for selected LSP services - other features should use the `node` field instead.
+   *
+   * Note that the `nodes` might not always be filled, even when the `node` field is set.
+   * However, when the `nodes` are filled, they should always also contain the value of the `node` field as one of their elements.
+   * Use the `iterateReferenceNodes` helper to iterate over all possible nodes of a reference, regardless of how they are stored.
    */
   nodes: T[];
   type: ReferenceType;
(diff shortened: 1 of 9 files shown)


packages/language/src/syntax-tree/ast.ts
```diff
@@ -683,11 +689,20 @@ export function createReference<T extends SyntaxNode>(
...
     type,
   };
 }
 
+export function* iterateReferenceNodes(ref: Reference): Iterable<SyntaxNode> {
+  if (ref.nodes.length > 0) {
+    yield* ref.nodes;
+  } else if (ref.node) {
+    yield ref.node;
+  }
```
ssmifi, line 716: _Question_: Can `node` be set without also being included in `nodes`? Technically, they can be set separately from each other in `assignReference`, but in practice `getMatchingSymbols` will include the `node`, right? In the iterator, it looks like that we assume that `node` will also be included in `nodes`.

packages/language/src/syntax-tree/ast.ts
```diff
@@ -658,6 +658,12 @@ export interface Reference<T extends SyntaxNode = SyntaxNode> {
   text: string;
   token: Token;
   node: T | null | undefined;
+  /**
+   * PLI generally only allows a reference to resolve to a single element.
+   * However, for the purpose of the language server, we want to be able to reference multiple elements at once.
+   * For example, in the case of ambiguous reference or when encountering a procedure and its forward declaration.
+   */
```
ssmifi, line 677: _Suggestion_: Perhaps we should explicitly state in the comment that `node` will be also included in `nodes`.

ssmifi (commented):
Works. I find the `node`/`nodes` duality a bit tricky, but I guess it makes sense to store the primary reference separately. Can you please check the question if I understood it correctly? Thanks.

packages/language/src/syntax-tree/ast.ts
msujew: I've added some more documentation that should hopefully clear this up.

ssmifi (approved):

Issue #568, opened 2026-02-04
URI handling refactor


Wagner-Laranjeiras:
**Problem**

We are seeing inconsistencies in how URIs are built and resolved due to differences between:

- The actual file system (production) vs. the virtual file system (tests)
- UNIX-like environments vs. Windows environments

These discrepancies have led to the introduction of multiple helper functions that act as environment-specific workarounds. This approach increases complexity, makes the code harder to reason about, and is not sustainable long term.

**Goal**

We need a consistent and reliable way to construct URIs such that their paths are normalized uniformly across:

- Windows and UNIX platforms
- Real and virtual file systems (production and test environments)

By standardizing URI construction and normalization, we can eliminate the need for workaround helpers and ensure predictable behavior regardless of the underlying environment.

**Expected Outcome**

A single, well-defined strategy for URI/path normalization
Reduced platform-specific branching and helper functions
Improved maintainability and correctness across environments

Wagner-Laranjeiras:
Note: adding a helper for `.replace(/\\/g, "/")` could be a good idea.

PR #590, opened 2026-02-11
[Bugfix] Config creation via Quick Fix

Wagner-Laranjeiras:
Issue #570 , item 9.

**Problem:**
When creating a plugin folder via Quick Fix from a nested file, only the filename was added to the program field in the program configuration file. The intermediate folders between the workspace root and the target file were not included in the computed path. As a result, the generated program entry point was invalid, preventing the program from resolving correctly.

**Solution:**
This change ensures that the full relative path (from the workspace root to the target file) is properly computed and written to the program field, so the generated entry point resolves as expected.

**Follow up task** (already in progress):
A follow-up task will introduce support for creating additional program entries via Quick Fix. This will allow new entries to be added without overwriting or altering any existing program configuration, thereby preserving current setups.

**EDIT:**
The follow-up task has now also been incorporated into this PR. With this enhancement, we first check for any existing program configurations and, if present, append the new program configuration alongside the existing ones rather than replacing them.
This ensures that previously defined program entries are preserved while allowing additional configurations to be added seamlessly.

### DIFF
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -98,10 +98,15 @@ export async function quickFixResolveInclude(
 export async function quickFixCreateConfig(
   diagnostic: Diagnostic,
 ): Promise<CodeAction | undefined> {
-  const entryUri = UriUtils.basename(URI.parse(diagnostic.data.entryUri));
-  if (!diagnostic.data.entryUri || !entryUri) {
+  const workspace = PluginConfigurationProviderInstance.getWorkspacePath();
+  const entryUri = diagnostic.data.entryUri;
+  if (!workspace || !entryUri) {
     return undefined;
   }
+  const programPath = URI.parse(
+    diagnostic.data.entryUri.replace(workspace, ""),
+  ).path.replace(/^\//, "");
+
   const action: CodeAction = {
     title: `Create a plugin configuration folder for this file.`,
     kind: CodeActionKind.QuickFix,
@@ -109,7 +114,7 @@ export async function quickFixCreateConfig(
     command: {
       title: "Create configuration folder",
       command: Commands.CREATE_CONFIG,
-      arguments: [entryUri],
+      arguments: [programPath],
     },
   };
 
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -115,7 +115,7 @@ export async function quickFixCreateConfig(
     command: {
       title: "Create configuration folder",
       command: Commands.CREATE_CONFIG,
-      arguments: [programPath, entryUri],
+      arguments: [programPath],
     },
   };
 
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -126,11 +126,11 @@ export async function quickFixCreateConfig(
     : rawPath;
 
   const action: CodeAction = {
-    title: `Create an startup configuration for this file.`,
+    title: `Create a startup configuration for this file.`,
     kind: CodeActionKind.QuickFix,
     diagnostics: [diagnostic],
     command: {
-      title: "Create an startup configuration",
+      title: "Create a startup configuration",
       command: Commands.CREATE_CONFIG,
       arguments: [programPath],
     },
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -100,7 +100,7 @@ export async function quickFixCreateConfig(
 ): Promise<CodeAction | undefined> {
   const workspace = PluginConfigurationProviderInstance.getWorkspacePath();
   const entryUri = diagnostic.data.entryUri as string;
-  if (!workspace.length || !entryUri || !entryUri.length) {
+  if (!workspace || !entryUri) {
     return;
   }
   const resolvedEntry = entryUri.startsWith("file://")
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -107,12 +107,18 @@ export async function quickFixCreateConfig(
     ? URI.parse(entryUri).fsPath.replace(/\\/g, "/")
     : entryUri.replace(/\\/g, "/");
 
-  const normalizedWorkspace = workspace.replace(/\\/g, "/");
-  const workspacePrefix = normalizedWorkspace + "/";
+  const workspaceParts = workspace
+    .replace(/\\/g, "/")
+    .split("/")
+    .filter((e) => e.length > 0);
+  const entryParts = resolvedEntry.split("/").filter((e) => e.length > 0);
 
-  const isInsideWorkspace = resolvedEntry.startsWith(workspacePrefix);
+  const isInsideWorkspace = workspaceParts.every(
+    (part, index) => part === entryParts[index],
+  );
+
   const rawPath = isInsideWorkspace
-    ? resolvedEntry.slice(workspacePrefix.length)
+    ? entryParts.slice(workspaceParts.length).join("/")
     : resolvedEntry;
 
   const programPath = UriUtils.processDriveLetter(rawPath).drive
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -120,11 +120,11 @@ export async function quickFixCreateConfig(
     : rawPath;
 
   const action: CodeAction = {
-    title: `Create a plugin configuration for this file.`,
+    title: `Create an startup configuration for this file.`,
     kind: CodeActionKind.QuickFix,
     diagnostics: [diagnostic],
     command: {
-      title: "Create configuration folder",
+      title: "Create an startup configuration",
       command: Commands.CREATE_CONFIG,
       arguments: [programPath],
     },
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -107,8 +107,8 @@ export async function quickFixCreateConfig(
     ? URI.parse(entryUri).fsPath.replace(/\\/g, "/")
     : entryUri.replace(/\\/g, "/");
 
-  const workspaceParts = workspace
-    .replace(/\\/g, "/")
+  const workspaceParts = URI.parse(workspace)
+    .fsPath.replace(/\\/g, "/")
     .split("/")
     .filter((e) => e.length > 0);
   const entryParts = resolvedEntry.split("/").filter((e) => e.length > 0);
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -100,16 +100,24 @@ export async function quickFixCreateConfig(
 ): Promise<CodeAction | undefined> {
   const workspace = PluginConfigurationProviderInstance.getWorkspacePath();
   const entryUri = diagnostic.data.entryUri as string;
-  if (!workspace || !entryUri) {
+  if (!workspace.length || !entryUri || !entryUri.length) {
     return;
   }
-  const workspaceStrippedUri = entryUri.startsWith(workspace)
-    ? entryUri.replace(workspace, "")
-    : entryUri;
-  const programPath = URI.parse(workspaceStrippedUri).path.replace(/^\//, "");
-  if (!programPath.length) {
-    return;
-  }
+  const resolvedEntry = entryUri.startsWith("file://")
+    ? URI.parse(entryUri).fsPath.replace(/\\/g, "/")
+    : entryUri.replace(/\\/g, "/");
+
+  const normalizedWorkspace = workspace.replace(/\\/g, "/");
+  const workspacePrefix = normalizedWorkspace + "/";
+
+  const isInsideWorkspace = resolvedEntry.startsWith(workspacePrefix);
+  const rawPath = isInsideWorkspace
+    ? resolvedEntry.slice(workspacePrefix.length)
+    : resolvedEntry;
+
+  const programPath = UriUtils.processDriveLetter(rawPath).drive
+    ? rawPath[0].toUpperCase() + rawPath.slice(1)
+    : rawPath;
 
   const action: CodeAction = {
     title: `Create a plugin configuration for this file.`,
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -99,14 +99,14 @@ export async function quickFixCreateConfig(
   diagnostic: Diagnostic,
 ): Promise<CodeAction | undefined> {
   const workspace = PluginConfigurationProviderInstance.getWorkspacePath();
-  const entryUri = diagnostic.data.entryUri;
+  const entryUri = diagnostic.data.entryUri as string;
   if (!workspace || !entryUri) {
     return;
   }
-  const programPath = URI.parse(entryUri.replace(workspace, "")).path.replace(
-    /^\//,
-    "",
-  );
+  const workspaceStrippedUri = entryUri.startsWith(workspace)
+    ? entryUri.replace(workspace, "")
+    : entryUri;
+  const programPath = URI.parse(workspaceStrippedUri).path.replace(/^\//, "");
   if (!programPath.length) {
     return;
   }
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -101,12 +101,15 @@ export async function quickFixCreateConfig(
   const workspace = PluginConfigurationProviderInstance.getWorkspacePath();
   const entryUri = diagnostic.data.entryUri;
   if (!workspace || !entryUri) {
-    return undefined;
+    return;
   }
   const programPath = URI.parse(entryUri.replace(workspace, "")).path.replace(
     /^\//,
     "",
   );
+  if (!programPath.length) {
+    return;
+  }
 
   const action: CodeAction = {
     title: `Create a plugin configuration for this file.`,
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -103,18 +103,19 @@ export async function quickFixCreateConfig(
   if (!workspace || !entryUri) {
     return undefined;
   }
-  const programPath = URI.parse(
-    diagnostic.data.entryUri.replace(workspace, ""),
-  ).path.replace(/^\//, "");
+  const programPath = URI.parse(entryUri.replace(workspace, "")).path.replace(
+    /^\//,
+    "",
+  );
 
   const action: CodeAction = {
-    title: `Create a plugin configuration folder for this file.`,
+    title: `Create a plugin configuration for this file.`,
     kind: CodeActionKind.QuickFix,
     diagnostics: [diagnostic],
     command: {
       title: "Create configuration folder",
       command: Commands.CREATE_CONFIG,
-      arguments: [programPath],
+      arguments: [programPath, entryUri],
     },
   };
 
--- a/packages/language/src/language-server/commands.ts
+++ b/packages/language/src/language-server/commands.ts
@@ -15,7 +15,7 @@ import { PluginConfigurationProviderInstance } from "../workspace/plugin-configu
 import { ExecuteCommandParams } from "vscode-languageserver";
 import { UriUtils } from "../utils/uri";
 import { PluginConfiguration } from "./constants";
-import { updateExistingConfig, createNewConfig } from "../utils/config";
+import { updateOrCreateConfig } from "../utils/config";
 
 export async function commandResolveInclude(params: ExecuteCommandParams) {
   const [uri, content] = params.arguments as string[];
@@ -55,16 +55,12 @@ export async function commandCreateConfig(
     );
   }
 
-  if (progConfigFile) {
-    await updateExistingConfig(
-      workspaceFolderUri,
-      configFilePath,
-      progConfigFile,
-      programPath,
-    );
-  } else {
-    await createNewConfig(workspaceFolderUri, programPath);
-  }
+  await updateOrCreateConfig(
+    workspaceFolderUri,
+    configFilePath,
+    progConfigFile,
+    programPath,
+  );
 
   if (!returnUpdatedFile) {
     return;
--- a/packages/language/src/language-server/commands.ts
+++ b/packages/language/src/language-server/commands.ts
@@ -32,8 +32,39 @@ export async function commandCreateConfig(params: ExecuteCommandParams) {
   if (!workspaceFolderUri || !params.arguments) {
     return;
   }
+  const programPath = params.arguments[0];
+  const entryUri = URI.parse(params.arguments[1]);
+  const test = PluginConfigurationProviderInstance.getProgramConfig(entryUri);
+  console.log(test);
+  const test2 = PluginConfigurationProviderInstance.pushConfigProgram(
+    workspaceFolderUri.path,
+    programPath,
+  );
+  console.log(test2);
+  // hasProgramConfig(entryUri);
+  if (test2) {
+    const FILE = await FileSystemProviderInstance.readFile(
+      UriUtils.joinPath(
+        workspaceFolderUri,
+        PluginConfiguration.PROGRAM_FILE_PATH,
+      ),
+    );
+    if (!FILE) return;
+    const textContent = JSON.parse(FILE);
+    console.log(textContent);
+    textContent.pgms.push({
+      program: programPath,
+      pgroup: "default",
+    });
+    await FileSystemProviderInstance.writeFile(
+      UriUtils.joinPath(
+        workspaceFolderUri,
+        PluginConfiguration.PROGRAM_FILE_PATH,
+      ),
+      JSON.stringify(textContent, null, 2));
+      return;
+  }
   try {
-    const entryUri = params.arguments[0];
     await FileSystemProviderInstance.writeFile(
       UriUtils.joinPath(
         workspaceFolderUri,
@@ -45,7 +76,7 @@ export async function commandCreateConfig(params: ExecuteCommandParams) {
           pgms: [
             {
               ...PluginConfiguration.DEFAULT_PROGRAM_FILE_CONTENT.pgms[0],
-              program: entryUri,
+              program: programPath,
             },
           ],
         },
--- a/packages/language/src/language-server/commands.ts
+++ b/packages/language/src/language-server/commands.ts
@@ -44,10 +44,11 @@ export async function commandCreateConfig(
 
   let progConfigFile: string | undefined;
   try {
-   progConfigFile =
-    await FileSystemProviderInstance.readFile(configFilePath);
+    progConfigFile = await FileSystemProviderInstance.readFile(configFilePath);
   } catch {
-    console.info(`[Info] No existing "pgm_conf.json" found in workspace — a new configuration file will be created.`);
+    console.info(
+      `[Info] No existing "pgm_conf.json" found in workspace — a new configuration file will be created.`,
+    );
   }
 
   if (progConfigFile) {
--- a/packages/language/src/language-server/commands.ts
+++ b/packages/language/src/language-server/commands.ts
@@ -26,15 +26,16 @@ export async function commandResolveInclude(params: ExecuteCommandParams) {
   }
 }
 
-export async function commandCreateConfig(params: ExecuteCommandParams) {
-  const workspaceFolderUri = URI.parse(
-    PluginConfigurationProviderInstance.getWorkspacePath(),
-  );
-
-  if (!workspaceFolderUri || !params.arguments) {
+export async function commandCreateConfig(
+  params: ExecuteCommandParams,
+  returnUpdatedFile = false,
+): Promise<string | undefined> {
+  const workspacePath = PluginConfigurationProviderInstance.getWorkspacePath();
+  if (!workspacePath.length || !params.arguments) {
     return;
   }
 
+  const workspaceFolderUri = URI.parse(workspacePath);
   const programPath = params.arguments[0];
   const configFilePath = UriUtils.joinPath(
     workspaceFolderUri,
@@ -54,4 +55,15 @@ export async function commandCreateConfig(params: ExecuteCommandParams) {
   } else {
     await createNewConfig(workspaceFolderUri, programPath);
   }
+
+  if (!returnUpdatedFile) {
+    return;
+  }
+
+  const fileContent = JSON.stringify(
+    await FileSystemProviderInstance.readFile(configFilePath),
+    null,
+    2,
+  );
+  return fileContent;
 }
--- a/packages/language/src/language-server/commands.ts
+++ b/packages/language/src/language-server/commands.ts
@@ -37,6 +37,10 @@ export async function commandCreateConfig(
 
   const workspaceFolderUri = URI.parse(workspacePath);
   const programPath = params.arguments[0];
+  if (!programPath.length) {
+    return;
+  }
+
   const configFilePath = UriUtils.joinPath(
     workspaceFolderUri,
     PluginConfiguration.PROGRAM_FILE_PATH,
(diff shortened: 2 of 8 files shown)


Wagner-Laranjeiras:
TODO: Add tests for the follow up part.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -653,6 +653,30 @@ export class PluginConfigurationProvider {
...
+  /**
+   * Adds or updates a ProgramConfig in the internal map and persists
+   * the updated configuration set for the given workspace.
+   *
+   * @param workspacePath - Absolute path to the workspace.
+   * @param programConfig - The program configuration to add.
+   * @returns `true` if the configuration was pushed successfully,
+   *          `false` if no existing configurations are present.
+   */
+  public pushConfigProgram(
```
ssmifi: _Minor_: Why not just name it descriptively, e.g., `addProgramConfig`?

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -653,6 +653,30 @@ export class PluginConfigurationProvider {
...
+    workspacePath: string,
+    programConfig: ProgramConfig,
+  ): boolean {
+    if (this.programConfigs.size === 0) {
+      return false;
```
ssmifi: _Question_: Can't we just create a new programConfig if currently none exists?

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -98,18 +98,24 @@ export async function quickFixResolveInclude(
...
   diagnostic: Diagnostic,
 ): Promise<CodeAction | undefined> {
-  const entryUri = UriUtils.basename(URI.parse(diagnostic.data.entryUri));
-  if (!diagnostic.data.entryUri || !entryUri) {
+  const workspace = PluginConfigurationProviderInstance.getWorkspacePath();
+  const entryUri = diagnostic.data.entryUri;
+  if (!workspace || !entryUri) {
     return undefined;
   }
+  const programPath = URI.parse(entryUri.replace(workspace, "")).path.replace(
```
ssmifi: _Question_: The replacement might fail when workspace is included multiple times in entryUri, right? E.g., if workspace is `/lib` and entryUri is `/lib/lib/lib2/test.pli`. 
Is it possible that workspace can be a substring inside entryUri that doesn't start at the first character?

packages/language/src/language-server/commands.ts
```diff
@@ -25,46 +26,50 @@ export async function commandResolveInclude(params: ExecuteCommandParams) {
...
-  const workspaceFolderUri = URI.parse(
-    PluginConfigurationProviderInstance.getWorkspacePath(),
-  );
-  if (!workspaceFolderUri || !params.arguments) {
+export async function commandCreateConfig(
+  params: ExecuteCommandParams,
+  returnUpdatedFile = false,
+): Promise<string | undefined> {
+  const workspacePath = PluginConfigurationProviderInstance.getWorkspacePath();
+  if (!workspacePath.length || !params.arguments) {
```
ssmifi: _Minor_: Can drop the `length`. Is also correct though, but should be consistent.

packages/language/src/language-server/commands.ts
```diff
@@ -25,46 +26,50 @@ export async function commandResolveInclude(params: ExecuteCommandParams) {
...
-        2,
-      ),
+  }
+
+  if (progConfigFile) {
+    await updateExistingConfig(
+      workspaceFolderUri,
+      configFilePath,
+      progConfigFile,
+      programPath,
```
ssmifi: _Question_: There is no check on `programPath`. Should we make sure that it is a non-empty string?

packages/language/src/utils/config.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export async function updateExistingConfig(
+  workspaceFolderUri: URI,
+  configFilePath: URI,
+  progConfigFile: string,
+  programPath: string,
+) {
+  const canAddToExistingConfig =
+    PluginConfigurationProviderInstance.pushConfigProgram(
+      workspaceFolderUri.path,
```
ssmifi: _Hint for https://github.com/zowe/zowe-pli-language-support/issues/568_: It is hard to follow if the workspaceFolderUri.path which is expected by `pushConfigProgram` matches the data stored in the PluginProvider. Should be revisited later.

packages/language/src/utils/config.ts
```diff
@@ -0,0 +1,90 @@
...
+      { program: programPath, pgroup: "default" },
+    );
+
+  if (!canAddToExistingConfig) {
+    return;
+  }
+
+  const textContent = JSON.parse(progConfigFile);
```
ssmifi: If `progConfigFile` is invalid JSON code, `JSON.parse` will throw an error.

packages/language/src/utils/config.ts
```diff
@@ -0,0 +1,90 @@
...
+  textContent.pgms.push({
```
ssmifi: _Question_: Does `pushConfigProgram` not already push to the program map? And also rebuild the configs?

packages/language/test/lsp/command-create-config.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  beforeEach(async () => {
+    vfs = new VirtualFileSystemProvider();
+    pluginConfig = new PluginConfigurationProvider();
+    setFileSystemProvider(vfs);
+    setPluginConfigurationProvider(pluginConfig);
+    await pluginConfig.init(WORKSPACE_PATH);
+  });
+
+  (test("appends a new program to an existing config file", async () => {
```
ssmifi: _Minor_: Interesting, but probably unintentional parenthreses. Tests do still work though :)

packages/language/test/lsp/command-create-config.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    const result = await commandCreateConfig(
+      {
+        command: Commands.CREATE_CONFIG,
+        arguments: ["b.pli"],
+      },
+      true,
+    );
+
+    expect(result).toBeDefined();
+    expect(JSON.parse(JSON.parse(result!))).toEqual({
```
ssmifi: Ah. I first thought, 2x `JSON.parse` is wrong, but this is more of an workaround: `commandCreateConfig` wraps the `fileContent` with `JSON.stringify`. You should just return the file content.

ssmifi (commented):
Looks good already, but the replacement might fail in some edge cases.

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: I was thinking that `push` would give it more of an idea that we might be just adding to an existing config, but indeed `addProgramConfig` is cleaner/more direct.

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: Yes. I'm currently doing it in a function called `createNewConfig`. I could mix both functions, but IMHO, it's easier to read and even maybe a bit cleaner when we handle things separately. I'm of course completely ok with the idea of mixing, if you suggest doing so.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Wagner-Laranjeiras: I might be wrong about those points, but I AFAIK the `.replace` will only act on the first `/lib` match, thus returning `/lib/lib2/test.pli` (which in this case I believe is the expected behavior).

Regarding the workspace being a substring inside `entryUri`, I can't say it's impossible, but I find highly improbable, since we compute the workspace folders right from the root (so usually` C:/ `for windows and `/Users` for Unix)

packages/language/src/language-server/commands.ts
Wagner-Laranjeiras: We have it in a subtle way on the `quickFixCreateConfig` function, but now that you mention it, it indeed looks brittle or at least not explicit enough. I'll add those checks 😉

packages/language/src/utils/config.ts
Wagner-Laranjeiras: Got it! Added to my notes and gonna work on it on a follow up 😄

packages/language/src/utils/config.ts
Wagner-Laranjeiras: Made this mistake before with `.readFile` 😝 Thanks for pointing this one out!

packages/language/test/lsp/command-create-config.test.ts
Wagner-Laranjeiras: Oki doki, got it!

packages/language/src/utils/config.ts
Wagner-Laranjeiras: Jaein. I had this same question when working on it, because in my head, the code on `pushConfigProgram` should be the job done, but that doesn't actually have any reflection on the config file, even though the new program config is computed internally. I might oversaw something, AFAICT we need this.

packages/language/test/lsp/command-create-config.test.ts
Wagner-Laranjeiras: Mark mentioned it on another PR, but I'm having a hard time getting rid of it. 😅

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
ssmifi: You are right regarding the replace, but the core issue remains. My example was chosen poorly. :)
`/something/lib/lib2/test.pli` would replace the lib in the middle and not the prefix, right?

packages/language/src/workspace/plugin-configuration-provider.ts
ssmifi: If they basically serve the same purpose and only differ in the fact if there is already an entry or not, you can probably just combine them.

packages/language/src/utils/config.ts
ssmifi: Hm, can you please check this again. Afair `pushConfigProgram` attempts to also write the file, right? This looks like a duplication at least. I can give it another look tomorrow as well.

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: Got it!

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Wagner-Laranjeiras: Hmmm that's true. I'll add a check there 😉  Thanks!

Lotes: I would like to add something: You need to be sure to split your paths properly.
If you replace "/lib" and the string begins with "/library2", you will create a folder that does not exist. Do not use replace or startsWIth, endsWith. Paths are strings, but should be handled not as such. Try to split by / and join with / where you need. Always check for equality of the single parts, not just the prefix and suffix. That is dangerous.

packages/language/src/language-server/commands.ts
```diff
@@ -25,46 +26,49 @@ export async function commandResolveInclude(params: ExecuteCommandParams) {
...
     );
-  } catch (err) {
-    console.error("Failed to create configuration: ", err);
+  } else {
+    await createNewConfig(workspaceFolderUri, programPath);
   }
+
+  if (!returnUpdatedFile) {
```
Lotes: Conditional return type... why? When the caller specifies this boolean it can also call the readFile.

packages/language/test/lsp/apply-quick-fixes.test.ts
```diff
@@ -146,8 +146,22 @@ describe("quickFixCreateConfig", () => {
     const result = await applyQuickFixes.quickFixCreateConfig(diagnostic);
 
     expect(result).toBeDefined();
-    expect(result!.title).toContain("Create a plugin configuration folder");
+    expect(result!.title).toContain("Create a plugin configuration ");
```
Lotes: I think in the quick fix we were calling the config "startup configuration".
We should be careful and decide for one term. We do not want to confuse neither  the user nor us.

packages/language/src/utils/config.ts
```diff
@@ -0,0 +1,89 @@
...
+              program: programPath,
+            },
+          ],
+        },
+        null,
+        2,
+      ),
+    );
+
+    await FileSystemProviderInstance.writeFile(
```
Lotes: It might make sense to wrap the config into readConfigFile/writeConfigFile, including the JSON handling, formatting and the casting to the correct type.
Then the code here becomes a bit shorter and you can reuse those methods everywhere you are handling the config files.

packages/language/test/lsp/command-create-config.test.ts
```diff
@@ -0,0 +1,98 @@
...
+import { commandCreateConfig } from "../../src/language-server/commands";
+import { Commands } from "../../src/language-server/constants";
+
+const WORKSPACE_PATH = "/workspace";
+const CONFIG_FILE_PATH = "/workspace/.pliplugin/pgm_conf.json";
+
+const INITIAL_PROGRAM = { program: "a.pli", pgroup: "default" };
+const INITIAL_CONFIG = JSON.stringify({ pgms: [INITIAL_PROGRAM] }, null, 2);
+
+describe("commandCreateConfig", () => {
```
Lotes, line 33: Really nice test cases!

packages/language/src/language-server/commands.ts
Wagner-Laranjeiras: Thanks Markus, I didn't thought about this case, but it makes sense (and we actually do this somewhere else in the code). I'm going to apply it. 

The conditional return type is to make the testing easier, so I can check exactly the expected behavior. I can remove it though, and call the read file function on the tests. What do you think?

packages/language/test/lsp/command-create-config.test.ts
Wagner-Laranjeiras: Thanks!

packages/language/src/utils/config.ts
Wagner-Laranjeiras: Gonna work on that.

Wagner-Laranjeiras:
Still need to work on Markus review.

packages/language/src/language-server/commands.ts
Lotes: Yes, please remove this conditional. It complicates the interface. I only saw it used inside of the tests.

Wagner-Laranjeiras: Got it, thanks! 😉 Will need to create a PluginConfigurationInstance to use for the tests thought.

packages/language/src/language-server/commands.ts
```diff
@@ -25,46 +26,38 @@ export async function commandResolveInclude(params: ExecuteCommandParams) {
   }
```
ssmifi, line 23: _Minor_: The error message is hardcoded to proc_grps, regardless of what the contents of `uri` are.

packages/language/src/language-server/commands.ts
```diff
@@ -25,46 +26,38 @@ export async function commandResolveInclude(params: ExecuteCommandParams) {
...
-      ),
-      JSON.stringify(
-        PluginConfiguration.DEFAULT_PROCESS_GROUP_FILE_CONTENT,
-        null,
-        2,
-      ),
+    progConfigFile = await FileSystemProviderInstance.readFile(configFilePath);
+  } catch {
+    console.info(
+      `[Info] No existing "pgm_conf.json" found in workspace — a new configuration file will be created.`,
```
ssmifi: All errors are considered "File not found". You could check the error code for `err?.code === 'ENOENT'` to make sure the file is missing. 
Even better, you can only call `updateOrCreateConfig` if the file is missing. 
There are errors, e.g., permission denied, where you also probably dont want to try to create a new file.

packages/language/src/language-server/commands.ts
```diff
@@ -25,46 +26,38 @@ export async function commandResolveInclude(params: ExecuteCommandParams) {
...
 }
 
-export async function commandCreateConfig(params: ExecuteCommandParams) {
-  const workspaceFolderUri = URI.parse(
-    PluginConfigurationProviderInstance.getWorkspacePath(),
-  );
-  if (!workspaceFolderUri || !params.arguments) {
+export async function commandCreateConfig(
+  params: ExecuteCommandParams,
+): Promise<undefined> {
```
ssmifi: _Minor_: `Promise<undefined>` is rather unusal. `Promise<void>` is more common.

packages/language/src/language-server/commands.ts
```diff
@@ -25,46 +26,38 @@ export async function commandResolveInclude(params: ExecuteCommandParams) {
...
+    progConfigFile = await FileSystemProviderInstance.readFile(configFilePath);
+  } catch {
+    console.info(
+      `[Info] No existing "pgm_conf.json" found in workspace — a new configuration file will be created.`,
     );
-  } catch (err) {
-    console.error("Failed to create configuration: ", err);
   }
+
+  await updateOrCreateConfig(
```
ssmifi: If `updateOrCreate` if called anyway, wouldn't it be more logical to also let it handle the `progConfigFile`? Then you would have all the file operations in that one function.

packages/language/src/utils/config.ts
```diff
@@ -0,0 +1,56 @@
...
+      );
+      await PluginConfigurationProviderInstance.writeProcessGroupsFile();
+    } catch (err) {
+      console.error(err);
+      return;
+    }
+  }
+
+  if (!progConfigFile) {
+    return;
```
ssmifi, line 63: I'm a bit confused here. At this point, there should be a config, right? Either read before or created above. So, why exit here? As I said earlier, maybe it would be easier to follow, if all the file handling would happen in this function.

packages/language/src/utils/config.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { URI } from "vscode-uri";
+import { PluginConfigurationProviderInstance } from "../workspace/plugin-configuration-provider";
+
+export async function updateOrCreateConfig(
+  workspaceFolderUri: URI,
+  configFilePath: URI,
+  progConfigFile: string | undefined,
+  programPath: string,
+) {
```
ssmifi: _Minor_: Could also get a `Promise<void>`.

packages/language/src/utils/config.ts
```diff
@@ -0,0 +1,56 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { URI } from "vscode-uri";
+import { PluginConfigurationProviderInstance } from "../workspace/plugin-configuration-provider";
+
+export async function updateOrCreateConfig(
```
ssmifi: Please add documentation for this utility function.

packages/language/src/utils/config.ts
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+  }
+
+  if (!progConfigFile) {
+    return;
+  }
+  try {
+    const textContent = JSON.parse(progConfigFile);
+    textContent.pgms.push({
```
ssmifi: _Suggestion_: Just to make sure, you could evaluate the content:
>if (!textContent || typeof textContent !== 'object' || !Array.isArray(textContent.pgms)) {
  console.error('Invalid configuration file format');
  return;
}

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -357,6 +358,54 @@ export class PluginConfigurationProvider {
     console.warn("No program config found, clearing existing configurations.");
   }
 
+  public async writeProcessGroupsFile(
+    content = PluginConfiguration.DEFAULT_PROCESS_GROUP_FILE_CONTENT,
+  ) {
```
ssmifi: Please give a proper return type. The `catch` returns the error object, which is probably not intended.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -357,6 +358,54 @@ export class PluginConfigurationProvider {
...
+    ),
+  ) {
+    const workspaceUri = URI.parse(this.getWorkspacePath());
+    try {
+      await FileSystemProviderInstance.writeFile(
+        UriUtils.joinPath(workspaceUri, PluginConfiguration.PROGRAM_FILE_PATH),
+        JSON.stringify(content, null, 2),
+      );
+    } catch (err) {
+      return err;
```
ssmifi: Same as above. Please give a proper return type.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -653,6 +702,30 @@ export class PluginConfigurationProvider {
...
+   * Adds or updates a ProgramConfig in the internal map and persists
+   * the updated configuration set for the given workspace.
+   *
+   * @param workspacePath - Absolute path to the workspace.
+   * @param programConfig - The program configuration to add.
+   * @returns `true` if the configuration was pushed successfully,
+   *          `false` if no existing configurations are present.
+   */
+  public addProgramConfig(
+    workspacePath: string,
```
ssmifi: Another hint for the uri refactoring: The other functions in this file use a resolved uri for the key of map. That might be another inconsistency here. Especially, since the caller here must resolve it correctly.

packages/language/src/workspace/plugin-configuration-provider.ts
ssmifi: I still don't get why we cannot simply add a program config here, if there are currently none. If this is possible the boolean return value is probably obsolete?

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -653,6 +702,30 @@ export class PluginConfigurationProvider {
...
+    programConfig: ProgramConfig,
+  ): boolean {
+    if (this.programConfigs.size === 0) {
+      return false;
+    }
+    this.programConfigs.set(programConfig.program, programConfig);
+    this.setProgramConfigs(
+      workspacePath,
+      [...this.programConfigs.values()].map(deserializeProgramConfig),
```
ssmifi: Is `this.setProgramConfigs(workspacePath, [...this.programConfigs.values()]);` not sufficient?

ssmifi (commented):
Looks better, but the controlflow between `commands`, `config`, and `plugin-configuration-provider` is still a bit confusing. 

_Suggestion_: Test explicitly in `config.`, e.g.,
```
const hasExistingConfigs = PluginConfigurationProviderInstance.hasRegisteredProgramConfigs();

if (!hasExistingConfigs) {
  // Create new config files
  await PluginConfigurationProviderInstance.writeProgramConfigFile(programPath);
  await PluginConfigurationProviderInstance.writeProcessGroupsFile();
}

// Always add the program config (whether to existing or newly created)
PluginConfigurationProviderInstance.addProgramConfig(
  workspaceFolderUri.path,
  { program: programPath, pgroup: "default" },
);

```
and don't involve `addProgramConfig` in the controlflow logic.

Wagner-Laranjeiras:
Thanks for the review @ssmifi . Already working on it!

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -99,24 +99,38 @@ export async function quickFixCreateConfig(
...
+    : entryUri.replace(/\\/g, "/");
+
+  const workspaceParts = URI.parse(workspace)
+    .fsPath.replace(/\\/g, "/")
+    .split("/")
+    .filter((e) => e.length > 0);
+  const entryParts = resolvedEntry.split("/").filter((e) => e.length > 0);
+
+  const isInsideWorkspace = workspaceParts.every(
+    (part, index) => part === entryParts[index],
```
Lotes, line 117: Can it happen that entryParts.length < workspaceParts.length?

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -357,6 +371,72 @@ export class PluginConfigurationProvider {
...
+   * Writes the program configuration file to the workspace.
+   *
+   * Uses a default content derived from the given program path unless an
+   * override is provided. Throws if the file cannot be written.
+   *
+   * @param programPath - Path of the program, used to generate default content.
+   * @param content - Configuration content to serialize and write.
+   */
+  public async writeProgramConfigFile(
+    programPath: string,
```
Lotes: This is very confusing. `programPath` only influences the default of the next parameter. If you would set the next parameter yourself, the first would become obsolete. I would expect only the content parameter.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -99,24 +99,38 @@ export async function quickFixCreateConfig(
...
   }
-  const programPath = URI.parse(entryUri.replace(workspace, "")).path.replace(
-    /^\//,
-    "",
+  const resolvedEntry = entryUri.startsWith("file://")
+    ? URI.parse(entryUri).fsPath.replace(/\\/g, "/")
+    : entryUri.replace(/\\/g, "/");
+
+  const workspaceParts = URI.parse(workspace)
+    .fsPath.replace(/\\/g, "/")
```
Lotes, line 111: `replace(/\\/g, "/")` should be very common. Maybe we already have a helper method for that.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Lotes: I mean: is it still working for this case?

Lotes (commented):
Thanks for your work on the comments. I found some minor things.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Wagner-Laranjeiras: Yep, it would try to compare with undefined and then return false (which is what we are looking for anyway)

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Wagner-Laranjeiras: Not that I know of, but I could for sure add it. Should I do it in this PR or would be better on a next one (so we don't bloat this one even more)?

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: Gonna look into it ;)

ssmifi: As far as I can tell, `content` is never passed to the function, right? Then just make it a `const` var for now.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
ssmifi: Make a note for https://github.com/zowe/zowe-pli-language-support/issues/568 and leave it for now.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
ssmifi: Do you have a test case for this?

packages/language/src/utils/config.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  const workspaceFolderUri = URI.parse(
+    PluginConfigurationProviderInstance.getWorkspacePath(),
+  );
+  const configFilePath = UriUtils.joinPath(
+    workspaceFolderUri,
+    PluginConfiguration.PROGRAM_FILE_PATH,
+  );
+
+  const fileExists =
```
ssmifi, line 54: Just a thought, maybe for the refactoring: `fileExists` should probably not be false if there are existings configs, right? But it is an extra safe guard. Leave it for now.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -207,6 +208,19 @@ export function isLibsDir(entry: LibsEntry): entry is LibsDirEntry {
   return (entry as LibsDirEntry).dir !== undefined;
 }
 
+/**
+ * Program Configuration types
+ * ProgramEntry: Represents a single program entry in the program configuration file.
+ * ProgramConfig: Represents the structure of the program configuration file.
+ */
+export type ProgramEntry = {
```
ssmifi, line 212: _Minor_: Please make it two comments, one for each type, so that the docs are displayed correctly (e.g. on hover over).

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Wagner-Laranjeiras: Gonna add it

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -653,6 +733,35 @@ export class PluginConfigurationProvider {
     return diagnostics;
   }
 
+  /**
+   * Adds or updates a ProgramConfig in the internal map and persists
+   * the updated configuration set for the given workspace.
+   *
+   * @param workspacePath - Absolute path to the workspace.
+   * @param programConfig - The program configuration to add.
```
ssmifi: _Minor_: Docs are outdated.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -653,6 +733,35 @@ export class PluginConfigurationProvider {
...
+    workspacePath: URI,
+    programConfig: ProgramConfig,
+    programPath: string,
+    textContent: PgmsConfig,
+  ) {
+    this.programConfigs.set(programConfig.program, programConfig);
+    this.setProgramConfigs(workspacePath.path, [
+      ...this.programConfigs.values(),
+    ]);
+    textContent.pgms.push({
```
ssmifi: _Question_: This always pushes the entry. Could it be that the entry already exisits? That would create a duplication.

ssmifi (commented):
Thank you, the controlflow looks much cleaner now. I have one major question about addProgramConfig and after that I think it is good to go.

packages/language/src/utils/config.ts
Wagner-Laranjeiras: Yes, I left just to be completely sure we wouldn't run into trouble.

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: I've added a check to make sure we don't duplicate it 😉

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -653,6 +726,39 @@ export class PluginConfigurationProvider {
...
+   *
+   * @param workspacePath - Absolute path to the workspace.
+   * @param programConfig - The program configuration to register.
+   * @param programPath - Path of the program to append to the config file.
+   * @param textContent - Parsed contents of the existing program config file.
+   */
+  public async addProgramConfig(
+    workspacePath: URI,
+    programConfig: ProgramConfig,
+    programPath: string,
```
ssmifi: `programPath` is already included in `programConfig`, isn't it? It is also called as 
```
PluginConfigurationProviderInstance.addProgramConfig(
    workspaceFolderUri,
    { program: programPath, pgroup: "default" },
    programPath,
    textContent,
);
```

So, `programPath` is probably superfluous?

Wagner-Laranjeiras: Yes it is, thanks for catching that.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -207,6 +208,16 @@ export function isLibsDir(entry: LibsEntry): entry is LibsDirEntry {
   return (entry as LibsDirEntry).dir !== undefined;
 }
 
+/** ProgramEntry: Represents a single program entry in the program configuration file. */
+export type ProgramEntry = {
+  program: string;
+  pgroup: string;
+};
+/** ProgramConfig: Represents the structure of the program configuration file. */
```
ssmifi, line 216: _Minor formatting_: Please format like the other comments if using ** even it is just a one-liner. Also add an empty line.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -653,6 +726,38 @@ export class PluginConfigurationProvider {
...
+   * @param textContent - Parsed contents of the existing program config file.
+   */
+  public async addProgramConfig(
+    workspacePath: URI,
+    programConfig: ProgramConfig,
+    textContent: PgmsConfig,
+  ) {
+    const existing = this.programConfigs.get(programConfig.program);
+    if (existing && existing.pgroup === programConfig.pgroup) {
+      console.error(
```
ssmifi, line 741: _Question_: Why also test for the `pgroup` here? Sure, programs might have different groups, but the key of the map only relies on the `program`. So even if the groups differ, the entry in the map would be overwritten, right?

ssmifi (commented):
Thanks. One question left.

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: What I wanted to avoid is to return from the function because we already had another program.

Let's say we have`{ program: "a.pli", pgroup: "custom" }` and want to add `{ program: "a.pli", pgroup: "default" }`, by checking only the program group I believe we would prevent the right config to be added instead of overwriting it.

ssmifi (approved):

PR #597, opened 2026-02-11
Add missing license headers

Lotes:
I think we need to look at the search pattern. Not all files are covered yet.


scripts/check-license-headers.mts
msujew, line 1: Suggestion: Can you adjust prettier so it checks this file as well? It definitely does not follow the formatting standards.

package.json
```diff
@@ -17,11 +17,14 @@
     "pretty": "prettier -w \"**/*.ts\"",
     "playground": "pnpm --dir packages/playground build",
     "playground:clean": "pnpm --dir packages/playground clean",
-    "watch": "tsc -b tsconfig.build.json --watch"
+    "watch": "tsc -b tsconfig.build.json --watch",
+    "license:check": "tsx ./scripts/check-license-headers.mts",
```
msujew, line 21: Suggestion: We should execute this as part of our CI.

msujew (commented):
Makes sense, thank you 👍

package.json
Lotes: I added it to the `ci.yml`

.github/workflows/ci.yml
```diff
@@ -49,3 +49,7 @@ jobs:
       if: success() || failure()
       run: |
         pnpm test
+    - name: Check license headers
+      if: success() || failure()
+      run: |
+        pnpm license:check
```
msujew, line 77: Suggestion: Can we move this in front of the `build` job? To ensure that the jobs fail early.

msujew (approved):
Thanks!

msujew:
Note that this currently fails on Windows, due to the difference in EOL. The comparison should be done EOL independent.

msujew:
We can also remove this code from the testing framework:

https://github.com/zowe/zowe-pli-language-support/blob/67bc0b60237177c49628b30d7fbc638a00c82911/packages/language/test/fourslash-harness/harness-parser.ts#L280-L284

.github/workflows/ci.yml
Lotes: I suggest to run a 2nd job, after the build job. Otherwise we check licenses for 3 operating systems

msujew:
> I suggest to run a 2nd job, after the build job. Otherwise we check licenses for 3 operating systems

I'm fine with this - can we also move the lint job into this as well? It suffers from the same issue. Also the job currently fails due to missing setup steps.

Issue #596, opened 2026-02-11
Test Findings -


asatklichov:
- [x] (1)   Once we remove the label in  line 8 warning is lost 

<img width="894" height="541" alt="Image" src="https://github.com/user-attachments/assets/a46c91b0-1398-4c7c-baac-e268cb765c60" />

 ```
HELLO:
   PROCEDURE OPTIONS(MAIN); 
   PUT LIST ('HELLO, WELCOME TO PLI/1 WORLD '); 
 END HELLO;

 Namez: procedure;
 end Namez; 

```

- [x] (2) Outline view is missing on fresh opened VSCode on first opened pl/i program - record is provided already 

- [x] (3)  Duplication of compiler options with comma in the config doesn’t show error
 

<img width="1006" height="398" alt="Image" src="https://github.com/user-attachments/assets/848ca9ca-24f7-4f25-99a0-56cd84d440eb" />

 - [x] (4) Provide better error message, using real M, N values.  e.g.  `PL/I statements must start from 99 as defined in MARGINS(99,199)`

<img width="967" height="522" alt="Image" src="https://github.com/user-attachments/assets/bbde7a12-44d2-4f5f-b309-7c52e8e21171" />

- [x]  (5) LOB (verify in XML as well) variables  does not provide good error message hint, if you e.g. remove '_'. 
Error should be Expected DBCLOB_LOCATOR, but received "DBCLOBLOCATOR" instead, not token.  

<img width="861" height="347" alt="Image" src="https://github.com/user-attachments/assets/fae02a15-c671-4b50-95b3-8c17196bb72c" />

<img width="770" height="309" alt="Image" src="https://github.com/user-attachments/assets/fb1c6446-7b80-4af6-8a57-fe02d9a01d17" />

- [x] (6)  XML data host and file reference variables: should show error for locator with xml as

<img width="548" height="122" alt="Image" src="https://github.com/user-attachments/assets/4d5b2ff6-cbba-4d98-8f66-469b5ca49dba" />

- [x] (7)  Numeric Host Arrays:  Negative value should not allowed, in spec. dimension range is 1-32767

<img width="398" height="89" alt="Image" src="https://github.com/user-attachments/assets/e23814bd-7ca6-43b2-ba6a-1a52a6b57f62" />

- [x] (8)  Numeric Host Arrays:  Negative value should not allowed, in spec. dimension range is 1-32767

[...]

PR #601, opened 2026-02-12
Validations on variables

Lotes:
Closes 7 and 11 of #596


packages/language/src/typesystem/assignability.ts
```diff
@@ -0,0 +1,21 @@
+import { TypeDescriptions } from "./descriptions";
```
ssmifi, line 11: License header :)

packages/language/src/validation/type-check-validator.ts
```diff
@@ -20,8 +28,126 @@ export function typeCheck(
...
+  if (TypeDescriptions.isArithmetic(description)) {
+    const witness = description.witnesses.witnesses[AttributeKind.Precision];
+    if (description.precision && witness?.token) {
+      if (typeof description.precision.fractionalDigitsCount === "number") {
+        if (
+          description.precision.fractionalDigitsCount >
+          description.precision.totalDigitsCount
+        ) {
+          acceptor(diagnosticFromCode(PLICodes.Error.IBM2436I, witness.token));
+        } else if (description.scale == ScaleMode.Float) {
```
ssmifi: _Minor_: `===`

packages/language/src/validation/type-check-validator.ts
```diff
@@ -20,8 +28,126 @@ export function typeCheck(
...
+            typeof upperBound === "number"
+          ) {
+            if (upperBound < 1) {
+              acceptor(
+                diagnosticFromCode(PLICodes.Error.IBM1295I, witness.token),
+              );
+            }
+          } else if (
+            typeof upperBound === "number" &&
+            typeof lowerBound === "number"
```
ssmifi, line 78: _Question_: I suppose the other cases are (currently) not relevant?

packages/language/src/validation/type-check-validator.ts
```diff
@@ -20,8 +28,126 @@ export function typeCheck(
...
+          acceptor(diagnosticFromCode(PLICodes.Error.IBM2436I, witness.token));
+        } else if (description.scale == ScaleMode.Float) {
+          acceptor(diagnosticFromCode(PLICodes.Error.IBM2424I, witness.token));
+        }
+      }
+    }
+  }
+}
+
+export function validateBound(
```
ssmifi: _Minor_: Currently not used outside of this unit.

packages/language/src/validation/type-check-validator.ts
```diff
@@ -20,8 +28,126 @@ export function typeCheck(
...
+  bound: Bound,
+  acceptor: ValidationAcceptor,
+  compilationUnit: CompilationUnit,
+) {
```
ssmifi, line 112: _Suggestion_: The return type seems to be a bit complex: number | "unbounded" | "unknown" | "invalid" | "computed" | undefined? Maybe we should make it explicit.

packages/language/src/validation/type-check-validator.ts
```diff
@@ -20,8 +28,126 @@ export function typeCheck(
...
+      }
+      if (!TypeDescriptions.isArithmetic(boundDescription)) {
+        //TODO handle other types
+        //CHAR=612
+        acceptor(
+          diagnosticFromCode(
+            PLICodes.Severe.IBM1948I,
+            bound.token,
+            "CONVERSION",
+            "612",
```
ssmifi: _Suggestion_: If we expect more codes, we should probably use an enum, especially since it is used at different locations now.

ssmifi (approved):
Nice, thanks. I just have some minor suggestions.

packages/language/src/validation/type-check-validator.ts
Lotes: I added some cases that are not relevant now :-(. I think there will be a lot to do still, but not in this PR.

packages/language/src/validation/type-check-validator.ts
Lotes: Currently I do not use these return types. I added them more as documentation.

packages/language/src/validation/type-check-validator.ts
Lotes: I think the parser misses some literals like for Picture type and Widechar sub type. This means we can only detect strings for now.

Lotes: Added a case for WIDECHAR. PICTURE literals are not available yet.

PR #603, opened 2026-02-12
Implement validator for SQL as XML types with locators

Lotes:
This closes point 6 of #596

### DIFF
--- a/packages/language/src/parser/sql-parser.ts
+++ b/packages/language/src/parser/sql-parser.ts
@@ -44,7 +44,7 @@ export function sqlAttributeStatement(
     t.TYPE,
   );
   state.consume(attributeStatement, CstNodeKind.SqlAttributeStatement_IS, t.IS);
-  let afterXmlAs: t.Token | null = null;
+  let afterXmlAsToken: t.Token | null = null;
   if (state.canConsume(t.XML)) {
     state.consume(
       attributeStatement,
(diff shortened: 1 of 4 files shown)


packages/language/src/validation/pp-validator.ts
```diff
@@ -20,5 +28,24 @@ export function registerPreprocessorValidationChecks(): ValidationChecks {
...
+  );
+}
+
+function XML_Locator_Check(
+  node: SqlAttributeStatement,
+  acceptor: ValidationAcceptor,
+  compilationUnit: CompilationUnit,
+): void {
+  if (node.isXml && node.body && isLocator(node.body) && node.xmlToken) {
+    acceptor(diagnosticFromCode(PLICodes.Severe.IBM3783I, node.xmlToken));
```
msujew: Suggestion: Since this is the same error that the parser is showing anyway, shouldn't we just handle this completely in the parser? I.e. just exclude certain parser branches by using `!node.isXml`. IMO, that would be cleaner, even if it shows the error on the token after the `AS` (which is probably the intended position).

Lotes: Ah ok now I understand... did it wrong xD

msujew (approved):
That's perfect, thank you!

packages/language/src/parser/sql-parser.ts
msujew: I believe `XML AS` is also invalid on `ROWID`, see the documentation [here](https://www.ibm.com/docs/en/db2-for-zos/12.0.0?topic=pli-host-variable-arrays-in#db2z_hostvariablearraypli__title__13). Or to say it better, it's only valid on `LOB` and `LOB_FILE` like tokens.

msujew: I believe this branch should also be disabled.

PR #604, opened 2026-02-15
[Bugfix] Windows absolute path programs match

Wagner-Laranjeiras:
This is a follow up on PR #569.

Absolute Windows paths added as program entry point were not being properly prefixed (missing `file:///`) and encoded (`:` after drive letter), which caused the match from such programs to fail.

This change standardize the path processing, so the matches should happen across different OSs without any complications.


ssmifi (approved):
LGTM

PR #605, opened 2026-02-16
Fix regression inside of `SQL TYPE IS XML AS ...`

Lotes:
In this approach I am just disabling the branches that are not relevant.
The previous approach tried to reduce redundancy, but that was probably too much.


ssmifi (approved):
LGTM

PR #607, opened 2026-02-17
Disable SQL syntax highlighting

Lotes:
This fix disables SQL as keyword for the Textmate file. Then the EXEC rule in the `pli.manual.json` can be applied.

The result is that the entire SQL statement is red now. I know that the `DECLARE` line was requested as highlighted, but I cannot imagine a quick solution here at the moment.

image

### DIFF
--- a/scripts/generate-tmlanguage.mts
+++ b/scripts/generate-tmlanguage.mts
@@ -27,10 +27,15 @@ const controlKeywords: string[] = [];
 const storageKeywords: string[] = [];
 
 for (const [text, type] of tokens.keywordMap.entries()) {
+  const lowerText = text.toLowerCase();
+  if (lowerText === "sql") {
+    //HACK: we exclude SQL as a keyword, because then its highlighting will be handled the same as CICS
+    continue;
+  }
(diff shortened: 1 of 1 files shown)


Lotes:
A cleaner solution could be another enum literal IGNORE and then filter SQL out.
Hopefully someone else has a better idea...

scripts/generate-tmlanguage.mts
```diff
@@ -27,10 +27,15 @@ const controlKeywords: string[] = [];
 const storageKeywords: string[] = [];
 
 for (const [text, type] of tokens.keywordMap.entries()) {
+  const lowerText = text.toLowerCase();
+  if (lowerText === "sql") {
+    //HACK: we exclude SQL as a keyword, because then its highlighting will be handled the same as CICS
```
ssmifi: _Question_: Can we make it a bit less hacky by using the name of the `SQL`  `TokenType`? And then maybe drop the `HACK:` from the comment? :)

ssmifi (approved):
Thanks!

PR #609, opened 2026-02-17
[Improvement] Absolute paths on Info Dialogue

Wagner-Laranjeiras:
**Minor improvement** - When creating program entries from absolute paths outside the workspace via info dialogue, the fullpath is displayed on the program field.

So instead of `"../../../../../../../pli-ext-programs/remote-pgm.pli"`
we would display `"/Users/pli-ext-programs/remote-pgm.pli"`

This **does not** affect the performance of the extension. The current implementation works as expected and the program is processed the same way in both scenarios.

A suggested follow up task would be wrapping this modification into a helper and possibly replace the `.relative`, since it's used more than once throughout the codebase. 

There's no urgent character on this, nor on the potential follow up task.

### DIFF
--- a/packages/vscode-extension/src/common/missing-config-handler.ts
+++ b/packages/vscode-extension/src/common/missing-config-handler.ts
@@ -12,7 +12,7 @@
 import * as vscode from "vscode";
 import * as path from "node:path";
 import * as fs from "node:fs";
-import { PluginConfiguration } from "pli-language";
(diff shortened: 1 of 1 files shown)


asatklichov:
@Wagner-Laranjeiras we have a code freeze now

msujew (approved):
Thanks, looks good to me 👍

Issue #608, opened 2026-02-17
Verify defunc validator code


ssmifi:
This legacy validator code is defunc since the refactoring. Please verify which validations are already replaced by newer versions and add the missing ones. 

```ts
// TODO: Refactor validation methods of this class into their own files and add tests
/**
 * Implementation of custom validations.
 */
export class PliValidator {
  /**
   * Verify programs contain at least one parsed statement
   */
  checkPliProgram(node: AST.Program, acceptor: ValidationAcceptor): void {
    if (node.statements.length === 0) {
      // TODO: Reimplement this validation and add tests
      // acceptor(Severity.S, PLICodes.Severe.IBM1917I.message, {
      //   code: PLICodes.Severe.IBM1917I.fullCode,
      //   range: getSyntaxNodeRange(node)!,
      //   uri: "", // TODO @montymxb Still need to supply URI for this document we're working in
      // });
    }
  }

  /**
   * Checks return options for mutually exclusive attributes
   */
  checkReturnsOption(
    node: AST.ReturnsOption,
    acceptor: ValidationAcceptor,
  ): void {
    const attrSet = new Set<string>();
    for (const attr of node.returnAttributes) {
      if (
        attr.kind === AST.SyntaxKind.ComputationDataAttribute &&
        attr.type !== null
      ) {
        const typ = tokens.DefaultAttribute.mapFromEnumLiteral(attr.type);
        attrSet.add(typ); // dupes are ok

        // look for a generally negated version of this attribute (there are several)
        if (attrSet.has(`UN${typ}`)) {
          // TODO: Reimplement this validation and add tests
          // acceptor(
          //   Severity.E,
          //   PLICodes.Error.IBM2462I.message(typ, `UN${typ}`),
          //   {
          //     code: PLICodes.Error.IBM2462I.fullCode,
          //     range: getSyntaxNodeRange(node)!,
          //     uri: "", // TODO @montymxb Still need to supply URI for this document we're working in
          //     // property: "returnAttributes",
          //     // node
          //   },
          // );
        }

        // look for a non-negated version of this type
        if (typ.startsWith("UN") && attrSet.has(typ.slice(2))) {
          // TODO: Reimplement this validation and add tests
          // acceptor(
          //   Severity.E,
          //   PLICodes.Error.IBM2462I.message(typ, typ.slice(2)),
          //   {
          //     code: PLICodes.Error.IBM2462I.fullCode,
          //     range: getSyntaxNodeRange(node)!,
[...]

PR #610, opened 2026-02-17
Add additional LS tests

ssmifi:
- Added tests for the margin and skipped code indicators. 
- Removed unused legacy validator code. https://github.com/zowe/zowe-pli-language-support/issues/608 tracks if we restored all old validations.


Wagner-Laranjeiras:
LGTM

Wagner-Laranjeiras (approved):

PR #613, opened 2026-02-18
Hide symbols from outline when not in open file

Lotes:
Included symbols were also shown. With this fix we suppress included declarations.

Be aware of that we are scanning all file tokens, so the preprocessor included stuff is not part of the tokens. The DeclareSymbolBuilder retrieves all declarations from the root declaration (via the AST). That is were we compare the nameToken URI with the document URI.

### DIFF
--- a/packages/language/src/language-server/document-symbol-builder.ts
+++ b/packages/language/src/language-server/document-symbol-builder.ts
@@ -116,7 +116,9 @@ class DeclareSymbolBuilder implements SymbolBuilder {
     for (const item of declareStatement.items.filter(
       (i: DeclaredItem) => i.kind === SyntaxKind.DeclaredItem,
     )) {
-      levelSymbols.push(...this.retrieveLevelSymbols(token.uri!, item, null, childSymbols));
+      levelSymbols.push(
+        ...this.retrieveLevelSymbols(token.uri!, item, null, childSymbols),
+      );
     }
 
     for (const levelSymbol of levelSymbols) {
@@ -172,7 +174,7 @@ class DeclareSymbolBuilder implements SymbolBuilder {
           continue;
         }
 
-        if(element.nameToken.uri?.toString() !== documentUri.toString()) {
+        if (element.nameToken.uri?.toString() !== documentUri.toString()) {
           continue;
         }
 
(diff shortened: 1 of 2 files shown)


packages/language/src/language-server/document-symbol-builder.ts
```diff
@@ -115,7 +116,9 @@ class DeclareSymbolBuilder implements SymbolBuilder {
     for (const item of declareStatement.items.filter(
       (i: DeclaredItem) => i.kind === SyntaxKind.DeclaredItem,
     )) {
-      levelSymbols.push(...this.retrieveLevelSymbols(item, null, childSymbols));
+      levelSymbols.push(
+        ...this.retrieveLevelSymbols(token.uri!, item, null, childSymbols),
```
ssmifi: _Question_: I am not sure under which circumstances the token uri can be undefined, but if it can be undefined this could be an issue for `retrieveLevelSymbols`. Are you sure the `!` is safe here?

packages/language/test/fourslash/linker/include-split-structure.ts
```diff
@@ -0,0 +1,33 @@
...
+
+preprocessor.expectTokens(`
+  DCL 1 person ALIGNED,
+        3 cust CHAR(30), 
+        3 addr,
+          5 street CHAR(20), 
+          5 zip CHAR(5); 
+  PUT LIST (person.addr.street);
+`);
+linker.expectLinks();
```
ssmifi, line 33: _Question_: Thanks for the test. But this does not actually tell us anything about the document symbols, right? But the reference is working, yes.

ssmifi (commented):
Thank you. LG. Just two small questions.

packages/language/src/language-server/document-symbol-builder.ts
Lotes: No, it is not safe. I fixed it, thanks.

I tested it only for one document, it was always the URI of that document. But I will not bet on it.

packages/language/test/fourslash/linker/include-split-structure.ts
Lotes: There are tests for document symbols, but no multi file tests. I can put some time on it to find out how to make it happen. Not sure if I can find a solution for the remaining task.

ssmifi: Thanks. Let's postpone this for now.

ssmifi (approved):

PR #615, opened 2026-02-19
[Bugfix] Fix overwritten process configs.

Wagner-Laranjeiras:
### Changes
- Reads and parses `proc_grps.json` from the workspace
- Updates the `libs` of the matching process group only, preserving all others
- Adds tests covering the happy path and multi-group scenarios


ssmifi (approved):
LGTM

PR #635, opened 2026-03-16
`inferType` for the preprocessor

### LINKED ISSUE #591: Type system: `inferType` for Preprocessor
Needed to handle types in the preprocessor.

The preprocessor type system is a subset of the PL/I type system. So, I think we can reuse big parts.

Lotes:
Closes https://github.com/zowe/zowe-pli-language-support/issues/591.

According to [this page](https://www.ibm.com/docs/en/epfz/6.1.0?topic=statements-declare-statement) this PR adds following features for the preprocessor:

* DIMENSION
* types: CHARACTER and FIXED without length or precision (length & precision seem to be forbidden)
* scopes: EXTERNAL/INTERNAL are allowed
* scan modes: RESCAN, NOSCAN, SCAN
* special attributes: ENTRY, BUILTIN => done


packages/language/src/preprocessor/instruction-cache.ts
```diff
@@ -29,6 +29,7 @@ export interface FileInstructionResult {
   diagnostics: Diagnostic[];
   statements: Statement[];
   result: InstructionGeneratorResult;
+  preprocessorTokens: Token[];
```
msujew: Question: What is the difference between `preprocessorTokens` and `tokens`? I believe these should be the same.

packages/language/src/typesystem/infer.ts
```diff
@@ -253,6 +253,7 @@ export class DefaultTypeInferer implements TypeInferer {
     const attributes = builder.collectAttributes(
       node.returnsToken,
       node.returnAttributes,
+      ast.getContainer(node, ast.SyntaxKind.Program) === compilationUnit.preprocessorAst
```
msujew: Suggestion: You can use `ast.isPreprocessorNode` for this instead (it does the same thing, but slightly more efficient).

packages/language/src/preprocessor/pli-lexer.ts
```diff
@@ -37,6 +37,7 @@ export interface LexerResult {
   statements: Statement[];
   evaluationResults: EvaluationResults;
   tokenReferences: Reference[];
+  preprocessorTokens: Token[];
```
msujew: Suggestion: These seem to unused.

packages/language/src/workspace/compilation-unit.ts
```diff
@@ -72,6 +72,7 @@ export interface CompilationUnit {
   uri: URI;
   compilerOptions: CompilerOptions;
   ast: Program;
+  preprocessorTokens: Token[];
```
msujew: Same here.

packages/language/src/typesystem/descriptions.ts
```diff
@@ -1506,10 +1548,13 @@ interface UnknownTypeDescription extends BaseTypeDescription {
   type: UnknownType;
 }
 
-function createUnknownTypeDescription(): UnknownTypeDescription {
+function createUnknownTypeDescription(common?: {
+  builtIn?: boolean;
```
Lotes, line 1553: This single field feels weird from a design perspective.
Unknown is like "any" and will hide type comparisons. That was the idea.

packages/language/test/test-builder.ts
```diff
@@ -1067,19 +1067,38 @@ export class TestBuilder {
...
+    }
+    return ranges;
+  }
+
+  private expectTypeFromTokens(
+    tokens: Token[],
+    label: string,
+    expectedType: TypeExpectation,
+    fileUri: string,
+  ): void {
```
msujew: Suggestion: This seems like a workaround for me. Right now, we cannot reference labels that do not live in the main document. This has mostly worked well so far, but looking at a few tests, this only works "by accident". IMO, we should attempt to improve this by taking all labels into account. Then we also no longer need to provide the additional file argument here.

msujew (commented):
Works well, but I think we can improve on the test-builder implementation changes.

packages/language/test/test-builder.ts
Lotes: I agree, it feels very hacky. I have not thought about a refactoring of the test API.

I also need the right token set here: depending on the label I have to reference different tokens for the binary search. How shall we distinguish between:

* I need the tokens that are going into the preprocessor parser, from
* I need the tokens that going into the PL/I parser

For include files there is only one type possible. The main file may have the same labels in both scenarios.
Now, when thinking loud, does it even makes sense to observe the second type of tokens? The second type makes no sense here, since we cannot label symbols that are only available after the preprocessor is done.

Lotes: Each label/index should be associated with an URI of the corresponding document.

packages/language/test/test-builder.ts
Lotes: These loops "this.files.values()" do not look right

packages/language/test/test-builder.ts
```diff
@@ -85,10 +87,16 @@ export type PliTestFile = {
   content: string;
 };
 
+/** uri, index */
+export type TestIndex = [string, number];
+
+/** uri, start, end */
+export type TestRange = [string, number, number];
```
msujew: Suggestion: Since we're already adding types for these, can we make these into objects? Much easier to read and reason about.

packages/language/test/utils.ts
```diff
@@ -311,6 +311,9 @@ interface ExpectedBase {
   rangeEndMarker?: string;
 }
 
+export type TestUtilsIndex = number;
+export type TestUtilsRange = [number, number];
```
msujew: Same here.

packages/language/src/typesystem/descriptions.ts
Lotes: I think the ideal solution would resolve the linked builtin variable or procedure and assigns its type to itself. Currently this seems to be fine. Will add a TODO.

msujew (approved):
Thank you, I think this looks good 👍

PR #637, opened 2026-03-16
Allow up to 7 character for NOT and OR options

### LINKED ISSUE #624: Multiple NOT symbols not recognized in compiler options
The PL/I Programming Guide says

> The NOT option specifies up to seven alternate symbols that can be used as the logical NOT operator.

And it looks like our implementation only accepts one.

Image

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/624

We previously assumed that only a single character was allowed for the `NOT` and `OR` options. However, the programming guide indicates that up to 7 characters are fine for both of them. This simply updates our compiler options parsing code to accept up 7 characters.


ssmifi (approved):
LGTM, wb :)

PR #638, opened 2026-03-16
Fix missing parser errors on EOF

msujew:
I noticed during manual testing that we seem to swallow up errors in case we trigger a rule call on EOF. Instead, we should've entered the rule and run into a parser error.

This change simply adjusts the behavior and adds a test for this. It also removes a few old tests that were no longer relevant.


Lotes (approved):
LGTM

PR #639, opened 2026-03-16
Fix false positive on 1213I diagnostic

### LINKED ISSUE #621: New false positive in 1.1.0
When I installed 1.1.0 to try it out it instantly showed me a new diagnostic which was not present in previous versions of the extension.

Image

When I compile the program with the PL/I compiler defaults it does not report any issues.

Attaching a compile listing [listing.txt](https://github.com/user-attachments/files/25690474/listing.txt).

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/621

The `IBM1213I` warning is only supposed to trigger when the procedure is not externally visible (meaning that it is hidden inside of another procedure).

This change fixes the false positive experienced in the issue and adjusts the feature to work like the compiler does. Also moves the validation code from the linking to the validation phase.

### DIFF
--- a/packages/language/src/linking/error.ts
+++ b/packages/language/src/linking/error.ts
@@ -125,15 +125,6 @@ export class LinkerErrorReporter {
   }
 
   /**
-   * W IBM1213I
-   */
-  reportUnreferencedSymbol(token: Token) {
-    this.accept(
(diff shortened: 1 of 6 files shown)


Lotes (approved):
Good replacement.

PR #642, opened 2026-03-17
Support comments in semantic tokens request

### LINKED ISSUE #626: Comment continued into margin breaks parser
When a comment is terminated after end of MARGIN, the reported diagnostic is not very helpful. The highlighting also does not reflect the fact that the comment is effectively unterminated. This makes it hard to understand what is happening in a large program if a user accidentally crosses the margin boundary.

Image

It would be helpful if the textmate grammar reflected the MARGIN end or if semantic tokens colored the code as comment like it does when there is not text after MARGIN end.

Image

Showing a diagnostic that there is text after MARGIN end on a given line would also be helpful to find the root cause of the issue.

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/626

Adds the parsed comments to our parser infrastructure and stores them for each file. Will be used by the semantic token request to correctly highlight all comments in a file.

### DIFF
--- a/packages/language/src/language-server/semantic-tokens.ts
+++ b/packages/language/src/language-server/semantic-tokens.ts
@@ -42,6 +42,7 @@ export enum SemanticTokenTypes {
   class,
   type,
   string,
+  comment,
 }
 
 const tokenModifiers = new Map<string, number>([]);
@@ -59,11 +60,21 @@ export function semanticTokens(
   range?: Range,
 ): number[] {
   const tokens = compilationUnit.services.files.getTokens(textDocument.uri);
-  if (!tokens) {
+  const comments = compilationUnit.services.files.getComments(textDocument.uri);
+  if (!tokens || !comments) {
     return [];
   }
+  let commentIndex = 0;
   const semanticTokens = new SemanticTokensBuilder();
   for (const token of tokens) {
+    // Process any comments that appear before the current token
+    while (
+      comments.length > commentIndex &&
+      comments[commentIndex].startOffset < token.startOffset
+    ) {
+      const comment = comments[commentIndex++];
+      handleCommentTokens(textDocument, semanticTokens, comment);
+    }
     const type = tokenType(token);
     if (type !== undefined) {
       semanticTokens.push(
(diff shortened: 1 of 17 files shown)


packages/language/src/language-server/semantic-tokens.ts
```diff
@@ -59,11 +60,21 @@ export function semanticTokens(
...
+  const comments = compilationUnit.services.files.getComments(textDocument.uri);
+  if (!tokens || !comments) {
     return [];
   }
+  let commentIndex = 0;
   const semanticTokens = new SemanticTokensBuilder();
   for (const token of tokens) {
+    // Process any comments that appear before the current token
+    while (
+      comments.length > commentIndex &&
```
Lotes: Why not `commentIndex < comments.length`? Reads weird haha ^^
You broke a read/write habit.

packages/language/test/fourslash/semantic-tokens/comments.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/// <reference path="../framework.ts" />
+
+// @wrap: main
+//// PUT("HELLO WORLD"); <|comment:// single line comment|>
+//// <|comment:/* multi line comment start|>
+////<|comment: multi line comment content|>
+////<|comment: multi line comment end */|> PUT("GOODBYE WORLD");
+
+semanticTokens.expectAt("comment");
```
Lotes, line 23: How about a test that hits the right margin?

packages/language/src/language-server/semantic-tokens.ts
```diff
@@ -75,9 +86,52 @@ export function semanticTokens(
       );
     }
   }
+  // Process any remaining comments after the last token
+  while (comments.length > commentIndex) {
```
Lotes: A test coverage thing: Do we have tests that hit this loop?

packages/language/src/preprocessor/compiler-options/parser.ts
```diff
@@ -276,10 +275,16 @@ export function parseAbstractCompilerOptions(
   input: string,
   uri?: URI,
   offset?: number,
+  line?: number,
+  column?: number,
 ): AbstractCompilerOptions {
+  offset ??= 0;
+  column ??= 0;
```
Lotes, line 284: Docs missing for these new parameters

packages/language/src/preprocessor/compiler-options/parser.ts
```diff
@@ -290,9 +295,21 @@ export function parseAbstractCompilerOptions(
   if (uri) {
     for (const token of tokens) {
       token.uri = uri;
-      token.startLine -= 1;
-      token.startColumn -= 1;
-      token.endLine -= 1;
+      token.startLine += line - 1;
```
Lotes, line 300: Why do we move the tokens and comments?

Lotes (commented):
Nice changes, thanks. There is one section I do not understand. Why do we move then tokens to a different position?

packages/language/src/preprocessor/compiler-options/parser.ts
msujew: Because the input is just "PP" even though it's actually "*PROCESS PP" on line three or something like that. We only receive a substring of the original text input here, which is why we need to adjust the token positions to their original place.

Lotes (approved):

PR #644, opened 2026-03-17
Improve test utils performance

msujew:
Fixes a performance issue in our test suite. This has resulted in 5s+ wait times for a list of ~160 test cases. This reduces it down to <1s.


ssmifi (approved):
Looks good. Thanks!

PR #645, opened 2026-03-18
Add CICS syntax highlighting

Lotes:
I created a script using Claude Code that scans all direct sub-pages of the CICS commands documentation.
It extracts the syntax diagram SVGs and looks for TEXT elements that are keywords.
It pipes everything into a sorted list of keywords and adds a markdown that contains all scanned keywords.
After scraping the data I needed to clean up the list a bit, since the SVG generator is not absolute consistent with parameters like single numbers or lower cased identifiers that normally are just place holders. These manual changes are in a range of 20-30 lines of removed keywords. They still exist in the markdown. Like "en" or "options" (lower-cased).

I hope I added them correctly to the `.tmLanguage` file.

Screenshot:
image


Lotes:
One open question now is: Shall be the first keyword of each command be a CONTROL token? This word trigger a separate coloring for the keyword like in the following screenshot (look at `ABEND` in comparison to the first screenshot):
image

Lotes:
I have found an easy way to get the CONTROL tokens: They are the starting words of the sub page titles until a parenthesis or and end $ comes. But I will check the titles again, sometimes there is a capital word the does not belong to the command...

msujew (commented):
Wouldn't it make more sense to generate something similar to the `tokens.ts` file? That way we get programmatic access to the tokens.

Lotes:
> Wouldn't it make more sense to generate something similar to the `tokens.ts` file? That way we get programmatic access to the tokens.

Yeah, that is true. I avoided it, because it is producing a lot of text that you cannot cross-check so easily against errors (I would say).
Currently, we do not need programmatic access. It would be good to have that in the future when implementing the actual parser.

I can make this change.

msujew:
> Currently, we do not need programmatic access.

We do soon, otherwise we cannot produce the semantic tokens for the CICS keywords.

packages/vscode-extension/syntaxes/cics-keywords.json
msujew, line 1: What do we need this for? I don't think it's used anywhere, right?

msujew (approved):
Looks good, just one question, see below. Also, fix the license headers.

packages/vscode-extension/syntaxes/cics-keywords.json
Lotes: I think I will make this more clear in a next step:
* `fetch-cics-keywords.mts` gathers all keywords from the online documentation
* the result has to be looked at manually (but I can also add a blacklist)
* then `generate-cics-keyword-tokens.ts` is called on this file togenerate the tokens

Lotes: Deleted.

Lotes:
Put back to draft, we need to adapt the tokenizer first :).

Lotes:
[Here](https://zowe.github.io/zowe-pli-language-support/pr/pr-645/?content=AQQQagogSiDiEC5gAUoHkDCEAiBVKEwayAKgJJoByAysABQCyIZlAlANzABQwvApABMApgGMANgEMATkOAkhADwAuwEQAtp7Hr2S4SwADJlq%2BugHJUaWDAZy0wDGga6ShcNDgQzHbcAgANCAwHMgxaEAAhCEpsUAjHbAhzSIAGAEZvBxBKLAMtXj9A4IxQ2hMQKH0SGBoybHNqygNMllcoMBADOjSU3pTWYAIARTrzSipYZp8CgKDgaiGDeYgDIP0AKmAAMXRbf338nT1DY1MLdGsQWxJ7R2c9N0gYeG986Nj3Z4hOIA) a link with a test scenario. Weird thing: In playground the semantic tokens for CICS work, but not for SQL and on VS Code it is vice versa.

Or copy paste this example:
```
 AVERAGE: PROCEDURE OPTIONS (MAIN); 
   %declare Text char;
   PUT LIST ('PROGRAM TO COMPUTE AVERAGE');
   EXEC CICS ABEND ABCODE('AB01') CANCEL;
   EXEC CICS START TRANSID('TRNL') INTERVAL(10000) REQID('NONGL');
   EXEC SQL SELECT * FROM XXX;
   PUT LIST ('PROGRAM TO COMPUTE AVERAGE');
 END AVERAGE; 
```

NOTE: I have an approval, but since there were some changes, I would prefer another review.

packages/language/src/language-server/semantic-tokens.ts
```diff
@@ -54,6 +56,32 @@ export const semanticTokenLegend: SemanticTokensLegend = {
   tokenModifiers: Array.from(tokenModifiers.keys()),
 };
 
+function tokenModeReconstructor() {
```
msujew: Question: We remove this once we have better SQL/CICS language support, right? If so, please add a todo here.

msujew (approved):
I believe we can take a look at the weird highlighting issue in a separate PR. I believe I found out what the problem is.

packages/language/src/language-server/semantic-tokens.ts
Lotes: Yes, I will do. This is just a temporary solution. I tried to keep it compact.

Lotes: And... it is gone...

packages/language/src/parser/cics-parser.ts
```diff
@@ -13,14 +13,21 @@ import * as ast from "../syntax-tree/ast";
...
-  execStatement.content = embeddedUnknownStatement(state);
+  while (!state.eof && !state.canConsume(t.Semicolon)) {
+    const token = state.token;
+    if (!token) {
+      break;
+    }
+    token.element = execStatement;
+    token.kind = CstNodeKind.ExecCicsStatement_COMMAND;
+    state.index++;
+  }
```
msujew: Suggestion: Can we use `embeddedUnknownStatement(state, CstNodeKind.ExecCicsStatement_COMMAND)` or similar instead?

msujew (approved):
Alright. Looks good to me. Thank you!

PR #648, opened 2026-03-19
Support multi-mode lexing for preprocessors

msujew:
* Moves the `funcs` array from a global to the tokenizer context to support multi-mode lexing.
* Adjusts the lexing of `EXEC` statement to produce tokens based on the respective mode.
* Adjusts the `EXEC` statement to use the generated tokens of the current mode (this is still needs to be extended).
* Adds semantic highlighting for exec statements to the semantic tokens.


packages/language/src/preprocessor/pli-lexer.ts
Lotes, line 1: It might be weird to have a pli-tokenizer and a pli-lexer, because both means the same. Only the location gives a hint.

packages/language/src/parser/cics-parser.ts
Lotes, line 1: Organization-wise, what would you prefer? A folder per tokenizer mode (PLI/, CICS/, SQL/) or prefixes per mode (pli-, sql-, cics-)?
I see that you created a folder for `tokenizer/`, I created one for `tokens/`. We should be careful.

packages/language/src/parser/tokenizer.ts
```diff
@@ -626,4 +93,4 @@ export function tokenize(
   };
 }
 
-initLexer(getDefaultCompilerOptions());
+updatePliTokenizer(getDefaultCompilerOptions());
```
Lotes, line 96: Where would we put the updateCics/SqlTokenizer?

packages/language/src/parser/cics-parser.ts
msujew: I think having `tokenizer/` and `tokens/` is good 👍

packages/language/src/preprocessor/pli-lexer.ts
msujew: The `pli-lexer` class needs to be refactored anyway (it shouldn't be a class in the first place) - we can do that separately.

packages/language/src/parser/tokenizer.ts
msujew: We should organize it similar to the `updatePliTokenizer`, if necessary.

Lotes: Ok, let's move on. We will see.

Lotes (approved):

PR #651, opened 2026-03-19
URI handling refactoring

### LINKED ISSUE #568: URI handling refactor

Wagner-Laranjeiras:
### Refactor URI handling with unified normalization API
Closes #568

**Summary**
Introduce three core functions in UriUtils (toUri, toNormalizedKey, toFilePath) that replace all ad-hoc URI construction, comparison, and path extraction across the codebase
Fix the VFS read/write key inconsistency by standardizing on toNormalizedKey for all map operations
Remove six workaround functions that existed only to patch platform-specific edge cases (processDriveLetter, handleDriveLetter, stringPath, normalizeForWindowsTests, reconstructFileUri, stripSchemaFromURIString)
Add a dedicated test suite with 32 tests covering cross-platform paths, fragment handling, encoding, and drive letter normalization

**Problem**
The codebase had five root causes driving URI-related bugs:

- Conflating URI.parse() and URI.file() — URI.parse expects a URI string (e.g. file:///path), URI.file expects a filesystem path (e.g. /path). Misusing one for the other produced scheme-less URIs, double-encoded paths, or missing file:// prefixes. Several test files and production call sites used URI.parse on bare paths.

- VFS read/write inconsistency — writeFile stored keys via uri.toString(true).toLowerCase(), but readFile used a separate reconstructFileUri() that manually concatenated file:// + path + fragment. These two functions could produce different strings for the same logical file.

- `#` in PL/I filenames treated as URI fragments — Filenames like A1@#_$ were split by URI.parse at the #, putting _$ into the fragment and truncating the path. The reconstructFileUri method existed solely to patch this on the read side.

- Drive letter casing and leading-slash variance — vscode-uri lowercases drive letters (C: → c:), but OS paths use uppercase. This spawned processDriveLetter, handleDriveLetter, normalizeForWindowsTests, and inline path[0].toUpperCase() patches.

- No canonical comparison form — Different call sites used .toString(), .toString(true), .path, .fsPath.replace(...), or manual file:// stripping. There was no single agreed-upon string form for equality checks or map lookups.

**Solution**
Built a consistent abstraction on top of vscode-uri (not replacing it) with three functions, each with a single responsibility:

**`UriUtils.toUri(input: string | URI): URI`**
Smart constructor that detects whether input is a URI string or a file path and calls the correct vscode-uri factory:

URI objects → returned as-is
Strings with a scheme (file://, memory://) → URI.parse
Windows drive paths (C:\, C:/) → backslash-normalized, then URI.file
Bare paths (/workspace/main.pli) → URI.file (correctly encodes #, spaces, etc.)
When to use: any time you need a URI object from a string whose format you can't guarantee. Replaces all mixed URI.parse/URI.file call sites.

**`UriUtils.toNormalizedKey(input: URI | string): string`**
Produces a canonical, case-insensitive string key for map lookups and equality checks. Built from decoded path components. Merges URI fragments back into the path to handle PL/I filenames containing #. Always includes the scheme. Always lowercase.

When to use: any time you store a URI in a Map/Set, or compare two URIs for logical equality.

**`UriUtils.toFilePath(input: URI | string): string`**
Returns a normalized, decoded file path string with no scheme. Fragments are re-appended. Backslashes are normalized to forward slashes. Windows drive letters are uppercased and the leading URI slash is stripped (/c:/path → C:/path).

When to use: any time you need a human-readable path for display, config values, or OS operations.


packages/language/src/workspace/file-system-provider.ts
```diff
@@ -267,51 +239,54 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
...
+      const sortedFiles = this.getSortedFiles();
       for (const filePath of sortedFiles) {
-        const fpl = filePath.toLowerCase();
-        if (fpl.endsWith(memberPart) && fpl.startsWith(normalizedDirPath)) {
-          return URI.parse(filePath);
+        if (
+          filePath.endsWith(memberPart) &&
+          filePath.startsWith(normalizedDirKey)
+        ) {
+          return UriUtils.toUri(filePath);
```
ssmifi, line 288: _Question_: Why do we need to adjust the schema before but not here? I think a comment explaining this would be sufficient.

packages/language/src/utils/uri.ts
```diff
@@ -29,66 +31,76 @@ export namespace UriUtils {
...
+  /**
+   * Smart constructor: detects whether input is a URI string or a file path
+   * and calls the correct vscode-uri factory.
+   * - URI objects are returned as-is
+   * - Strings starting with a Windows drive letter (e.g. C:\) go through URI.file
+   * - Strings with a URI scheme (file://, memory://) go through URI.parse
+   * - All other strings (bare paths) go through URI.file, which correctly
+   *   encodes special characters like '#' and spaces
+   */
+  export function toUri(input: string | URI): URI {
```
ssmifi, line 53: In general, it is nice to have a function that handles all the details and just returns a valid uri. In many cases this boils down to URI.parse -> UriUtils.toUri. I would double check the other cases if Uris are handled correctly there to justify this approach. But it already looks much cleaner than before.

packages/language/src/utils/uri.ts
```diff
@@ -29,66 +31,76 @@ export namespace UriUtils {
...
-        drive: match[1],
-      };
-    }
-
-    // No drive letter found
-    return {
-      path: path,
-      drive: null,
-    };
+  export function toNormalizedKey(input: URI | string): string {
```
ssmifi, line 88: _Question_: Is this safe for case-sensitive file systems?

packages/language/src/utils/uri.ts
```diff
@@ -29,66 +31,76 @@ export namespace UriUtils {
...
+    const uri = typeof input === "string" ? toUri(input) : input;
```
ssmifi, line 89: _Minor_: You could delegate this to `toUri`, right?

packages/language/src/utils/uri.ts
```diff
@@ -29,66 +31,76 @@ export namespace UriUtils {
...
-    if (driveLetter) {
-      result = result.replace(driveLetter, "");
+  /**
+   * Returns a normalized, decoded file path string with no scheme.
+   * - Fragments are re-appended (handles PL/I filenames containing '#')
+   * - Backslashes are normalized to forward slashes
+   * - Windows drive letters are uppercased and the URI leading slash is
+   *   stripped (e.g. /c:/path -> C:/path)
+   */
+  export function toFilePath(input: URI | string): string {
```
ssmifi, line 102: _Question_: Could we use this also as the key instead of `toNormalizedKey`? The question is probably about the lower case. Is this a requirement?

packages/language/src/utils/uri.ts
```diff
@@ -100,31 +112,36 @@ export namespace UriUtils {
...
-  // See GitHub Issue #568
-  function normalizeForWindowsTests(path: string) {
-    if (path.startsWith("/")) {
-      path = path.substring(1);
+  /**
+   * Strips a Windows drive letter prefix (e.g. "C:") from a path, returning
+   * the path portion and the extracted drive letter separately.
+   */
+  function splitDrive(path: string): { path: string; drive: string | null } {
+    if (/^[A-Z]:\//.test(path)) {
```
ssmifi: What about small drive letters?

packages/language/src/utils/uri.ts
```diff
@@ -100,31 +112,36 @@ export namespace UriUtils {
...
+    const fromParts = parts(fromPath);
+    const toParts = parts(toPath);
 
     const shareWorkspace = Boolean(
       fromParts.length && fromParts[0] === toParts[0],
     );
     if (isWindows && !shareWorkspace) {
-      const windowsPath = driveLetter ? driveLetter + toPath : toPath;
-      return normalizeForWindowsTests(windowsPath);
+      return driveLetter ? driveLetter + toPath : toPath;
```
ssmifi: This still relies on the runtime platform which makes it inconvenient to test. Can we make this platform-independent?
Also, shouldn't the drives be compared as well?

packages/language/src/utils/uri.ts
```diff
@@ -166,25 +174,27 @@ export namespace UriUtils {
     fallback = to,
     isWindows = UriUtils.isWindows,
```
ssmifi: The naming of the function is a bit misleading and same platform-dependency as before.

ssmifi (commented):
Thank you for pushing this. FS can be exhausting. The devil is in the detail. :)
In general this looks way cleaner than before, which is already a good sign. 
I left some higher level comments/questions to think about before marking this as ready, but in doubt just ask msujew. :)

packages/language/src/utils/uri.ts
```diff
@@ -29,68 +31,75 @@ export namespace UriUtils {
...
-    drive: string | null;
-  } {
-    // Check for leading slash before drive: /C:/ or /c:/
-    if (/^\/[a-zA-Z]:\//.test(path)) {
-      const drive = path.substring(1, 3); // Extract "C:" and lowercase
-      return {
-        path: path.substring(1), // Remove leading slash
-        drive: drive,
-      };
+  export function toUri(input: string | URI): URI {
```
msujew, line 53: Suggestion: There are two new occurences of `URI.parse` that make sense to be replaced in [here](https://github.com/zowe/zowe-pli-language-support/blob/f00cd4151e4335844ed76a468c423b968bc9619e/packages/language/src/preprocessor/instruction-interpreter.ts#L2567-L2575).

packages/language/src/utils/uri.ts
```diff
@@ -29,68 +31,75 @@ export namespace UriUtils {
...
+   * - Backslashes are normalized to forward slashes
+   * - Windows drive letters are uppercased and the URI leading slash is
+   *   stripped (e.g. /c:/path -> C:/path)
+   */
+  export function toFilePath(input: URI | string): string {
+    const uri = typeof input === "string" ? toUri(input) : input;
+    let path = uri.path;
+    if (uri.fragment) {
+      path += "#" + uri.fragment;
+    }
```
msujew: Question: I'm a bit sceptical about this in particular. I've delved into URI parsing/encoding/decoding lately again and this shouldn't be necessary. I.e. if we parse file URIs like `/file#with_fragment.pli` correctly, these should result in `file:///file%23with_fragment.pli` strings, without a fragment. Appending the fragment to the path is likely not the right approach - if we have to do something like this, then some previous step is wrong.

msujew (commented):
Looks really good, thank you!

I had to put one of our [test files](https://github.com/zowe/zowe-pli-language-support/blob/development/packages/language/test/fourslash/preprocessor/include/with-member-validation.todo.ts) into a `todo` state. Can you see whether it works again?

Wagner-Laranjeiras:
thanks for the review Mark! Working on it.

packages/language/src/utils/uri.ts
Wagner-Laranjeiras: My struggle currently is that in cases such as `UriUtils.toUri("file:///path/A1@#_$"), we need to use `.parse`, which messes with the path.

I even tried running `.file` and then `.parse`, but it still separates the scheme. We have no problem in the cases where `.file` are used, but everywhere we need `.parse` and we have a `#`, we need to append the fragment again.

I agree that it looks brittle. It also bothers me, but I can't think of another approach as of now. 😅

msujew: I see. In this case, we actually need to escape the `#` (i.e. replace it with `%23`) manually, at least assuming the file is using a `file` scheme. We should be able to do this in the `toUri` function. We can safely assume that file URIs don't make use of fragments.

Maybe, we can also do this before the `toUri` function. Ideally, we should be able to do this escaping whenever we see something that is *definitely* a file path. If there's no good way to do this though, placing the logic in `toUri` is fine with me 👍

Wagner-Laranjeiras: I dealt with it inside the `toUri`. The advantage of it is that we don't need to call another function to escape hashes manually every time we think is necessary. Instead we do it inside the function, for the schemes that we use (`file`, the ones you mentioned on your PR and AI suggested also `memory`).

msujew (approved):
Thanks, looks great 👍

packages/language/src/utils/uri.ts
Wagner-Laranjeiras: BTW, the test that was in `todo` is now active and passing. 😉

PR #652, opened 2026-03-23
Test quickfixes with the Fourslash framework

Lotes:
Plus: Migrated 2 quick fix test sections


packages/language/test/test-builder.ts
```diff
@@ -402,6 +408,81 @@ export class TestBuilder {
...
+    const doc = TextDocument.create("file://inmemory", "plaintext", 1, content);
+    return TextDocument.applyEdits(doc, edits);
+  }
+
+  private codeActionCacheByLabel = new Map<string, [string, CodeAction[]][]>();
+  async expectCodeActionAt(
+    label: string,
+    expectedActionLabel: string,
+    expectedCodeAfter: string,
+  ): Promise<void> {
```
msujew, line 421: Well, this doesn't really work. Since `expectCodeActionAt` is async, it will result in an uncaught-promise error if it fails and crash the test runtime. This needs an `await` in the test and some changes to the runner logic to make `await` as a top level statement available.

msujew (commented):
Ah, I forgot about the age-old issue of async tests...

msujew (approved):
Thank you, works great!

PR #653, opened 2026-03-23
Support SQL include statements

msujew:
Related to https://github.com/zowe/zowe-pli-language-support/issues/472.

Implements the requested `EXEC SQL INCLUDE file;` feature:

* Adds `SQLCA` and `SQLDA` as proper standalone builtin files.
* Adjusts our test builder a bit to improve URI handling (needed to deactivate one test file, is waiting on https://github.com/zowe/zowe-pli-language-support/pull/651)
* Other SQL statements simply fall into the `EmbeddedUnknownStatement` AST, which prevent any errors from appearing.


packages/language/src/language-server/text-documents.ts
```diff
@@ -444,6 +446,8 @@ export function resetDocumentProviders() {
   );
   BuiltinDocuments.set(BuiltinsTextDocument);
   BuiltinDocuments.set(BuiltinsMacroTextDocument);
+  BuiltinDocuments.set(BuiltinsSqlcaDocument);
+  BuiltinDocuments.set(BuiltinsSqldaDocument);
```
Lotes, line 450: I am wondering about the `set(...)` method. It makes me think that the previous value gets overwritten. Why is it not `add(...)`?

packages/language/src/parser/tokenizer/pli-tokenizer.ts
```diff
@@ -81,9 +81,11 @@ function tokenizeIdentifier(
   // Specific handling for EXEC (likely EXEC SQL or EXEC CICS)
   if (previousToken?.tokenTypeIdx === tokens.EXEC.tokenTypeIdx) {
     if (image === "SQL") {
+      context.advance(3, false);
       context.switchMode(TokenizerMode.SQL);
       return context.createTokenInstance(tokens.SQL);
```
Lotes, line 86: You don't have keywords or? When EXEC SQL appears we do not do a syntax highlight as in CICS, right?
During the CICS implementation I also had to swap the current keywords.

packages/language/src/language-server/text-documents.ts
msujew: The previous value is overriden - it usually just doesn't matter for the `BuiltinDocuments` in particular, since they are read-only. They just use the same interface as the `EditorDocuments` which actually make use of this behavior.

packages/language/src/parser/tokenizer/pli-tokenizer.ts
msujew: Not right now, no. But for this PR, it's not important.

packages/language/src/language-server/text-documents.ts
Lotes: Ah I see, it gets overridden when the URI is the same.
May call it `insertOrReplace` instead. `set` sounds absolute, which is not true as far as I can see.

packages/language/src/parser/tokenizer/pli-tokenizer.ts
```diff
@@ -81,9 +81,11 @@ function tokenizeIdentifier(
...
     } else if (image === "CICS") {
+      context.advance(4, false);
```
Lotes, line 88: Why is it advance(4) for CICS and in SQL advance(3)?

Lotes (approved):
Looks good. Only one question.

packages/language/src/language-server/text-documents.ts
msujew: `set` is also used by `Map` in JavaScript. So it is consistent with the existing naming conventions. Internally, it's a `Map#set` operation.

packages/language/src/parser/tokenizer/pli-tokenizer.ts
msujew: Because `"CICS".length === 4 && "SQL".length === 3`. We still need to generate the respective tokens in addition to changing the lexer mode.

Lotes: Ah! I though we advance in tokens. That was mit thinking mistake. Thx

packages/language/src/language-server/text-documents.ts
Lotes: That is true. I am just reflecting how I see it. `set(key, value)` has a different semantic as `set(value)`. The key is sticked to the value which makes one wondering how this function is meant.

Lotes: `setByDocumentUri(doc)` ? :D It is just a minor thing. I will not fight for it. It just made me wondering. During the read you have to lookup `set` to understand it. Good would be if the lookup would not be needed. That is my point. But I am not so sure if that is the case if we would change it now. Maybe the reader then also needs to lookup `setByDocumentUri` or whatever...

PR #655, opened 2026-03-25
Introduce ANY and VARARG for builtin declarations

Lotes:
ANY shall indicate that we accept any type as parameter.
VARARG shall indicate that the parameter is variadic. This implies it must be an array and that it only can be placed once at last parameter if a procedure.

Both KEYWORDS are not inside the PL/I language spec. They will be only used for builtin procedures.

### DIFF
--- a/packages/language/src/validation/internal-codes.ts
+++ b/packages/language/src/validation/internal-codes.ts
@@ -19,47 +19,6 @@ export const InternalCodes = {
     message: (label: string, file: string, uri: string) =>
       `Expected diagnostic at label "${label}" to be in file "${file}" but received: ${uri}`,
   },
-  UnknownIdentifier: {
-    code: "UNKNOWN_IDENTIFIER",
-    severity: Severity.E,
-    message: (name: string) => `Unknown identifier '${name}'`,
-  },
-  BuiltinAttributeUsage: {
-    code: "BUILTIN_ATTRIBUTE_USAGE",
-    severity: Severity.E,
-    message: (attribute: string) =>
-      `The attribute '${attribute}' is a builtin attribute and cannot be used in non-builtin files.`,
-  },
-  VariadicParameterNotLast: {
-    code: "VARIADIC_PARAMETER_NOT_LAST",
-    severity: Severity.E,
-    message: (name: string) =>
-      `The variadic parameter '${name}' must be the last parameter in the parameter list.`,
-  },
-  VariadicParameterMultiple: {
-    code: "VARIADIC_PARAMETER_MULTIPLE",
-    severity: Severity.E,
-    message: (name: string) =>
-      `The variadic parameter '${name}' cannot be used because there is already another variadic parameter in the parameter list.`,
-  },
-  VariadicParameterIsFixedArray: {
-    code: "VARIADIC_PARAMETER_IS_FIXED_ARRAY",
-    severity: Severity.E,
-    message: (name: string) =>
-      `The variadic parameter '${name}' cannot be a fixed array. Variadic parameters must be declared with an assumed size (e.g. (*)) and cannot specify a size.`,
-  },
-  VariadicParameterNotAnArray: {
-    code: "VARIADIC_PARAMETER_NOT_AN_ARRAY",
-    severity: Severity.E,
-    message: (name: string) =>
-      `The variadic parameter '${name}' must be an array. Variadic parameters must be declared with parentheses to indicate they are arrays (e.g. (*)).`,
-  },
-  VariadicNonParameter: {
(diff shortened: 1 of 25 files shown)


packages/language/test/fourslash-harness/harness-parser.ts
```diff
@@ -213,7 +213,7 @@ class HarnessTestParser {
         const uri =
           fileName === UnnamedFile
             ? DEFAULT_FILE_URI
-            : URI.file(fileName).toString();
+            : URI.parse(fileName).toString();
```
Lotes: In order to allow custom builtin files within the tests.

packages/language/src/validation/internal-codes.ts
```diff
@@ -24,6 +24,42 @@ export const InternalCodes = {
     severity: Severity.E,
     message: (name: string) => `Unknown identifier '${name}'`,
   },
+  BuiltinAttributeUsage: {
+    code: "BUILTIN_ATTRIBUTE_USAGE",
+    severity: Severity.E,
+    message: (attribute: string) =>
+      `The attribute '${attribute}' is a builtin attribute and cannot be used in non-builtin files.`,
+  },
```
msujew: Suggestion: `InternalCodes` are for codes that are internal to the language server. `DiagnosticURIMismatch` cannot be seen by a user of the language server. We should probably move everything else to somewhere else and add a comment here explaining that. I've overlooked this while reviewing https://github.com/zowe/zowe-pli-language-support/pull/552.

packages/language/src/validation/internal-codes.ts
```diff
@@ -24,6 +24,42 @@ export const InternalCodes = {
     severity: Severity.E,
     message: (name: string) => `Unknown identifier '${name}'`,
   },
+  BuiltinAttributeUsage: {
+    code: "BUILTIN_ATTRIBUTE_USAGE",
```
msujew: Suggestion: These codes are visible to the user. This will show up as `PL/I/(BUILTIN_ATTRIBUTE_USAGEE)` on diagnostics. Note the double `EE` at the end, because we append the severity to the code. They should follow the pattern of other LSP specific codes, like [here](https://github.com/zowe/zowe-pli-language-support/blob/development/packages/language/src/validation/lsp-codes.ts).

packages/language/src/validation/type-check-validator.ts
```diff
@@ -36,9 +40,89 @@ export function typeCheck(
     stmt,
     compilationUnit,
   );
-  if (TypeDescriptions.isUnknown(description)) {
-    // ignore further type checking
-    return;
+
+  if (compilationUnit.uri.scheme === BuiltinsUriSchema) {
```
msujew, line 44: Question: Is this testing that our builtins are correctly defined? I would assume that this isn't really necessary.

packages/language/src/validation/type-check-validator.ts
```diff
@@ -164,3 +248,65 @@ function validateBound(
...
+    index,
+    count: procedureStmt.parameters.length,
+  };
+}
+
+export function BUILTIN_NoMultipleVariadicParameters(
+  stmt: ast.ProcedureStatement,
+  acceptor: ValidationAcceptor,
+  compilationUnit: CompilationUnit,
+) {
```
msujew, line 301: Applies here as well.

packages/language/test/fourslash-harness/harness-parser.ts
msujew: See my comments below. I don't think there is a need to test/validate builtins.

packages/language/src/validation/type-check-validator.ts
Lotes: What sounds wrong in your eyes/ears :D? Is it the compilation unit that I am asking?
Or shall I reuse a function from us, if there is any?

packages/language/src/validation/internal-codes.ts
Lotes: So, internal codes should flag errors that are only internally relevant and should not bubble up to the outside?

packages/language/src/validation/type-check-validator.ts
Lotes: You mean the check whether we are in a builtin file, I guess...

packages/language/test/fourslash-harness/harness-parser.ts
Lotes: How would you do the test against a builtin file? The builtin status influences the validation as you can see.

Lotes: Comment for later revert added

msujew (approved):

PR #657, opened 2026-03-26
Support host variables in EXEC SQL statements

### LINKED ISSUE #393: Host variables support
Currently we have a support for [declaring host variables and indicator variables ](https://www.ibm.com/docs/en/db2-for-zos/12.0.0?topic=statements-declaring-host-variables-indicator-variables-in-pli) in PL/I. 

Now add host variables support inside PL/I exec sql statement,   see [data access with host variables ](
https://www.ibm.com/docs/en/db2-for-zos/12.0.0?topic=applications-data-access-host-variables)

e.g.` EXEC SQL PREPARE STATEMENT FROM :STATEMENT;`

`STATEMENT `is a field defined in PL/I code. 
` DCL STATEMENT CHAR(255) VARYING;`

e.g.1 
```
    EXEC SQL                                                                   
       FETCH SRV_DEF_CURSOR                                                     
       INTO :DCLPP_SRV_DFI_01.DOM_ID                                            
           ,:DCLPP_SRV_DFI_01.SRV_ID                                            
           ,:DCLPP_SRV_DFI_01.OPN_ID                                            
           ,:DCLPP_SRV_DFI_01.ORI_ID                                            
           ,:DCLPP_SRV_DFI_01.MAJ_VRS_ID                                        
           ,:DCLPP_SRV_DFI_01.CIC_SYS_CD;
```

e.g.2
```
 EXEC SQL
    SELECT EMPNO, LASTNAME, DEPT
        INTO :CBLEMPNO, :CBLNAME, :CBLDEPT
        FROM EMP
        WHERE EMPNO = :EMPID;

```

e.g.3
```
 EXEC SQL DESCRIBE STATEMENT INTO :SQLDA;
EXEC SQL DESCRIBE STATEMENT INTO :SQLCA;
```

e.g.4
```
EXEC SQL                                                    
                SELECT COUNT (*) INTO :WS-RESWORK-ROWCNT                 
                FROM EVENTS.ITEMS WHERE COLOR = ‘BLUE’;
```
      
e.g.5
```
 EXEC SQL PREPARE STATEMENT FROM :STATEMENT;
```

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/393

* Adjusts the existing `EmbeddedUnknownStatement` parser to also store the tokens used as host variables.
* The instruction generator emits each token as a `SqlHostVariableInstruction`
* The instruction interpreter emits a special token that indicates that the next token is a host variable reference
* The parser now has a new statement `Unit` that allows to reference these host variables

This is rather convoluted, but required in order to get linking to work as expected. Essentially, we require special semantics to forward tokens from the preprocessor to the normal parser so that these references to normal PL/I variables can be resolved.


packages/language/src/parser/unknown-parser.ts
```diff
@@ -30,10 +31,25 @@ export function embeddedUnknownStatement(
     if (!token) {
       break;
     }
-    token.element = unknownStatement;
-    token.kind = kind;
-    // Advance manually
-    state.index++;
+    if (token.tokenTypeIdx === t.Colon.tokenTypeIdx) {
```
Lotes, line 34: Will this not also enable CICS to have host variables? I used the same function...

Lotes: It looks like it is not the case. The error message is a bit cryptic, but I guess that is ok for now.

image

Lotes (approved):
Looks good.

packages/language/src/parser/unknown-parser.ts
msujew: AFAIK, the lexer for CICS does not support colons, so that's the reason for the cryptic error message.

Lotes: Yes. I am wondering if CICS needs linked Identifiers in this state?!
But that is not part of this PR...

PR #658, opened 2026-03-26
Annotate builtin preprocessor procedures

Lotes:


packages/language/src/workspace/builtins.ts
```diff
@@ -923,38 +923,383 @@ export const BuiltinsTextDocument = TextDocument.create(
...
+ /**
+  * \`COLLATE\` returns a \`CHARACTER\` string of length 256 comprising the 256 possible character
+  * values one time each in the collating order.
+  * @returns {CHARACTER} string of length 256 comprising the 256 possible character values one time each in the collating order
+  */
+ COLLATE: PROC RETURNS(CHARACTER); END;
+
+ /**
+  * \`COMMENT\` converts a \`CHARACTER\` expression into a comment.
+  * @param {CHARACTER} text Expression that is to be converted to a comment. \`text\` should have \`CHARACTER\` type, and if not, it is converted thereto.
```
msujew, line 935: Question: How should the markdown look like for this when we render this into a hover response? In TypeScript/Langium, we usually just use `@param text`, which gets rendered like this:

Image

Meaning it results in ``*@param* `text` - ...``.

msujew (approved):
Great, looks very nice! Just one question regarding the parameter documentation format.

packages/language/src/workspace/builtins.ts
Lotes: I think it is enough for the Hover to display the description and the procedure signature itself.
For the signature help service we will use the parameter descriptions.

msujew: Alright, makes sense. Feel free to merge 👍

PR #660, opened 2026-03-27
Add hover by using JSDoc comments for builtin preprocessor procedures

Lotes:
Uses JSDoc for retrieving documentation info and displays it in the hover tooltip.

I only show the signature and the JSDoc description. The parameters should be part of the signature help service, which will follow in a later PR.

image

FTM: This should only work for preprocessor builtins.


packages/language/test/fourslash/hover/type-related/functions-builtin.ts
```diff
@@ -15,7 +15,4 @@
 //// STR = <|1>SUBSTR('Hello, World!', 1, 5);
 
 //TODO fix the SUBSTR builtin signature, it is actually just a forward declaration
-hover.expectMarkdownAt(
-  1,
-  hover.codeBlock(`SUBSTR: PROC (value) RETURNS (); END;`),
-);
+hover.expectMarkdownAt(1, hover.codeBlock(`SUBSTR: PROC(VALUE) RETURNS();`));
```
Lotes: May revert this, only builtin preprocessor functions should be affected

packages/language/src/workspace/compilation-unit.ts
```diff
@@ -202,8 +213,14 @@ export async function createCompilationUnit(
       return cachedProcessGroup;
     },
     mutex: createMutex(),
-    reset() {
+    async reset() {
       services.files.clear();
+      if (uri.scheme !== BuiltinsUriSchema) {
```
Lotes: Feels very hacky...

packages/language/src/workspace/builtins.ts
Lotes, line 1: Keep in mind the right margin of 72 characters... also for comments

packages/language/src/utils/search.ts
```diff
@@ -80,3 +80,43 @@ export function completionTokenIndexSearch(
   }
   return end;
 }
+
+export function binaryTokenIndexLeftMost(
```
Lotes: May remove. I only need one of them

packages/language/src/language-server/hover-request.ts
```diff
@@ -423,3 +439,40 @@ export function hoverRequest(
...
+  const commentTokens = compilationUnit.services.files.getComments(uri);
+  if (!commentTokens) {
+    return null;
+  }
+  const index = binaryTokenIndexRightMost(
+    commentTokens,
+    labelPrefix.nameToken.startOffset,
+  );
+  if (index > -1) {
+    let description: string | null = null;
```
Lotes: Add: only for builtin macros FTM.

packages/language/src/documentation/jsdoc.ts
```diff
@@ -0,0 +1,764 @@
...
+   */
+  renderLink?(link: string, display: string): string | undefined;
+}
+
+/**
+ * Parses a JSDoc from a `CstNode` containing a comment.
+ *
+ * @param node A `CstNode` from a parsed Langium document.
+ * @param options Parsing options specialized to your language. See {@link JSDocParseOptions}.
+ */
```
msujew, line 94: Suggestion: This comment needs adjustments.

packages/language/src/documentation/jsdoc.ts
```diff
@@ -0,0 +1,764 @@
...
+  readonly start?: RegExp | string;
+  /**
+   * The symbol that start a line of your comment format. Defaults to `*`.
+   */
+  readonly line?: RegExp | string;
+  /**
+   * The end symbol of your comment format. Defaults to `*\/`.
+   */
+  readonly end?: RegExp | string;
+}
```
msujew: Suggestion: This is only required for a framework, not for a language that we control.

packages/language/src/language-server/hover-request.ts
```diff
@@ -325,6 +332,15 @@ const generateReferenceTokenMarkup: MarkupGenerator = ({ unit, token }) => {
...
     }
+    if (ref.node.kind === SyntaxKind.LabelPrefix) {
+      const jsDocsComment = getJSDocsCommentBeforeLabelPrefix(
+        ref.node as LabelPrefix,
+        unit,
+      );
+      if (jsDocsComment) {
+        return jsDocsComment;
+      }
+    }
```
msujew, line 338: Suggestion: The way this is coded will skip the type based hover. IMO I would prefer both comment & type info.

packages/language/src/language-server/hover-request.ts
```diff
@@ -423,3 +439,40 @@ export function hoverRequest(
...
+    return null;
+  }
+  const commentTokens = compilationUnit.services.files.getComments(uri);
+  if (!commentTokens) {
+    return null;
+  }
+  const index = binaryTokenIndexRightMost(
+    commentTokens,
+    labelPrefix.nameToken.startOffset,
+  );
```
msujew: Suggestion: This is too eager. In the following example, it will show the comment for the hover of `PROC2`:

```
/** SOME HOVER */
PROC1: PROC; END;
PROC2: PROC; END;
X = PROC2(); // <-- shows SOME HOVER as hover
```

Make sure there's a test case for this.

packages/language/src/workspace/compilation-unit.ts
```diff
@@ -202,8 +213,14 @@ export async function createCompilationUnit(
...
+        const unit = await getBuiltinUnit();
+        const macroUnit = await getBuiltinMacroUnit();
+        services.files.set(unit.services.files.get(unit.uri)!);
+        services.files.set(macroUnit.services.files.get(macroUnit.uri)!);
+      }
```
msujew: Suggestion: Since these files are available on every unit, the better abstraction would be to implement this behavior in the `FileStore` itself? We already have some behavior like this related to the `TextDocuments` of builtins vs editor documents.

packages/language/src/workspace/file-store.ts
```diff
@@ -36,8 +37,11 @@ export class FileStore {
...
 
-  *getAllTokens(): IterableIterator<Token> {
+  *getAllTokens(includeBuiltins = false): IterableIterator<Token> {
     for (const file of this.map.values()) {
+      if (!includeBuiltins && file.uri.scheme === BuiltinsUriSchema) {
+        continue;
+      }
       yield* file.tokens;
     }
   }
```
msujew, line 55: The suggestion above would also make this workaround obsolete.

packages/language/src/documentation/jsdoc.ts
msujew, line 1: Suggestion: Please add the MIT license header from Langium here - otherwise it would be a license violation.

msujew (commented):
Makes sense to include the Langium jsdoc parser in here 👍

There are a few issues though that make the hover behave incorrectly though. See my comments below.

packages/language/src/documentation/jsdoc.ts
Lotes: Needed to extend our License header checker by introducing primary (EPL/Zowe) and secondary headers (MIT/Typefox).

msujew:
@Lotes There is no need to remove the IPL from the code - both licenses should be placed in the file.

packages/language/src/language-server/hover-request.ts
Lotes: True, got mixed up function calls in a weird way. Fixed.

packages/language/src/workspace/file-store.ts
Lotes: I could not get rid of this, but it changed slightly. What was your idea here?

packages/language/src/workspace/compilation-unit.ts
Lotes: Like this?

msujew: Yep, that works for me 👍

packages/language/src/utils/search.ts
```diff
@@ -80,3 +80,23 @@ export function completionTokenIndexSearch(
   }
   return end;
 }
+
+export function binaryTokenIndexRightMost(
+  tokens: Token[],
+  offset: number,
+): number {
```
msujew, line 107: Suggestion: Please add a comment explaining what this function does explicitly.

packages/language/src/language-server/hover-request.ts
```diff
@@ -423,3 +441,54 @@ export function hoverRequest(
...
+      if (tokenBeforeLabelPrefix.startOffset > commentToken.endOffset) {
+        return null;
+      }
+    }
+    if (isJSDoc(commentToken)) {
+      const jsDoc = parseJSDoc(commentToken);
+      return takeWhile(jsDoc.elements, isJSDocParagraph)
+        .flatMap((p) => (p as JSDocParagraph).inlines)
+        .map((inline) => inline.toMarkdown())
+        .join("\n");
```
msujew: Suggestion: It probably makes more sense to use `jsDoc.toMarkdown()`?

packages/language/src/workspace/file-store.ts
```diff
@@ -37,7 +42,13 @@ export class FileStore {
...
 
   *getAllTokens(): IterableIterator<Token> {
+    const baseFileUris = new Set(
+      this.baseFiles.map((file) => file.uri.toString()),
+    );
```
msujew: Suggestion: I don't believe these change, can we precompute them?

packages/language/src/language-server/hover-request.ts
```diff
@@ -423,3 +441,54 @@ export function hoverRequest(
...
+       * Edge case: two declarations, only first has comment, second gets hovered:
+       * /** comment *\/
+       * DCL1
+       * DCL2 //<-- hover here: no comment included
+       */
+      const tokenBeforeLabelPrefix = tokens[tokenIndex - 1];
+      if (tokenBeforeLabelPrefix.startOffset > commentToken.endOffset) {
+        return null;
+      }
+    }
```
msujew, line 491: Question: How do we want to deal with procedures with aliases? I.e. `COMPLEX` and `CPLX`? As far as I can tell, this will only provide a hover, if we explicitly write something like this, right?

```
/** Comment A */
COMPLEX:
/** Comment B */
CPLX: ...
```

msujew (commented):
Looks pretty good now. Just a few improvements, see below 👍

packages/language/src/language-server/hover-request.ts
Lotes: Good catch...

packages/language/src/language-server/hover-request.ts
Lotes: I wanted to exclude the params & the returns. Because the params are already in the signature help. But I can add all of the information. Should be not a problem.

packages/language/src/language-server/hover-request.ts
Lotes: I think we just have to scan the tokens backwards and take the first comment. We skip all IDs and colon, as far as needed.
I can add a test for this scenario. In the preprocessor builtins there was actually only one case that I have overlooked.

Lotes: Since builtins are under our control, I would like to proceed with the current solution.

Lotes: Currently I fast forward to the first alias.

packages/language/src/language-server/hover-request.ts
```diff
@@ -217,7 +222,7 @@ function getIncludeItemRepresentation(
   if (!node.filePath || !node.relativeFilePath) {
     return null;
   }
-  const fileUri = UriUtils.toUri(node.filePath);
+  const fileUri = URI.parse(node.filePath);
```
msujew: Question: `toUri` should call `parse` if it matches the expected format. Does this not work as expected here?

msujew (approved):
Thanks, looks good 👍

packages/language/src/language-server/hover-request.ts
Lotes: Puh do not ask ^^... maybe I was confused seeing that FILE-uri
I changed it back. Crazy.

PR #666, opened 2026-04-02
[Improvement] Implement `json-parser`, LS attachment for `.pliplugin`, and quick fixes for unresolved `proc_grps` libs

Wagner-Laranjeiras:
_Closes_ #535 , _but also implement other changes. I wanted to make sure we would be able to act on the diagnostics, so I worked on implementing at least a quick fix, to make it easier to do it in the future. Turns out some other changes were necessary to accomplish it, so we ended up with such a big PR._

## Summary

This work improves how `.pliplugin` JSON files (especially `proc_grps.json`) are handled end-to-end:

- The language server can now participate in these documents **without running the PL/I pipeline**
- Plugin configuration validation surfaces **clearer diagnostics**
- Code actions can remove invalid `libs` entries:
  - Individually
  - Or in bulk
- Uses `jsonc-parser` to ensure edits remain **structural and JSONC-tolerant**

---

### VS Code Extension

- Extended `documentSelector` to include:
  - `**/.pliplugin/*.json`
- This allows the PL/I language client to:
  - Participate in plugin config files
  - Enable code actions and related LSP behavior

---

### Language Server – Compilation

- **`compilation-unit.ts`**
  - Skips creating compilation units for:
    - Paths under `.pliplugin/`
    - Files ending in `.json`
- Result:
  - Only plugin-related diagnostics (e.g., `COPC*`) are applied
  - PL/I compiler diagnostics are avoided for JSON files

---

### Plugin Configuration Provider

- Improved validation and diagnostics for process-group configuration:
  - Includes `COPC01`-style unresolved library entries
- Maintains a snapshot:
  - `lastProcGrpsUnresolvedLibEntries`
- Enables code actions to:
  - Access **all unresolved `(pgroup, lib)` pairs**
  - Not just those present in a single request

---

### Quick Fixes (`apply-quick-fixes.ts`)

- Uses `jsonc-parser`:
  - `parseTree`
  - `findNodeAtLocation`
  - `modify` / `applyEdits`
- Removes `libs[]` entries by **logical path**
- Ensures **stable formatting**

#### Behavior

- `applyQuickFixes(diagnostics, documentUri)`:
  - When targeting workspace `proc_grps.json`
  - And at least one unresolved-lib diagnostic is present:
    - Builds fixes from provider snapshot:
      - Per-lib removal
      - “Remove all” (if ≥ 2 distinct pairs)
  - Avoids duplicating per-diagnostic logic

---

### Connection Handler

- Passes `params.textDocument.uri` into `applyQuickFixes`
- Wires command (e.g., `REMOVE_DEAD_LIB`) through `executeCommand`

---

## Dependencies

- Adds `jsonc-parser`
- Uses bundler-safe import:
  - `jsonc-parser/lib/esm/main.js`
- Avoids runtime module resolution issues

---

## Tests

- Extended:
  - `apply-quick-fixes.test.ts`
    - `quickFixRemoveUnresolvedLib`
    - `quickFixRemoveAllUnresolvedLibsFromPairs`
    - Snapshot + `documentUri` behavior
- Adjusted plugin configuration tests as needed

---

## How to Verify

1. Open:
   - `/.pliplugin/proc_grps.json`
   - Include multiple invalid `libs` entries

2. Confirm:
   - Diagnostics are shown
   - Quick fixes include:
     - Individual removals
     - “Remove all” (when applicable)

3. Verify:
   - `.pliplugin` JSON does **not** show PL/I diagnostics

### DIFF
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -46,30 +46,6 @@ function isProcGrpsDocumentUri(documentUri: string): boolean {
   return UriUtils.equals(documentUri, expected);
 }
 
-async function readProcGrpsText(): Promise<
-  { uri: URI; text: string } | undefined
-> {
-  const workspacePath = PluginConfigurationProviderInstance.getWorkspacePath();
-  if (!workspacePath) {
-    return undefined;
-  }
-  const procGrpsUri = UriUtils.joinPath(
-    UriUtils.toUri(workspacePath),
-    ".pliplugin",
-    "proc_grps.json",
-  );
-  try {
-    const content = await FileSystemProviderInstance.readFile(procGrpsUri);
-    if (content === undefined) {
-      return undefined;
-    }
-    return { uri: procGrpsUri, text: content };
-  } catch (err) {
-    console.error(err);
-    return undefined;
-  }
-}
-
 export async function quickFixResolveInclude(
   diagnostic: Diagnostic,
 ): Promise<CodeAction | undefined> {
@@ -303,17 +279,21 @@ export async function quickFixRemoveUnresolvedLib(
   if (!data.path) {
     return undefined;
   }
-  const read = await readProcGrpsText();
-  if (!read) {
-    return undefined;
+  const procGrpsSnapshot =
+    PluginConfigurationProviderInstance.getLastProcGrpsSnapshot();
+  if (!procGrpsSnapshot) {
+    return;
   }
-  const { uri: procGrpsUri, text } = read;
+  const { text: procGrpsText, uri: procGrpsUri } = procGrpsSnapshot;
+  if (!procGrpsText || !procGrpsUri) {
+    return;
+  }
 
   let newContent: string;
   try {
     newContent = jsoncApplyEdits(
-      text,
-      jsoncModify(text, data.path, undefined, JSONC_FORMAT),
+      procGrpsText,
+      jsoncModify(procGrpsText, data.path, undefined, JSONC_FORMAT),
     );
   } catch (err) {
     console.error("Failed to build proc_grps edit for remove lib:", err);
@@ -338,6 +318,8 @@ export async function quickFixRemoveUnresolvedLib(
  */
 export async function quickFixRemoveAllUnresolvedLibs(
   pairs: readonly PluginConfigUnresolvedLibData[],
+  procGrpsSnapshotText: string,
+  procGrpsSnapshotUri: URI,
   relatedDiagnostics: Diagnostic[],
 ): Promise<CodeAction | undefined> {
   const unique: (PluginConfigUnresolvedLibData & { path: JSONPath })[] = [];
@@ -356,13 +338,7 @@ export async function quickFixRemoveAllUnresolvedLibs(
   if (unique.length < 2) {
     return undefined;
   }
-
-  const read = await readProcGrpsText();
-  if (!read) {
-    return undefined;
-  }
-  const { uri, text: initialText } = read;
-  let text = initialText;
+  let text = procGrpsSnapshotText;
 
   unique.sort((a, b) => {
     // Sort from highest index first so earlier indices remain valid when removing right-to-left.
@@ -371,11 +347,10 @@ export async function quickFixRemoveAllUnresolvedLibs(
 
   for (const { path } of unique) {
     try {
-      text = jsoncApplyEdits(
-        text,
-        jsoncModify(text, path, undefined, JSONC_FORMAT),
-      );
+      const edits = jsoncModify(text, path, undefined, JSONC_FORMAT);
+      text = jsoncApplyEdits(text, edits);
     } catch (err) {
+      console.error("Failed at path:", JSON.stringify(path));
       console.error(
         "Failed to build proc_grps edit from unresolved lib path:",
         err,
@@ -391,7 +366,7 @@ export async function quickFixRemoveAllUnresolvedLibs(
     command: {
       title: "Remove all unresolved libs",
       command: Commands.REMOVE_DEAD_LIB,
-      arguments: [uri.toString(), text],
+      arguments: [procGrpsSnapshotUri.toString(), text],
     },
   };
 }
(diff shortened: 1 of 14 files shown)


packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -26,6 +27,90 @@ import { Commands, PluginConfiguration } from "../constants";
...
 import { PLICodes } from "../../validation/pli-codes";
+import {
+  applyEdits,
+  findNodeAtLocation,
+  modify,
+  parseTree,
+  type JSONPath,
+  type Node,
+} from "jsonc-parser/lib/esm/main.js";
+
```
msujew, line 31: Suggestion: I understand why we have to do things like this. However, to prevent errors where we accidentally import `jsonc-parser`, can we expose this API through a separate file? Similar to how we do it for `Uri` already.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -26,6 +27,90 @@ import { Commands, PluginConfiguration } from "../constants";
...
+} as const;
+
+/**
+ * JSON path for `modify`/`remove` of one `libs[]` string entry, or undefined if not found.
+ */
+function findLibRemovalPath(
+  root: Node,
+  pgroup: string,
+  lib: string,
+): JSONPath | undefined {
```
msujew: Info: This doesn't work well in the presence of duplicate entries. AFAIK this will simply remove the first entry it can find, not the one that the user wanted to modify. See the PR comment above on how to fix this.

msujew (commented):
This works pretty well already for the issue at hand, thanks!

However, for some of the things I have in mind, this design won't be capable enough. Concretely, I want to be able to retrieve any position+uri+path info of any config field. Not just the libs. The current design isn't extensible without replicating code.

The idea would be to have something like this:

```
interface JsonItem<T> {
  value: T;
  meta?: JsonItemMeta // optional
}

interface JsonItemMeta {
  range: Range;
  uri: URI;
  path: JsonPath
}
```

This data should already be gathered during parsing, so that we don't have to look it up after the fact. Looking up JSON paths based on concrete values is error-prone (see https://github.com/zowe/zowe-pli-language-support/pull/666#discussion_r3044691787).

What do you think about this? You've been digging deeper into this, so I'm curious about your input.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -250,8 +293,168 @@ export function quickFixResolveAmbiguousReference(
...
+  if (aParent === bParent) {
+    const aIdx = pathIndex(aPath);
+    const bIdx = pathIndex(bPath);
+    if (aIdx !== undefined && bIdx !== undefined) {
+      // Remove from right to left in the same array so indices stay valid.
+      return bIdx - aIdx;
+    }
+  }
+  return pathIdentity(aPath).localeCompare(pathIdentity(bPath));
+}
```
msujew: Suggestion: I feel like this is overly complicated. Can we create a string representation of `JSONPath` (like by using `path.join("/")` and then compare it? That should be more than fast enough.

packages/language/test/lsp/apply-quick-fixes.test.ts
msujew, line 1: Question: Can we try to migrate some of these tests to the new `verify.expectCodeActionAt` API? Or do these tests require too much setup work?

packages/language/src/utils/jsonc.ts
```diff
@@ -0,0 +1,35 @@
...
+export {
+  applyEdits as jsoncApplyEdits,
+  findNodeAtLocation as jsoncFindNodeAtLocation,
+  modify as jsoncModify,
+  parseTree as jsoncParseTree,
+  type JSONPath,
+  type Node,
+  parse as jsoncParse,
+  printParseErrorCode as jsoncPrintParseErrorCode,
+  type Node as JsonNode,
```
msujew: Minor: `type Node` exported twice?

msujew (commented):
I noticed some unexpected output when removing multiple libs:

```
"libs": [
        "cpy",
        "more-test", // invalid
        "test" // invalid
      ],
```

When removing both with one code action. I get this error in the output:

```
Failed to build proc_grps edit from unresolved lib path: TypeError: Cannot read properties of undefined (reading 'offset')
    at setProperty (c:\Users\Mark\Source\Repos\zowe-pli-language-support\packages\vscode-extension\out\language\main.cjs:65641:35)
    at modify (c:\Users\Mark\Source\Repos\zowe-pli-language-support\packages\vscode-extension\out\language\main.cjs:65786:10)
    at quickFixRemoveAllUnresolvedLibs (c:\Users\Mark\Source\Repos\zowe-pli-language-support\packages\vscode-extension\out\language\main.cjs:79189:31)
    at async handleMultipleUnresolvedLibs (c:\Users\Mark\Source\Repos\zowe-pli-language-support\packages\vscode-extension\out\language\main.cjs:79212:18)
    at async applyQuickFixes (c:\Users\Mark\Source\Repos\zowe-pli-language-support\packages\vscode-extension\out\language\main.cjs:79222:36)
    at async c:\Users\Mark\Source\Repos\zowe-pli-language-support\packages\vscode-extension\out\language\main.cjs:79635:23
```

Wagner-Laranjeiras:
Thanks Mark, gonna work on the items you pointed out!

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -250,8 +269,141 @@ export function quickFixResolveAmbiguousReference(
...
+      text = jsoncApplyEdits(text, edits);
+    } catch (err) {
+      console.error("Failed at path:", JSON.stringify(path));
+      console.error(
+        "Failed to build proc_grps edit from unresolved lib path:",
+        err,
+      );
+      return undefined;
+    }
+  }
```
msujew, line 360: Suggestion: Since `jsoncModify` returns an `Edit[]`, can't we merge all those edits together into one edit array that is only applied once? That should remove the need to sort the `unique` elements in the first place.

msujew (approved):
Looks good to me now. Thank you!

Just one suggestion for a code improvement, see below.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
Wagner-Laranjeiras: Hey Mark, thanks again for the review 😊 

I've tried like this:

```
  let text = procGrpsSnapshotText;
  const allEdits: JsonEdit[] = [];
  for (const { path } of unique) {
    try {
      const currentEdit = jsoncModify(text, path, undefined, JSONC_FORMAT);
      allEdits.push(...currentEdit);
    } catch (err) {
      console.error("Failed at path:", JSON.stringify(path));
      console.error(
        "Failed to build proc_grps edit from unresolved lib path:",
        err,
      );
      return undefined;
    }
  }
  try {
    text = jsoncApplyEdits(text, allEdits);
  } catch (err) {
    console.error("Failed to apply edits: ", err);
    return undefined;
  }
```

but I've got an "Overlapping edit" error. When I debug it (used also AI) it seemed like it's a formatting problem. When we remove each un resolved lib, the spaces and the comma around it also go away. When we try to remove multiple items at once, the spacings after one lib would collapse with the spacings before the next. If we drop the formatting, the document starts getting ugly.

It wouldn't be a problem at all to keep investigating for a solution, but would it be worth the time?

msujew: Ah, I see. Alright, then please ignore this 👍

PR #667, opened 2026-04-07
Add SQL lexer tokens

msujew:
Adds all the tokens that the COBOL ls also supports. Only the tokens, no parsing :)


packages/language/src/parser/tokenizer/shared.ts
```diff
@@ -49,23 +49,19 @@ export class TokenizerContext {
...
-        this.keywords = pliKeywords;
         break;
       case TokenizerMode.CICS:
         this.funcs = cicsFuncs;
-        this.keywords = cicsKeywords;
         break;
       case TokenizerMode.SQL:
-        this.funcs = pliFuncs; // switch to sqlFuncs when implemented
-        this.keywords = pliKeywords; // switch to sqlKeywords when implemented
+        this.funcs = sqlFuncs; // switch to sqlFuncs when implemented
```
Lotes: You could remove the comment, since it is implemented

packages/language/src/parser/tokens/sql-tokens.ts
```diff
@@ -0,0 +1,1008 @@
...
+  name: "]",
+  pattern: Lexer.NA,
+});
+export const QuestionMark = createToken({
+  name: "?",
+  pattern: Lexer.NA,
+});
+
+// Literal tokens (beyond the standard string and numeric literals)
+// TODO
```
Lotes: TODO, can we grab some Regexp from the COBOL LS?

scripts/generate-tmlanguage.mts
```diff
@@ -23,30 +25,105 @@ const manual = JSON.parse(
...
+function longestCommonPrefix(words: string[]): string {
+  if (words.length === 0) return "";
+  let prefix = words[0];
+  for (let i = 1; i < words.length; i++) {
+    let j = 0;
+    while (
+      j < prefix.length &&
+      j < words[i].length &&
+      prefix[j] === words[i][j]
+    )
```
Lotes: { ... } missing.

scripts/generate-tmlanguage.mts
```diff
@@ -23,30 +25,105 @@ const manual = JSON.parse(
...
 }
 
-function toPattern(keywords: string[]) {
-  const patterns: string[] = [];
-  for (const keyword of keywords) {
-    let keywordPattern = "";
-    for (const char of keyword) {
-      keywordPattern += tokens.escapeRegExp(char);
+// Unnecessary (?) optimization to avoid generating a huge regex with each keyword as an alternative
+// Instead, we serialize a trie structure into the regex, which should be more efficient to execute
```
Lotes: A trie? Interesting! Can you give a minimal example with three words with some overlap in the prefix? I am a bit curious how it would look like.

Lotes: Looks like you are refering to a whole section of code and not just the next function...

scripts/generate-tmlanguage.mts
Lotes: As well as below... I do not insist on having these extra braces, but I thought it is a sort of standard we are following.

Lotes (commented):
Looks good so far. I like the common patterns between the different languages. Very nice.

scripts/generate-tmlanguage.mts
msujew: Yes, it's this + the following functions until `toPattern`.

packages/language/src/parser/tokens/sql-tokens.ts
msujew: I've actually removed those now. They are just special string literals. During parsing, we should ignore them (handle them like normal string literals), and perhaps add code in the validation. The COBOL ls ANTLR grammar uses them, but it's completely unnecessary.

scripts/generate-tmlanguage.mts
```diff
@@ -23,30 +25,105 @@ const manual = JSON.parse(
...
   }
 }
 
-function toPattern(keywords: string[]) {
-  const patterns: string[] = [];
-  for (const keyword of keywords) {
-    let keywordPattern = "";
-    for (const char of keyword) {
-      keywordPattern += tokens.escapeRegExp(char);
+// Unnecessary (?) optimization to avoid generating a huge regex with each keyword as an alternative
```
Lotes: Is it really unnecessary? Theoretically: You could benchmark with the code before and after this change...

scripts/generate-tmlanguage.mts
Lotes: You could move the whole trie computation to an extra file (one responsibility). Then the comment can be on top of the file and can explain more details.

scripts/generate-tmlanguage.mts
Lotes: I would expect that the RegExp engine would create a minimal DFA in the end.
Is it about optimizing the construction time?

msujew: I would also think that the RegExp engine itself would create a minimal DFA. In our case, it simply generates a shorter regexp which ends up in the vscode extension.

Lotes: We agreed on removing the trie. It is simpler and we can not measure any performance gain.

Lotes (approved):

PR #669, opened 2026-04-10
Add signature help for builtin procedures

Lotes:
This is a follow-up-PR on #660 . **Please review that PR first, so I can rebase.**

image

This PR adds a signature help. It will display a parameter info, when you are in a procedure invocation and hit Cmd+Shift+Space (on Mac). It also gets triggered after hitting `(` (open parenthesis) or `,` (comma).


msujew:
FYI I changed the PR base to be able to review this properly. Please change it back to dev after merging https://github.com/zowe/zowe-pli-language-support/pull/660.

packages/language/src/documentation/jsdoc.ts
```diff
@@ -377,10 +377,20 @@ function parseJSDocComment(context: ParseContext): JSDocComment {
...
+      if (isJSDocParagraph(element)) {
+        if (!lastTagElement) {
+          elements.push(element);
+        } else {
+          lastTagElement.content.inlines.push(...element.inlines);
+        }
+      } else {
+        lastTagElement = element;
+        elements.push(element);
+      }
```
msujew, line 393: Question: What issue issue is this fixing?

packages/language/src/parser/parser.ts
msujew, line 1: Question: The changes in this file seem rather convoluted to me. Tracking tokens through multiple parser functions, loops, re-assigning them, and so on. Isn't using `state.token` (at the start) and `state.last` (at the end) sufficient to track the start-end positions of all statements?

packages/language/src/syntax-tree/ast.ts
```diff
@@ -3494,16 +3508,33 @@ export function createProcedureCall(): ProcedureCall {
...
+}
+
+export function createProcedureCallArgumentBounds(): ProcedureCallArgumentBounds {
+  return {
+    kind: SyntaxKind.ProcedureCallArgumentBounds,
+    container: null,
+    startToken: null,
+    endToken: null,
+  };
+}
```
msujew, line 3524: Suggestion: This seems like a workaround due to the poor existing structure, rather than a genuine AST node. I had this idea for a while: Why don't we replace the part after `CALL` simply with a `MemberCall` in the AST? The current structure leads to a lot of code duplication.

packages/language/src/language-server/signature-help-request.ts
```diff
@@ -0,0 +1,280 @@
...
+                    kind: "markdown",
+                    value: parameterDocumentation.get(name)!,
+                  }
+                : undefined,
+          };
+        }),
+      },
+    ],
+    activeSignature: 0,
+    activeParameter: callInfo.arguments[callInfo.argumentIndex].parameterIndex,
```
msujew: I'm getting errors during the runtime of the language server here in the presence of an empty argument list:

```
[Error - 12:11:05 PM] Request textDocument/signatureHelp failed.
  Message: Request textDocument/signatureHelp failed with message: Cannot read properties of undefined (reading 'parameterIndex')
  Code: -32603
```

Example:

```
 %DCL A CHARACTER;
 %A = SUBSTR // <-- type a `(` after SUBSTR
```

packages/language/test/fourslash/signature-help/help-builtin-procedure.ts
```diff
@@ -0,0 +1,34 @@
...
+  hover.codeBlock(`SUBSTR: PROC(string, offset, length) RETURNS(CHARACTER);
+   DECLARE string CHARACTER;
+   DECLARE offset FIXED;
+   DECLARE length FIXED OPTIONAL;
+ END;`);
+const expectedParameterMarkdown = `{FIXED} offset Expression that specifies the starting
+position of the substring in \`string\`.
+
+\`offset\` should have \`FIXED\` type, and if not, it is
+converted thereto.`;
```
msujew: Suggestion: Now that I see this in action, I'm not too fond of the `{...}` notation at the start. It is `code`, but isn't marked as such in the markdown.

Image

msujew (commented):
I noticed a couple of issues that I didn't trace down in the code:

1. The feature supplies the first argument signature help even on the procedure token itself. It should only do that after the `(` token:

Image

2. The result of the signature help request isn't really in the expected format. It should provide the whole procedure signature with the parameter highlighted, then a comment - this would be in line with how other language servers operate. Instead, we provide a label, then comment, then the whole signature without highlighting. See this example in TypeScript:

Image

Note: You can trigger the signature help manually using `Ctrl+Shift+Space` (probably `Cmd+Shift+Space` on Mac).

packages/language/src/documentation/jsdoc.ts
Lotes: My JSDoc comments are a mix of JSDocTags and JSDocParagraphs. I would expect to have only the first part of a JSDOC to be a paragraph and the rest are tags with some line breaks and empty lines in it.
The problem with the input comments is that they 1. need to be wrapped after margin 72 and 2. they have multiline tags, like a parameter with multiple explanations.

packages/language/src/parser/parser.ts
Lotes: I need the commas and opening/closing brackets. The start-end span contains also the whitespaces (or comments). I can try to use .token and .last instead, but I would prefer having the current given spans.
The code becomes more complicated, but maybe I can apply some pattern to hide the complexity a bit. Look at the examples where the "apply..." function helps to keep track.

packages/language/test/fourslash/signature-help/help-builtin-procedure.ts
Lotes: Now I added a label with the entire signature, where the selected parameter is marked.
Followed by a line `NAME: TYPE`.
Followed by the parameter documentation.
Followed by the whole definition of the procedure (should I remove this maybe?).

image

msujew: Great! Looks much better. I'd also be in favor of removing the whole definition. Users can always see it by hovering over the procedure name.

Lotes: OK! But what shall I put into the documentation field then?
I would suggest to put the first JSDoc paragraphs until the first tag.

Lotes: Or put nothing in there.

packages/language/src/syntax-tree/ast.ts
Lotes: I think that is a quite big change. I suggest to create another PR before merging this one about making this change happen first. Then, I would have to pause this PR for a while. Or I finish this one first and refactor the approach after changing the CallStatement to use a member call.

packages/language/test/fourslash/signature-help/help-builtin-procedure.ts
msujew: > I would suggest to put the first JSDoc paragraphs until the first tag.

Yes, that's what it's meant for. The general documentation of the respective function without the parameter/return type information. See for example:

image

Is generated from this documentation:

image

packages/language/test/fourslash/signature-help/help-precision.ts
```diff
@@ -0,0 +1,36 @@
...
+
+signatureHelp.expectParameterIndexAt("g", 1);
+signatureHelp.expectParameterIndexAt("h", 1);
+
+signatureHelp.expectParameterIndexAt("i", 2);
+signatureHelp.expectParameterIndexAt("j", 2);
+signatureHelp.expectParameterIndexAt("k", 2);
+
+signatureHelp.expectNoHelp("l"); //; belongs to the statement, not the call
+signatureHelp.expectNoHelp("m");
```
msujew: Suggestion: Please give these labels some better names.

msujew (approved):
Thanks, looks good 👍

Please create an issue for the CallStatement refactoring :)

packages/language/src/syntax-tree/ast.ts
Lotes: [Issue](https://github.com/zowe/zowe-pli-language-support/issues/672) created.

packages/language/src/parser/parser.ts
Lotes: Will put it into issue #672

packages/language/test/fourslash/signature-help/help-precision.ts
Lotes: Done, thanks

PR #675, opened 2026-04-14
Refactor ProcedureCall

### LINKED ISSUE #672: Parser refactoring: Reuse MemberCall within CallStatement
Needed for #622 

Currently `MemberCall` is used to model procedure calls on procedures with a `returns` attribute.
All procedures without `returns` attribute are called using the `CallStatement`.

But `CallStatement` handle arguments differently. So we have 2 different parsers for argument passing.

The task is to reuse the `MemberCall` to reduce redundant code.
There is also the notion of statement procedures, where the arguments are passed via parameter name.

Also mind the assignment of `startToken/endToken` for arguments. See [here](https://github.com/zowe/zowe-pli-language-support/pull/669#discussion_r3072235687).

Lotes:
This pull request refactors how procedure calls are represented in the language's AST and parser, unifying them under the `ReferenceItem` node and removing the older `ProcedureCall` structure. Argument lists and dimensions are now consistently handled as arrays, and related logic in the parser, AST, and code generation has been updated to match. This simplifies the codebase and makes procedure calls more flexible and uniform.

### DIFF
--- a/packages/language/src/language-server/signature-help-request.ts
+++ b/packages/language/src/language-server/signature-help-request.ts
@@ -19,6 +19,7 @@ import {
   LabelPrefix,
   MemberCall,
   ProcedureStatement,
+  ReferenceItem,
   SyntaxKind,
 } from "../syntax-tree/ast";
 import { extractDeclaration } from "../typesystem/stringify";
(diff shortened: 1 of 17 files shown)


packages/language/src/language-server/signature-help-request.ts
```diff
@@ -135,102 +136,42 @@ function tryGetCallInfo(
...
-    return getCallInfoFromCallStatement(callStatement, offset, unit);
+  if (callStatement && callStatement.call && !memberCall) {
+    return getCallInfoFromReferenceItem(callStatement.call, offset, unit);
   }
   assertType<MemberCall>(memberCall);
   assertType<CallStatement>(callStatement);
-  const callStatementOffset = callStatement.call?.procedure?.token.startOffset;
+  if (memberCall.element === null || callStatement.call === null) {
+    return null;
+  }
```
msujew, line 149: Suggestion: I don't think a call statement can contain a member call, and a member call can never contain a call statement. This code is rather complicated. I believe this section can be completely removed.

packages/language/src/parser/parser.ts
```diff
@@ -6149,13 +6149,19 @@ const referenceItem = rule(
...
+    ) {
+      if (state.canConsumeFirst(dimensions.first())) {
+        // "Normal" dimensions, that simply target variables in their references
+        const dim = dimensions.rule(state);
+        dim && element.dimensions.push(dim);
+      } else if (state.canConsumeFirst(dimensionsWithTypes.first())) {
+        // Dimensions that can also target types in their references
+        const dim = dimensionsWithTypes.rule(state);
+        dim && element.dimensions.push(dim);
+      }
```
msujew, line 6166: Suggestion: This requires reading the `canConsumeFirst` multiple times (up to four times if `dimensionWithTypes` matches). We can improve on this:
```suggestion
    let withoutType = false;
    let withType = false;
    while (
      (withoutType = state.canConsumeFirst(dimensions.first())) ||
      (withType = state.canConsumeFirst(dimensionsWithTypes.first()))
    ) {
      if (withoutType) {
        // "Normal" dimensions, that simply target variables in their references
        const dim = dimensions.rule(state);
        dim && element.dimensions.push(dim);
      } else if (withType) {
        // Dimensions that can also target types in their references
        const dim = dimensionsWithTypes.rule(state);
        dim && element.dimensions.push(dim);
      }
```

msujew (commented):
Looks really nice. I believe we can simplify some pieces now.

packages/language/src/language-server/signature-help-request.ts
Lotes: True. MemberCalls cannot contain Call Statements.

But MemberCalls and CallStatements can contain nested MemberCalls. Look at the new tests. Good that we found this one.

msujew (approved):
I see, that makes sense now 👍

PR #677, opened 2026-04-16
Add PL/I builtin documentation

Lotes:
Added documentation for all builtin procedures.


packages/language/src/syntax-tree/ast.ts
```diff
@@ -15,6 +15,23 @@ import { assertUnreachable } from "../utils/common";
 import { isObject } from "../utils/types";
 import type { CompilationUnit } from "../workspace/compilation-unit";
 
+export enum DataType {
```
Lotes, line 18: Revisit validation, that only builtin files can have ANY

packages/language/src/typesystem/descriptions.ts
```diff
@@ -1913,6 +1920,9 @@ export namespace TypeDescriptions {
       scanMode:
         attributes[AttributeKind.ScanMode]?.value ??
         DefaultValues[AttributeKind.ScanMode],
+      isDataTypeGeneric:
```
Lotes, line 1932: Respect this flag in assignability test.

packages/language/src/documentation/jsdoc.ts
Lotes, line 1: Is checkArguments working properly for builtins?

packages/language/src/typesystem/descriptions.ts
Lotes: Added a comment

Lotes:
@msujew The builtin files have some diagnostics. 2 parser errors that I cannot locate/identify. Also when opening the builtin file, the syntax highlighting is broken at some point. I would like to fix this, but currently I have no idea what is going wrong.

packages/language/src/parser/parser.ts
```diff
@@ -5178,7 +5179,9 @@ const returnsAttribute = rule(
       (state.canConsumeFirst(computationDataAttribute.first()) ||
         state.canConsumeFirst(dateAttribute.first()) ||
         state.canConsumeFirst(valueListAttribute.first()) ||
-        state.canConsumeFirst(valueRangeAttribute.first()))
+        state.canConsumeFirst(valueRangeAttribute.first()) ||
+        state.canConsumeFirst(anyAttribute.first()) ||
+        state.canConsumeFirst(anyAttribute.first()))
```
msujew: Suggestion: Duplicate
```suggestion
        state.canConsumeFirst(anyAttribute.first()))
```

packages/language/src/parser/parser.ts
```diff
@@ -5193,6 +5196,12 @@ const returnsAttribute = rule(
       } else if (state.canConsumeFirst(valueRangeAttribute.first())) {
         const attr = valueRangeAttribute.rule(state);
         attr && element.attrs.push(attr);
+      } else if (state.canConsumeFirst(anyAttribute.first())) {
+        const attr = anyAttribute.rule(state);
+        attr && element.attrs.push(attr);
+      } else if (state.canConsumeFirst(anyAttribute.first())) {
+        const attr = anyAttribute.rule(state);
+        attr && element.attrs.push(attr);
```
msujew: Suggestion: Duplicate
```suggestion
      } else if (state.canConsumeFirst(anyAttribute.first())) {
        const attr = anyAttribute.rule(state);
        attr && element.attrs.push(attr);
```

packages/language/src/typesystem/infer.ts
```diff
@@ -556,7 +564,19 @@ export class DefaultTypeInferer implements TypeInferer {
...
+      const literalValue = expression.value.value;
+      const match = pattern.exec(literalValue);
+      if (match) {
+        return match[1]
+          .slice(1, -1)
+          .replace(/("")/g, '"')
+          .replace(/('')/g, "'");
+      }
+      //if it does not match the pattern, it was produces by the PP parser, which removes quotes on its own
+      return literalValue;
```
msujew: Suggestion: This seems rather fragile. Can't we simply look for `"` and `'` at the start of the string and look for the last index of the respective closing `"` or `'` to find the content?

packages/language/src/parser/parser.ts
```diff
@@ -5206,6 +5215,43 @@ const returnsAttribute = rule(
...
+      CstNodeKind.AnyAttribute_ANY,
+      tokens.ANY,
+    );
+    if (
+      state.tryConsume(
+        element,
+        CstNodeKind.AnyAttribute_OpenAngle,
+        tokens.LessThan,
+      )
+    ) {
```
msujew, line 5229: Minor: I would try to keep this inline with other PLI attributes, which use `()` for additional data type options/arguments. Or is there a special technical reason for using `<>`?

msujew (commented):
Just tried it out, works really well! Thank you.

I just have just one more or less complicated problem: I'm not sure anymore about the convention of `{TYPE} [var]` for procedure parameters in the markdown docs.

1. It doesn't highlight it as code, which it pretty sure is.
2. Other programming languages usually do not show the type explicitly in the hover. E.g. TypeScript only does ``*@param* `var` - ...`` in its hover. Just the name of the parameter.

Since we can always infer the parameter signature from the definition of the procedure itself, I'm wondering whether we actually need to write `@param <def> <text>` at all, or whether `@param <text>` is enough, and the hover renders the correct parameter `<def>` based on the index. What do you think?

msujew:
@Lotes as for the parser errors - I couldn't reproduce those. The highlighting issues seem to stem from this:

image

There are a bunch of those in the builtin files.

msujew:
Note that there's also an error in the `MEMCONVERT` definition:

image

packages/language/src/typesystem/infer.ts
Lotes: I made this because of `(x[nu]?|a|e|b[43x]?|g[x]?|ux|wx|i|m)`.
I would prefer using the RegExp from the tokens, but that does not work so well.

packages/language/src/parser/parser.ts
Lotes: It would get confused with the dimensions otherwise. You could have an ANY<CHARACTER>(10) or ANY(CHARACTER)(10). It also looks weird in my eyes to have round parentheses twice.

packages/language/src/typesystem/infer.ts
msujew: Ok, but I don't see why we need the regexp in the first place. If you have a string like `"something"ux` you can simply look for `literalValue.charAt(0)` and search for the last index of that value within `literalValue` to get the `"something"` part. No complex regex matching necessary, since the suffix cannot contain the quote character.

Lotes:
I was able to fix all parsing errors.

What s left are linking errors for builtin type functions, which were not in my focus.

packages/language/src/parser/parser.ts
msujew: Oh, I misunderstood the grammar. Wouldn't it make more sense to have `ANY(CHARACTER(10))`? Since the `(10)` is more closely related to `CHARACTER` than it is to `ANY`.

Lotes:
> Since we can always infer the parameter signature from the definition of the procedure itself, I'm wondering whether we actually need to write `@param <def> <text>` at all, or whether `@param <text>` is enough, and the hover renders the correct parameter `<def>` based on the index. What do you think?

Puh, about the index: This is bad for now, because we have some overloaded functions which are solved via variadic parameters right now. There I introduced an artificial new parameter that summarizes all parameters (The overloaded functions are marked with a todo tag: MEMINDEX, REPATTERN, PLIPARSE).

The type definition is redundant, true. I thought it would be better to have it INSIDE the comment in order to not compute the type information. But we do infer the type nonetheless.
I think I could remove this using search and replace. I have no good other reason to keep it.

Lotes:
@msujew I overlooked the Type builtin functions (there are linking errors). I am not sure about these yet. They take types and not values. I think we should add them in a separate PR.

packages/language/src/parser/parser.ts
Lotes: Mh actually CHARACTER was not a good example, since it can be the string length or an array declaration. ANY<X>(N) means an array with arbitrary X in it.
Arbitrary shall mean: all other attributes are ignored, like the string length. But I think I accidently used this Array notation also as string length.

WDYT about this?

msujew: Ok, I see. In that case, that's alright with me 👍

Lotes:
> The type definition is redundant, true. I thought it would be better to have it INSIDE the comment in order to not compute the type information. But we do infer the type nonetheless. I think I could remove this using search and replace. I have no good other reason to keep it.

I removed the type information, it gets computed now.

msujew (approved):
Alright, thanks!

PR #678, opened 2026-04-23
Add CICS LINK statement

Lotes:
Added the first and most used CICS statement "LINK".

Every statement is made of multiple specifications like "PROGRAM(A)" or "DATALENGTH(B)". Here A and B are data values. Each data value rule has a certain type in mind (data VALUE seems to be a number, data AREA seems to be a string, names are also strings... let's see if this terminology works).

I grouped the rules into 3 sections: statements, specifications and data values (hopefully we will reuse all of them).

Another thing is the anyOrder helper function: For LINK it was not important in which order the specifications came up, so I asked for all of them in any order. Duplicates get a diagnostic, but the parsing will not be stopped.


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -598,6 +602,31 @@ function runSqlHostVariableInstruction(
...
+    startColumn,
+    // Start and end position are the same
+    startOffset,
+    startLine,
+    startColumn,
+    undefined,
+  );
+  // Emit both tokens now
+  context.tokens.push(varMarkerToken, token);
+}
```
msujew, line 614: Suggestion: Extract this into a common function? Like `emitSyntheticToken(context, token, type)`.

packages/language/src/parser/cics-parser.ts
```diff
@@ -14,20 +14,414 @@ import * as t from "./tokens";
...
+      const unknownStatement = ast.createEmbeddedUnknownStatement();
+      traverseAllNodes(cicsAst, (node) => {
+        if (node.kind === ast.SyntaxKind.CicsReferenceItem && node.ref) {
+          const token = node.ref.token;
+          token.immediateFollow = false;
+          unknownStatement.hostVariables.push(token);
+        }
+        return TraversalState.Continue;
+      });
+      execStatement.content = unknownStatement;
```
msujew: Question: This looks quite hacky. What is this supposed to do exactly? I would expect the `cicsAst` to be the content of the `execStatement`.

Lotes: The `token.immediateFollow = false;` was copied from the SQL equivalent.

The AST is thrown away, since we only need the variables so far. But I can readd it of course.

Lotes: Added the AST + variables.

packages/language/src/parser/cics-parser.ts
```diff
@@ -14,20 +14,414 @@ import * as t from "./tokens";
...
-    CstNodeKind.ExecCicsStatement_Semicolon,
-    t.Semicolon,
-  );
+  if (state.canConsumeFirst(cicsStatement.first())) {
+    const cicsAst = cicsStatement.rule(state);
+    if (cicsAst) {
+      traverseAllNodes(cicsAst, (node) => {
+        if (node.kind === ast.SyntaxKind.CicsReferenceItem && node.ref) {
+          const token = node.ref.token;
+          token.immediateFollow = false;
```
msujew: Suggestion: Setting `token.immediateFollow` to `false` should be done during the parsing of `cicsReferenceItem` already. We probably want to refactor this part eventually, and moving it to `cicsReferenceItem` itself is more maintainable.

msujew (approved):
Thanks, just one minor change, see below.

PR #681, opened 2026-04-28
Improve error messages for qualified names

### LINKED ISSUE #671: SQLCA and SQLDA
We had SQLCA and SQLDA built-in support  before implementing #391 .
This should not impact previous way of working. 

Below code should give correct IBM code 
e.g. `IBM1623I S       6.0  The SQLCODE is unknown`.  or 
 `IBM1623I S       6.0    The dot-qualified reference SQLCA.SQLCODE is unknown.` 
once hover-overed

Image

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/671

Simply adds additional checking for member calls/qualified name calls that improve the error message. Previously, we generated an error on each segment of a member call. Now we simply generate a single error for the whole reference.

### DIFF
--- a/packages/language/test/fourslash/linker/partial-declared-qualified.ts
+++ b/packages/language/test/fourslash/linker/partial-declared-qualified.ts
@@ -13,6 +13,9 @@
 
 // @wrap: main
 //// DCL A FIXED BIN;
+//// DCL 1 X, 2 Y FIXED BIN;
 //// <|A.B|> = 1;
+//// <|X.Y.Z.W|> = 1;
 
 verify.expectExclusiveDiagnosticsAt("A.B", code.Severe.IBM1623I);
+verify.expectExclusiveDiagnosticsAt("X.Y.Z.W", code.Severe.IBM1623I);
--- a/packages/language/test/fourslash/linker/partial-declared-qualified.ts
+++ b/packages/language/test/fourslash/linker/partial-declared-qualified.ts
@@ -12,6 +12,7 @@
 /// <reference path="../framework.ts" />
 
 // @wrap: main
+//// DCL A FIXED BIN;
 //// <|A.B|> = 1;
 
 verify.expectExclusiveDiagnosticsAt("A.B", code.Severe.IBM1623I);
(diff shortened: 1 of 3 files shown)


packages/language/test/fourslash/linker/partial-declared-qualified.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+// @wrap: main
+//// DCL A FIXED BIN;
+//// <|A.B|> = 1;
+
+verify.expectExclusiveDiagnosticsAt("A.B", code.Severe.IBM1623I);
```
Lotes, line 20: Is this also possible for a missing structure member? Such that you have 3 or more parts: `a1.a2.a3. ...`

msujew: I don't follow. Can you give a more extensive example?

Lotes: Maybe I misunderstood the code, but I was wondering why this linking error only goes 2 levels deep.

```
DCL 1 PERSON,
      2 NAME CHARACTER(10),
      2 AGE FIXED,
      2 IDs,
        3 PERSO_NUM FIXED,
        ...;
```

Try to link
* "PERSON.NO_MEMBER"
* or "PERSON.NAME.NO_MEMBER"
* or "PERSON.IDs.NO_MEMBER"
* or "PERSON.IDs.PERSO_NUM.NO_MEMBER"

I was thinking this PR is about these FQNs, too.

msujew: It does work for more than just 2 levels. I can adjust the test for this case, but the test is supposed to show that even if `A` is defined, the full `A.B` reference receives an error. Previously, the language server only flagged the `B` reference part.

Lotes: Oki Doki! Thanks for adjusting the test. Good that it works.

Lotes (approved):
LGTM, thank you.

PR #683, opened 2026-04-28
Add documentation for builtin type functions

### LINKED ISSUE #679: Add documentation for builtin type functions
There is a handful of type functions that are not covered yet. Since they are handled differently, let's handle them in a extra task/PR.

We need the hover request and the signature help.

@msujew I overlooked the Type builtin functions (there are linking errors). I am not sure about these yet. They take types and not values. I think we should add them in a separate PR.

_Originally posted by @Lotes in https://github.com/zowe/zowe-pli-language-support/issues/677#issuecomment-4326876580_

Lotes:
Closes #679

### DIFF
--- a/code_samples/plugin-example/builtins.pli
+++ b/code_samples/plugin-example/builtins.pli
@@ -68,6 +68,9 @@
 
  CALL QUICKSORT(BITS, 1, 30);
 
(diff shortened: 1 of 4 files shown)


msujew (approved):
Great 👍

PR #685, opened 2026-04-28
Bugfix: Crashes on type system

### LINKED ISSUE #682: ls crashed on certain files
Once we open below file language server crashes, sample file is provided

Lotes:
Closes #682

I still get an "Maximum call stack size exceeded.", but for a different problem around a 1M LOC file on my machine.


packages/language/src/typesystem/composite-type-builder.ts
```diff
@@ -97,6 +97,16 @@ export class DefaultCompositeTypeBuilder implements CompositeTypeBuilder {
...
+    //required for "generic" parameters only coming from builtin functions
+    const isDataTypeGeneric =
+      attributes.witnesses.witnesses[AttributeKind.DataTypeIsGeneric]?.value ??
+      false;
+    if (isDataTypeGeneric) {
+      const dataType =
+        attributes.witnesses.witnesses[AttributeKind.DataType]?.value ??
+        DataType.Unknown;
+      return dataType === DataType.Structure || dataType === DataType.Union;
+    }
```
msujew, line 107: Suggestion: I don't believe those null-coalescing operators are required here, due to the thruthyness check and direct comparisons.
```suggestion
    const isDataTypeGeneric = attributes.witnesses.witnesses[AttributeKind.DataTypeIsGeneric]?.value;
    if (isDataTypeGeneric) {
      const dataType = attributes.witnesses.witnesses[AttributeKind.DataType]?.value;
      return dataType === DataType.Structure || dataType === DataType.Union;
    }
```

msujew (approved):
Looks good to me 👍

PR #688, opened 2026-04-29
Refactor plugin config management and enable remote include resolution

### LINKED ISSUE #465: Improve Post-Processing of Process Groups
Stemming from #455.

See [`postProcessProcessGroups` in **plugin-configuration-provider.ts**](https://github.com/zowe/zowe-pli-language-support/blob/development/packages/language/src/workspace/plugin-configuration-provider.ts#L425-L429).  We're currently awaiting each stat check sequentially on all directory entries that are returned, which can become a problem when we have a large number of checks to perform. We should avoid blocking on these checks, and compute them in parallel instead. That, or we should consider exposing something like `withFileTypes` in our readDir implementation.

### LINKED ISSUE #629: Remote resolution of include members (copybooks)
Depend on task completion (https://github.com/zowe/zowe-pli-language-support/issues/668)  

Currently include files are resolved locally (in workspace) and external (in disk drive). 
We should provide another mechanism which resolves include members remotely if system can not 
find it in workspace or disk. 

Suggested solution is integration a Zowe explorer as like in HLASM - has demoed to team already.
Inlude members will be resolved based on URI.  
"zowe-ds:/activezoweprofile/dataset-qualified-name"  "
(e.g. AZAT.PL1.APP.DBRMLIB is a dataset qualified name )

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/629
Closes https://github.com/zowe/zowe-pli-language-support/issues/465

Refactors major parts of the plugin config management - A lot of it has grown too large to manage well. The PR splits off a lot of the parts of the plugin config management and separates them into its own modules. Still should hopefully make this part of the code easier to improve in the future, for example, when we want to show diagnostics of compiler options within the plugin config json files.

Also adjusts some other parts to enable resolution of non-file URIs. This can be tested by using #629, for which access to a mainframe via the zowe-explorer extension is required.

Adds some additional UX improvements that display what the language server is doing. This is active, when the language server takes a longer-than-usual (500ms) time to process the plugin config files. Can be simulated by adding a timeout into the code.


packages/language/src/config/compiler-options-merge.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+  if (groupOptions) {
+    for (const option of groupOptions) {
+      const parsed = parseAbstractCompilerOptions(option.value);
+      abstractOptions.options.push(...parsed.options);
+      abstractOptions.tokens.push(...parsed.tokens);
+      abstractOptions.issues.push(...parsed.issues);
+    }
+  }
+  return { abstractOptions, issues: abstractOptions.issues };
```
ssmifi, line 55: _Question:_ Should we continue to provide the explicit `issue` redundancy? I suppose it makes sense in the `ProgramConfig` in case there are additional issues apart from the compiler options, but in this case I'm unsure. If we want to do this pattern more commonly, we should maybe use a dedicated interface. Wdyt?

packages/language/src/config/lib-expander.ts
```diff
@@ -0,0 +1,321 @@
...
+  while (queue.length > 0) {
+    const { lib, uri } = queue.shift()!;
+    const dirEntries = (await safeReadDir(fs, uri)) ?? [];
+
+    const files = new Map<string, string>();
+    const datasetMembers = new Map<string, string>();
+    for (const [name, fileType] of dirEntries) {
+      if (fileType & FileType.Directory) {
+        queue.push({
+          lib: `${lib}/${name}`,
```
ssmifi, line 226: _Question_: In general, does this work for all file systems? Could there be libs with windows-style backslashes or do we make sure that this is already normalized? Same with directory depth count.

packages/language/src/config/lib-expander.ts
```diff
@@ -0,0 +1,321 @@
...
+          uri: UriUtils.joinPath(uri, name),
+        });
+        continue;
+      }
+      if (!(fileType & FileType.File)) {
+        continue;
+      }
+      const lowerName = name.toLowerCase();
```
ssmifi, line 234: _Question_: Does this hold for case-sensitive file systems?

packages/language/src/config/lib-expander.ts
```diff
@@ -0,0 +1,321 @@
...
+    const datasetMembers = new Map<string, string>();
+    for (const [name, fileType] of dirEntries) {
+      if (fileType & FileType.Directory) {
+        queue.push({
+          lib: `${lib}/${name}`,
+          uri: UriUtils.joinPath(uri, name),
+        });
+        continue;
+      }
+      if (!(fileType & FileType.File)) {
```
ssmifi, line 231: _Question_: Probably a minor question... I was just wondering if this also holds at symbolic links, e.g., to directories. Probably an edge case, but maybe we should document an expectation here.

packages/language/src/config/loader.ts
```diff
@@ -0,0 +1,381 @@
...
+    return { config: undefined, diagnostics };
+  }
+
+  const configs: ProgramConfig[] = [];
+  const children = pgmsNode.children ?? [];
+  for (let i = 0; i < children.length; i++) {
+    const entry = readProgramConfig(children[i], i, configUri);
+    if (entry) {
+      configs.push(entry);
+    }
```
ssmifi, line 93: _Question_: Should we return some kind of diagnostics/error if the entry returns undefined, e.g., due to malformed or missing required fields?

packages/language/src/config/loader.ts
```diff
@@ -0,0 +1,381 @@
...
+  uri: URI,
+): ProgramConfig | undefined {
+  if (!node || node.type !== "object") {
+    return undefined;
+  }
+  const path: JSONPath = ["pgms", index];
+  const program = readStringField(node, "program", [...path, "program"], uri);
+  const pgroup = readStringField(node, "pgroup", [...path, "pgroup"], uri);
+  if (!program || !pgroup) {
+    return undefined;
```
ssmifi, line 151: _Suggestion_: Report what is missing.

packages/language/src/config/loader.ts
```diff
@@ -0,0 +1,381 @@
...
+
+function readStringField(
+  parent: JsonNode | undefined,
+  key: string,
+  path: JSONPath,
+  uri: URI,
+): JsonItem<string> | undefined {
+  const value = findProperty(parent, key);
+  if (!value || value.type !== "string") {
+    return undefined;
```
ssmifi, line 257: _Question/Suggestion_: In general, all these read functions silently return undef if the type is wrong. Should we report this to the user?

packages/language/src/config/path-resolver.ts
```diff
@@ -0,0 +1,53 @@
...
+  scheme?: string,
+): URI {
+  const pathType = UriUtils.computePathType(lib);
+  if (pathType === UriUtils.PathType.URI) {
+    return UriUtils.parse(lib);
+  }
+  if (pathType === UriUtils.PathType.Absolute) {
+    return UriUtils.file(lib).with({ scheme: scheme ?? workspace.scheme });
+  }
+  return UriUtils.joinPath(workspace, lib);
```
ssmifi, line 38: _Question_: I think this is intentional, but to make sure: Is it ok to escape the workspace path, e.g., with a relative path?

packages/language/src/config/loader.ts
```diff
@@ -0,0 +1,381 @@
...
+  );
+}
+
+/**
+ * Internal-to-loader Diagnostic -> LspDiagnostic converter. The provider
+ * exposes its own conversion for runtime-discovered diagnostics (e.g. from
+ * lib expansion); this helper handles diagnostics produced inside the
+ * pure parse step.
+ */
+export function toLspDiagnostic(
```
ssmifi, line 362: _Question_: Looks quite similiar to `diagnosticToLSP` in `types.ts`. Should we provide a single more generic function here or would you like to keep it separated? Would also be fine, I think.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -103,8 +100,10 @@ export async function quickFixResolveInclude(
     );
     return;
   }
+  // newFileContent comes from raw JSON.parse so groups carry plain string
+  // names, not the JsonItem-wrapped form of the loaded ProcessGroup schema.
   const groupToUpdate = newFileContent.pgroups.find(
-    (g: ProcessGroup) => g.name === progConfig.pgroup,
+    (g: { name: string }) => g.name === progConfig.pgroup.value,
```
ssmifi, line 108: _Question_: Could the `parentFolder` already be included in `groupToUpdate`?

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -33,13 +29,16 @@ const JSONC_FORMAT = {
   formattingOptions: { tabSize: 2, insertSpaces: true },
```
ssmifi, line 29: _Minor_: Could also use these settings here in the JSON stringify at Line 112.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
```diff
@@ -433,19 +435,19 @@ export async function applyQuickFixes(
...
       case CODE_MACRO_CASE:
         action = quickFixUppercaseText(diagnostic);
         if (action) actions.push(action);
         break;
       case CODE_UNRESOLVED_LIB:
-        action = await quickFixRemoveUnresolvedLib(diagnostic);
+        action = await quickFixRemoveUnresolvedLib(diagnostic, workspace);
         if (action) actions.push(action);
         break;
     }
```
ssmifi, line 463: _Suggestion_: Unrelated to this PR, personally, I find `quickFixResolveAmbiguousReference` and maybe also `quickFixRemoveAllUnresolvedLibs` a bit hard to follow. Maybe we should refactor this a bit in the future for easier maintainability.

packages/language/src/language-server/commands.ts
```diff
@@ -9,22 +9,26 @@
...
  */
 
-import { FileSystemProviderInstance } from "../workspace/file-system-provider";
 import { ExecuteCommandParams } from "vscode-languageserver";
 import { updateOrCreateConfig } from "../utils/config";
 import { UriUtils } from "../utils/uri";
+import { WorkspaceContext } from "../workspace/workspace-context";
 
-export async function commandResolveInclude(params: ExecuteCommandParams) {
+export async function commandResolveInclude(
```
ssmifi, line 17: _Suggestion_: For general code maintenance, all these functions could use a line of doc and a proper return definition.

packages/language/src/language-server/commands.ts
```diff
@@ -9,22 +9,26 @@
...
+  params: ExecuteCommandParams,
+  workspace: WorkspaceContext,
+) {
   const [uri, content] = params.arguments as string[];
```
ssmifi, line 21: _Suggestion_: This is probably currently not an issue with our clients, but in general the `arguments` should be checked properly. Otherwise, incomplete data might crash the handler.

packages/language/src/preprocessor/include-resolver.ts
```diff
@@ -0,0 +1,446 @@
...
+ * Looks up `name` (and optionally `name + ext` for each extension) in a
+ * pre-built file index. Returns the on-disk filename of the first match,
+ * preserving its casing, or undefined.
+ */
+function findIndexedFile(
+  files: Map<string, string>,
+  name: string,
+  extensions: readonly string[],
+): string | undefined {
+  const lower = name.toLowerCase();
```
ssmifi, line 192: _Suggestion/Question_: Connected to the `lib-expander` key question. Perhaps we should provide a function that creates the keys for file maps so that we can change it at one place (in case it has to be changed because of the fs)?

packages/language/src/preprocessor/include-resolver.ts
```diff
@@ -0,0 +1,446 @@
...
+ * basename needs to match the include's `ddname`.
+ */
+function tryResolveDDNameMember(
+  lib: LibsDDEntry,
+  ddname: string,
+  memberName: string,
+  workspace: URI,
+  scheme: string,
+): URI | undefined {
+  if (!lib.path.toLowerCase().endsWith(ddname.toLowerCase())) {
```
ssmifi: _Question_: Is this supposed to match complete member names? This would also match something like `lib.path = "cpy/ABCDEFG"` with `ddname = "DEFG"`.

packages/language/src/utils/config.ts
```diff
@@ -74,15 +72,14 @@ export async function updateOrCreateConfig(programPath: string): Promise<void> {
...
+      program: plainItem(programPath),
+      pgroup: plainItem("default"),
+      compilerOptions: [],
     });
 
     parsedTextContent.pgms.push({ program: programPath, pgroup: "default" });
-    await PluginConfigurationProviderInstance.writeProgramConfigFile(
-      parsedTextContent,
-    );
+    await config.writeProgramConfigFile(parsedTextContent);
```
ssmifi: _Minor_: Theoretically, the change in memory can happen even if the write to disk was unsuccessful, which would create an inconsistency. Perhaps the order here should be swapped.

packages/language/src/utils/messages.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/**
+ * Notification sent to the LS when the workspace's plugin configuration changes.
+ */
+export const WorkspaceDidChangePlipluginConfigNotification =
+  "workspace/didChangePlipluginConfig";
```
ssmifi: _Tiny_: I would probably use a capital `P` for `Plugin`.

packages/language/src/utils/messages.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const ExistingFileRequest = "pli/existingFileRequest";
+
+export const UpdateOperation = "pli/updateOperation";
```
ssmifi: _Minor_: Since `WorkspaceDidChangePlipluginConfigNotification` has a doc, I would provide one here, too. `ExistingFileRequest` is maybe self-explanatory, but `UpdateOperation` might be a bit vague.

packages/language/src/utils/messages.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
```
ssmifi, line 10: _Suggestion_: We could/should add the payload defintions here, too. IISC, `UpdateOperation` carries the `title` field.

packages/language/src/utils/messages.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
```
ssmifi: _Minor Suggestion_: We could use a namespace `Messages` here, which might increase readability at call sights, but I'm also fine with keeping the amount of namespaces low.

packages/language/src/utils/promises.ts
```diff
@@ -119,3 +121,29 @@ export class Deferred<T = void> {
...
+  title: string,
+  timeout = 500,
+): Cancellation {
+  if (!connection) {
+    // If no connection is available, we can't send progress updates, so we return a no-op cancellation function.
+    return () => {};
+  }
+  let running = false;
+  const startTimeout = setTimeout(() => {
+    connection?.sendNotification(UpdateOperation, { title });
```
ssmifi: _Minor_: `connection` is set, correct?

packages/language/src/utils/uri.ts
```diff
@@ -34,6 +34,29 @@ export namespace UriUtils {
     path.charAt(2) === "/" &&
```
ssmifi: _Question_: What about \ in windows paths? (It is considered in other functions here...)

packages/language/src/utils/uri.ts
```diff
@@ -34,6 +34,29 @@ export namespace UriUtils {
...
+    Absolute,
+    Relative,
+    URI,
+  }
+
+  export function computePathType(path: string): PathType {
+    if (
+      isWindowsAbsolutePath(path) ||
+      isUnixAbsolutePath(path) ||
+      isHomePath(path)
```
ssmifi, line 48: _Question_: What is the rationale behind the classification of the home dir? Is this important for other mechanisms downstream? These never get expanded and would result in something lile `file://~/documents` by the UriUtils, wouldn't they?

packages/language/src/workspace/file-system-provider.ts
```diff
@@ -232,64 +215,28 @@ export class VirtualFileSystemProvider implements FileSystemProvider {
   }
```
ssmifi, line 219: Maybe not practically relevant right now, but btw, `deleteFile` looks unsafe. `indexOf` returns -1 if the key is invalid, which would result in `splice(-1, 1)` which would remove the last elememt, wouldn't it?

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -1109,14 +712,14 @@ export class PluginConfigurationProvider {
     workspacePath: URI,
     programConfig: ProgramConfig,
   ) {
-    const existing = this.programConfigs.get(programConfig.program);
+    const existing = this.programConfigs.get(programConfig.program.value);
```
ssmifi: _Question_: Is this correct? Aren't the keys the resolved URIs, but the `program.value` is the raw entry?

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -278,26 +149,31 @@ type ProcGrpsSnapshot = {
  */
 export class PluginConfigurationProvider {
   /**
-   * Prebuilt list of glob patterns for library file matching.
+   * Direct prefix-and-extension index for library file membership checks.
+   * Maps a lower-cased URI prefix (`<workspace>/<libDir>/`) to the set of
+   * lower-cased extensions allowed in that lib. A file is a lib candidate
+   * iff its parent directory matches a prefix and its extension matches the
+   * prefix's set. Replaces the previous minimatch-based glob patterns —
+   * cheaper per call and avoids depending on minimatch for this hot path.
```
ssmifi: _Question_: However, `minimatch` is still needed/included. Could/Should we get rid of it entirely?

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -829,91 +628,21 @@ export class PluginConfigurationProvider {
...
     );
-  }
-
-  private createConfigTextDocument(
-    fileName: string,
-    text: string,
-  ): TextDocument {
-    const uri = this.getConfigUri(fileName);
-    return TextDocument.create(uri, "jsonc", 0, text);
+    this.postProcessProgramConfigs();
```
ssmifi, line 642: _Minor_: `setProcessGroupConfig` does also call `postProcessProgramConfig`. Probably idempotent, but wasteful?

packages/language/test/test-workspace.ts
```diff
@@ -0,0 +1,60 @@
...
+} from "../src/workspace/file-system-provider";
+import { WorkspaceContext } from "../src/workspace/workspace-context";
+
+let _defaultTestWorkspace: WorkspaceContext | undefined;
+
+/**
+ * Workspace context returned by the test parse helpers when no workspace is
+ * threaded through explicitly. Production code never reaches for this —
+ * `WorkspaceContext` is always passed in. For tests, having a single
+ * mutable default keeps fixtures concise.
```
ssmifi, line 26: _Questions_: Is this actually safe in parallel test execution?

packages/vscode-extension/src/extension/progress.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function registerProgressReporter(client: BaseLanguageClient): void {
+  let dispose = new vscode.Disposable(() => {});
+  client.onNotification(UpdateOperation, (operation) => {
+    if (operation && typeof operation.title === "string") {
+      const title = operation.title;
+      if (title === "") {
+        dispose.dispose();
+      } else {
+        dispose = vscode.window.setStatusBarMessage("$(loading~spin) " + title);
```
ssmifi: What happens to the potentially overwritten message in `dispose`? Should that be disposed before explicitly?

packages/vscode-extension/src/extension/main.ts
```diff
@@ -167,8 +169,10 @@ async function startLanguageClient(
     clientOptions,
   );
 
-  // Register custom decorator types.
+  // Register custom connection message handlers.
   registerCustomDecorators(client, settings);
+  registerFileSystemProvider(client);
+  registerProgressReporter(client);
```
ssmifi: _Minor_: The handlers do not create disposables. Shouldn't they in case the extension is reloaded or deactivated?

packages/vscode-extension/src/extension/file-system-provider.ts
```diff
@@ -0,0 +1,74 @@
...
+  );
+  client.onRequest(Messages.Stat, async (uriString: string) => {
+    const uri = vscode.Uri.parse(uriString);
+    const stat = await vscode.workspace.fs.stat(uri);
+    return {
+      isFile: stat.type === vscode.FileType.File,
+      isDirectory: stat.type === vscode.FileType.Directory,
+    };
+  });
+  client.onRequest(Messages.WriteFile, async (uriString, value) => {
```
ssmifi: Isn't this a mismatch? In `language/file-system.ts`, the `sendRequest` in `writeFile` expects one object `{ uri, value }`.

packages/vscode-extension/src/language/file-system.ts
```diff
@@ -62,21 +51,18 @@ export class VSCodeFileSystemProvider implements FileSystemProvider {
...
     if (result) {
-      if (isPathSearch(options)) {
-        return options.path.with({ path: result });
-      } else {
-        // not a path, take result as full URI
-        return UriUtils.toUri(result);
-      }
-    } else {
-      return undefined;
+      return path.with({ path: result });
```
ssmifi, line 63: _Minor_: Should we add a comment that only the path is returned anyway and that the other URI fields remain untouched?

packages/vscode-extension/src/language/main.ts
```diff
@@ -15,39 +15,74 @@ import {
...
+// This provider is used when running the language server in a Node environment
+// Usually preserved for VS Code desktop use cases
+// It optimizes file system access by directly using Node's fs module,
+// rather than sending requests to the client as the browser provider does.
+//
+// For non-file URIs, it falls back to the VSCodeFileSystemProvider which sends requests to the client.
+class NodeFileSystemProvider extends VSCodeFileSystemProvider {
+  override readFile(uri: URI): Promise<string | undefined> {
+    if (uri.scheme !== "file") {
+      return super.readFile(uri);
```
ssmifi, line 38: _Minor_: Inconsistency: The super readFile returns undef on errors, the fs readFile throws an error.

packages/vscode-extension/src/language/main.ts
```diff
@@ -56,68 +91,48 @@ class NodeFileSystemProvider implements FileSystemProvider {
...
-      return undefined;
-    }
-    return result;
+    return result ?? undefined;
   }
 
   /**
    * Stats for file or directory
    * @param uri URI of file or directory to stat
    * @returns Stats object. If stat fails, returns a stats object with both isFile and isDirectory false.
```
ssmifi: _Minor_: This was the behavior before, but now it throws an exception if `stat()` fails.

packages/vscode-extension/src/language/main.ts
```diff
@@ -56,68 +91,48 @@ class NodeFileSystemProvider implements FileSystemProvider {
...
+    });
     if (!files.length) {
       return undefined;
     }
     const result = UriUtils.toUri(files[0]);
-    if (!result) {
-      return undefined;
-    }
-    return result;
+    return result ?? undefined;
```
ssmifi: _Minor_: This is redundant, isn't it?

packages/vscode-extension/src/language/main.ts
```diff
@@ -56,68 +91,48 @@ class NodeFileSystemProvider implements FileSystemProvider {
     }
     await fs.promises.writeFile(uri.fsPath, value, "utf8");
```
ssmifi: _Question_: This looks a bit odd. The file should be written regardless of whether the directory could be created?

ssmifi (commented):
Thank you very much for the extensive refactoring. It’s great to see us straightening out the whole filesystem topic. Apologies for all the questions this time, but the PR and the topic itself are relatively extensive.

packages/language/src/config/loader.ts
msujew: The user should already see a diagnostic on the file, because it no longer matches the JSON schema, right?

packages/language/src/config/loader.ts
msujew: I believe this should be covered by the JSON schema already - I wouldn't try to solve this in this PR anyway, it's out of scope.

packages/language/src/config/path-resolver.ts
msujew: I don't follow, can you give an example?

packages/language/src/config/loader.ts
msujew: We can eventually choose to better integrate the config loading into the compilation unit lifecycle - this function already existed beforehand, just in the `plugin-configuration-provider.ts` file. Out of scope for this PR.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
msujew: The `procGrpsConfig.computedLibsSet.has(parentFolder)` check beforehand should cover that AFAIK.

packages/language/src/language-server/code-actions/apply-quick-fixes.ts
msujew: Yes, they are a bit all over the place - with the new infrastructure, they should likely be easier to refactor.

packages/language/src/preprocessor/include-resolver.ts
msujew: Good point - I'll add a test for this (we don't have any yet for this code path).

packages/language/src/utils/messages.ts
msujew: Out of scope for this PR. I've created https://github.com/zowe/zowe-pli-language-support/issues/690 to track this.

packages/language/src/utils/promises.ts
msujew: Yes, but the type checker complains anyway, because the flow analysis "restarts" when entering a new lambda block.

packages/language/src/utils/uri.ts
msujew: I'm not sure either - it was there before I started moving things around. I'm not sure how well this works at all, given that I only have a Windows PC right now for testing.

packages/language/src/workspace/plugin-configuration-provider.ts
msujew: Right, this was here before already. I'll remove that code though - I believe it's dead code.

packages/language/src/workspace/plugin-configuration-provider.ts
msujew: It's still needed for glob-like Patterns. I'll remove that comment part.

packages/language/test/test-workspace.ts
msujew: Not sure, but AFAIK we don't have parallel tests, and so far, we probably don't need it.

packages/vscode-extension/src/extension/file-system-provider.ts
msujew: Good catch!

packages/language/src/config/path-resolver.ts
ssmifi: The question was if it should be possible to escape the workspace by having a relative path like `../../someFile`.

msujew: Yes, that's an explicit requirement :)

packages/language/src/config/compiler-options-merge.ts
msujew: I believe this was there before already, out of scope for now.

packages/language/src/config/lib-expander.ts
msujew: I've added additional tests for this 👍

packages/language/src/config/lib-expander.ts
msujew: Adjusted the virtual file system to be (optionally) case sensitive. Testable using the `// @case-sensitive: true` annotation. Added a test for this.

msujew:
@ssmifi I believe I was able to address most of your points. I've added a bunch of tests for some of the more special behavior (some were before my PR untested). There are a few things that I'd say are out of scope for this PR.

packages/language/src/utils/uri.ts
ssmifi: @Wagner-Laranjeiras Can you add this question to your current stress test?

.claude/settings.local.json
```diff
@@ -0,0 +1,15 @@
+{
+  "permissions": {
```
ssmifi: Probably committed by accident?

ssmifi (commented):
Thanks for the clarifications and fixes. Looks good to me, but I think you probably checked in a .claude file by accident. Can you please check?

.claude/settings.local.json
msujew: Ups, yeah, I'll add a github ignore rule for claude files :)

packages/language/src/config/lib-expander.ts
msujew: If the fs identifies it as a symbolic link, then it will try to resolve that to a file or directory. Meaning that this should work as expected.

ssmifi (approved):
Thanks!

PR #689, opened 2026-04-29
Polish sysparm, system, lp evaluation

### LINKED ISSUE #633: Read config value and inject into preprocessor statements
Resolve SYSPARM, LP(x),  SYSTEM(x) etc. compiler options and provide its value to macros, sample code for all use cases provided internally.

ssmifi:
Fixes https://github.com/zowe/zowe-pli-language-support/issues/633

The builtins functionality was already implemented, however there were some minor issues that prevented the provided code to pass correctly, which should be fixed in this update:

- Subtractions were evaluated right-to-left
- Changes to the plugin config (compiler options) did not automatically update all compilation units (caches)
- Set LP default value
- Enabled the existing test case and added new ones for #633 and binary expressions
- Added node types to the tsconfig to eliminate the `console` errors in the IDE.

The test code was shared internally.

### DIFF
--- a/packages/language/src/workspace/compilation-unit.ts
+++ b/packages/language/src/workspace/compilation-unit.ts
@@ -403,8 +403,10 @@ export class CompilationUnitHandler {
       }
       const allDiagnostics = diagnosticsToLSP(unit, unit.diagnostics.getAll());
       for (const file of unit.services.files.keys()) {
-        if (!EditorDocuments.get(file) || isVirtualFile(file)) {
-          // do not report diagnostics for virtual files or files not currently open in the editor
+        // Check synchronously if the file is currently open in the editor.
+        // Use has() not get() - we don't want to load from disk, just check if already open.
+        // Do not report diagnostics for virtual files or files not currently open in the editor.
+        if (!EditorDocuments.has(file) || isVirtualFile(file)) {
           continue;
         }
         const fileDiagnostics = allDiagnostics.get(file);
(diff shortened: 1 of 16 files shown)


packages/language/src/workspace/compilation-unit.ts
```diff
@@ -440,7 +442,16 @@ export class CompilationUnitHandler {
...
             if (textDocument) {
-              this.process(unit, textDocument, connection, cancellationToken);
+              await this.process(
+                unit,
+                textDocument,
+                connection,
+                cancellationToken,
+              );
+              // Clear include cache to force re-resolution with updated config
+              unit.services.includeCache.clear();
```
msujew: Question: Why clear this after the `process` call? Does this do anything?

msujew (commented):
Looks good. Just one question, see below.

packages/language/src/workspace/compilation-unit.ts
ssmifi: I added both lines so that `reindex` behaves the same as `updateUri` in this regard, but you are right: Regarding the provided test code, it is enough to revalidate the lsp caches. So, you would be more comfortable if this got removed, yes?

msujew: Yes. I don't think clearing the cache here has any meaningful benefit 👍

msujew (approved):

Issue #382, opened 2025-09-02
[Feature] Support RULES compiler option


ssmifi:
The RULES compiler option includes many toggles for validation (see pg, page 70ff).  All/Most of the validations should be considered.

PR #691, opened 2026-05-06
Add RULES(MULTICLOSE) implementation

### LINKED ISSUE #559: Support for RULES(MULTICLOSE) and Better Support for RULES(NOMULTICLOSE)
With the compiler option `RULES(MULTICLOSE)` multiple groups can be closed with a single `END entry-label` statement and it is not a syntax error.

With the compiler option `RULES(NOMULTICLOSE)` the `END entry-label` statement should terminate a procedure and
- report errors about unclosed groups within the procedure
- NOT report errors outside of the procedure (which it currently does, probably because instead of closing the procedure it closes the last opened group)

### LINKED ISSUE #673: Closing end-statements appropriately
Currently when we forget end for the procedures, and put some valid code after it, error is shown on different "END" statements. 
Let's enhance and improve multiple close statements.  Source code is shared already. Below is  just a snippet. 
```

 SQLMSGP: PROC;
 
 exec sql include sqlca;
  /* CHECK BY SQLCODE IF SQL QUERY RUN SUCCESSFUL OR NOT       */
 IF SQLCODE < 0 THEN  
 DO;
    PUT LIST('ERROR MESSAGE WAS RECEIVED');
    PUT LIST(SQLCA.SQLCODE);                               /*SQLCA*/      
    PUT SKIP;

    CALL SQLMSGP;

    STOP;
 END;

```
 
Image

ssmifi:
Fixes https://github.com/zowe/zowe-pli-language-support/issues/559
Closes https://github.com/zowe/zowe-pli-language-support/issues/673
and part of https://github.com/zowe/zowe-pli-language-support/issues/382

Implements the RULES(MULTICLOSE) feature. Basically, the logic is:
- With RULES(MULTICLOSE) (or alwaysOptional=true):
  - Unlabeled END: Consumed by first (innermost) compound statement that sees it
  - Labeled END: Only consumed if label matches one of the current statement's labels
  - If END exists but doesn't match, it's left for outer scope
  - However, at least one matching END is required
- With RULES(NOMULTICLOSE) (and alwaysOptional=false):
  - END is required - parsing error reported if missing
- END at `package` is always optional

Also extends the `test-builder` to accept regexes at diagnostic messages and refactored the code a bit to remove duplicated code.

Two additional observations: 
1. The mainframe compiler actually reports a warning `IBM1120I` when a multiclose is applied. This is not part of this PR, but will be implemented in a followup.
2. The `test-builder` grew quite a lot over the past year. Maybe we should put some refactoring work in it to improve long-term maintainability.

### DIFF
--- a/packages/language/src/parser/parser-lookahead.ts
+++ b/packages/language/src/parser/parser-lookahead.ts
@@ -123,37 +123,60 @@ function performIfLookahead(state: ParserState): boolean {
   return false;
 }
 
+/**
+ * Performs lookahead to detect END statement and its optional label.
+ * Skips any label prefixes (e.g., "FOO: END") before the END keyword.
+ * 
+ * @returns 
+ *   - `false` if no END statement found
+ *   - `true` if unlabeled END found (e.g., "END;")
+ *   - `string` if labeled END found (e.g., "END FOO;" returns "FOO")
+ */
 export function performEndStatementLookahead(
   state: ParserState,
-): { hasEnd: true; label: string | null } | { hasEnd: false } {
+): boolean | string {
   const lookahead = (la: number) => state.peek(la);
   let index: number = 1;
   let token: tokens.Token | undefined = undefined;
-
+  
   while ((token = lookahead(index)) && !tokenMatcher(token, tokens.END)) {
     const idToken = lookahead(index);
     const colonToken = lookahead(index + 1);
     if (!idToken || !colonToken) {
-      return { hasEnd: false };
+      return false;
     }
     if (
       !tokenMatcher(idToken, tokens.ID) ||
       !tokenMatcher(colonToken, tokens.Colon)
     ) {
-      return { hasEnd: false };
+      return false;
     }
     index += 2;
   }
-
+  
   if (!token || !tokenMatcher(token, tokens.END)) {
-    return { hasEnd: false };
+    return false;
   }
-
-  // Check for label reference after END
-  const labelToken = lookahead(index + 1);
-  if (labelToken && tokenMatcher(labelToken, tokens.ID)) {
-    return { hasEnd: true, label: labelToken.image };
-  } else {
-    return { hasEnd: true, label: null };
+  
+  // To verify that the end is actually an END statement and not part of an expression,
+  // we check for the semicolon, potentially preceded by an ID label.
+  let nextToken = lookahead(index + 1);
+  if (!nextToken) {
+    return false; 
   }
+  
+  if (tokenMatcher(nextToken, tokens.Semicolon)) {
+    return true;
+  }
+  
+  if (!tokenMatcher(nextToken, tokens.ID)) {
+    return false;
+  }
+  
+  const semicolonToken = lookahead(index + 2);
+  if (!semicolonToken || !tokenMatcher(semicolonToken, tokens.Semicolon)) {
+    return false; 
+  }
+  
+  return nextToken.image;
 }
--- a/packages/language/src/parser/parser-lookahead.ts
+++ b/packages/language/src/parser/parser-lookahead.ts
@@ -124,59 +124,81 @@ function performIfLookahead(state: ParserState): boolean {
 }
 
 /**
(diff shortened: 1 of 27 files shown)


packages/language/src/parser/parser-lookahead.ts
```diff
@@ -123,23 +123,37 @@ function performIfLookahead(state: ParserState): boolean {
   return false;
 }
 
-export function performEndStatementLookahead(state: ParserState): boolean {
+export function performEndStatementLookahead(
+  state: ParserState,
+): { hasEnd: true; label: string | null } | { hasEnd: false } {
```
msujew: Suggestion: I believe we can simplify the type here to be just `boolean | string`. Maybe add a comment to explain what this means.

packages/language/src/parser/parser-lookahead.ts
```diff
@@ -123,23 +123,37 @@ function performIfLookahead(state: ParserState): boolean {
...
+    return { hasEnd: false };
+  }
+
+  // Check for label reference after END
+  const labelToken = lookahead(index + 1);
+  if (labelToken && tokenMatcher(labelToken, tokens.ID)) {
+    return { hasEnd: true, label: labelToken.image };
+  } else {
+    return { hasEnd: true, label: null };
+  }
```
msujew: Suggestion: We need an additional check here, that the next token is actually a semicolon. Please add a test for the following - it should currently fail on your branch:

```
TEST: PROC;
  DCL END CHAR;
  END = "HELLO_WORLD"; // <-- the parser thinks this is an end statement, but fails to parse it
END TEST;
```

packages/language/src/parser/parser.ts
```diff
@@ -1221,6 +1230,57 @@ const endStatement = rule(
...
+      return endStatement.rule(state);
+    } else {
+      // Label doesn't match - leave END for outer scope to consume
+      return null;
+    }
+  } else {
+    // NOMULTICLOSE: END is required for each compound
+    return endStatement.rule(state);
+  }
+}
```
msujew, line 1290: Question: Is this how it works in the compiler? From my understanding, the compiler will always treat `END` statements as optional, but show a dedicated error in case an `END` statement has been skipped even though the `NOMULTICLOSE` is enabled. I.e. the parser always behaves the same, there is just an additional error message for `NOMULTICLOSE`.

The difference in behavior is that our parser will return an unexpected AST, as we don't correctly close the procedure.

msujew (commented):
Works well already. We can probably improve a few things 👍

packages/language/src/parser/parser.ts
ssmifi: I double checked. So, the main difference is that in nomulticlose an **error** is reported, and in multiclose you will get a **warning**. It is not obvious, because our build script actually exits with a failure and does not show the error of the compiler (at least in my test). 

> 0 IBM1403I E              The end of the source was reached before the logical end of the program. Null statements and END statements will be inserted as necessary to complete the program.

> [Warning] IBM1120I at line 13 in file c.pli: Multiple closure of groups. END statements will be inserted to close intervening groups. 

So, I assume you want both versions to be parsable without error. In that case, I have to add the different diagnostics to the PR. Do you agree?

msujew: Nevermind, the compiler error you're seeing does not match the documentation - Keep it is as is, since this code matches the compiler result. 👍 

Should we replace the EOF error in the parser state with with `IBM1403I`? Also can we report on `IBM1120I` when we encounter it in the parser?

packages/language/src/parser/parser-lookahead.ts
ssmifi: Good catch, thanks! I'll add the test case as well.

ssmifi:
As discussed, the parser is now emitting IBM1120I. EOF stays as it is for now. I also added two END variable test cases and a filter to specifically test for parser errors to the test builder.

msujew (approved):
Thanks, looks good to me 👍

PR #692, opened 2026-05-06
Isolate CICS dialect for reusage

Lotes:
DRAFT!

Current status:

- [x] bring over the CICS grammar and parser-related types
- [x] reuse dialect in PL/I

Later PRs:
- find a proper exchange type (propose to v2 API)
- add test cases against a "dummy dialect" in PL/I language
- syntax highlighting & semantic tokens


packages/dialect-cics/src/antlr/CICSLexer.g4
Lotes, line 1: The structure of this package was copied from the [DaCo dialect](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/tree/new-daco).

packages/language/src/parser/exec-parser.ts
```diff
@@ -13,417 +13,61 @@ import * as ast from "../syntax-tree/ast";
...
+  const prefixLength =
+    /^CICS\s*/i.exec(cicsFragmentToken.image)?.[0].length || 0;
+  const startOffset = cicsFragmentToken.startOffset + prefixLength;
+  const endOffset = cicsFragmentToken.endOffset + 1;
+  const statementText = state.textDocument
+    .getText()
+    .substring(startOffset, endOffset);
+  const preprocessor = new CICSPreprocessor();
+  const { diagnostics, identifiers } =
+    await preprocessor.execute(statementText);
```
Lotes, line 54: This should be the API v2 call.

packages/dialect-cics/src/scripts/rebuildGrammar.ts
```diff
@@ -0,0 +1,53 @@
+import { statSync } from "fs";
+import { spawnSync } from "child_process";
+import { exit } from "process";
+
+const LEXER_INPUT = "./src/antlr/CICSLexer.g4",
+  LEXER_OUTPUT = "./src/generated/CICSLexer.ts",
+  PARSER_INPUT = "./src/antlr/CICSParser.g4",
+  PARSER_OUTPUT = "./src/generated/CICSParser.ts";
+
+if (shouldRebuild(LEXER_INPUT, LEXER_OUTPUT)) {
```
Lotes: Checks if input is younger than output.
Conditional build, used to save time during development.

packages/dialect-cics/src/engine/preprocessor.ts
```diff
@@ -0,0 +1,65 @@
...
+import { CICSParser } from "../generated/CICSParser";
+import {
+  CollectingErrorListener,
+  CollectingIdentifierVisitor,
+  ParseError,
+} from "./parsing";
+
+export interface ICICSPreprocessorResult {
+  diagnostics: ParseError[];
+  identifiers: antlr.Token[];
```
msujew: Suggestion: Can we create an abstraction for this?

packages/language/src/parser/tokenizer/shared.ts
```diff
@@ -394,16 +366,15 @@ export function tokenizeIdentifier(
...
-        context.switchMode(TokenizerMode.SQL);
-        return context.createTokenInstance(tokens.SQL);
-      } else if (image === "CICS") {
-        context.advance(4, false);
-        context.switchMode(TokenizerMode.CICS);
-        return context.createTokenInstance(tokens.CICS);
+    if (
+      previousToken?.tokenTypeIdx === tokens.EXEC.tokenTypeIdx &&
+      (image === "SQL" || image === "CICS")
+    ) {
```
msujew: Suggestion: Since we are now dialect agnostic, the `image === "SQL" || image === "CICS"` comparison can be removed.

packages/language/src/parser/tokenizer/shared.ts
```diff
@@ -394,16 +366,15 @@ export function tokenizeIdentifier(
...
+      while (i < context.length && context.input[i] !== ";") {
+        i++;
```
msujew, line 371: TODO: We probably need to eventually come back to this and see what we do about the `;` issue. In case there something like `EXEC SQL INSERT ";" INTO SOME_TABLE;`, we will actually run into a rather messy situation.

msujew (commented):
Thanks, I think this is going in the right direction. A few general remarks below.

packages/dialect-cics/src/test/positives.test.ts
```diff
@@ -0,0 +1,36 @@
...
+      expect(
+        diagnostics,
+        diagnostics.length > 0
+          ? `Error at line ${line}: ${diagnostics[0].message}`
+          : undefined,
+      ).toHaveLength(0);
+      line++;
+    }
+  });
+});
```
msujew: Suggestion: Can we split each statement into its own test? That way, we don't fail the whole test suite just because we fail to parse one statement.

packages/language/src/parser/parser-state.ts
```diff
@@ -49,6 +51,7 @@ export enum RecoveryResult {
 export type RecoveryFunction = () => RecoveryResult;
 
 export class ParserState {
+  readonly textDocument: TextDocument;
```
msujew: Suggestion: I don't think this belongs here, since there is no 1:1 correspondence between the textDocument and the token list. If someone were to start using it inside of the pli-parser, they would receive unexpected results. Instead, the `StatementParser` type and `preprocessorParse` function should accept a `TextDocument` as a second argument that can be used to pass around.

Edit: Having fully reviewed the code, having a required `textDocument` here leads to a lot of new dependencies in our tests, which is a code smell.

packages/language/src/linking/resolver.ts
```diff
@@ -507,3 +506,19 @@ export function getReferenceLocations(
...
+  );
+
+  if (token && token.kind === CstNodeKind.ExecCicsStatement_COMMAND) {
+    const element = token.element;
+    if (element && element.kind === SyntaxKind.CicsExecStatement) {
+      const hostVariableReferences = element.hostVariables;
+      token = binaryTokenSearch(hostVariableReferences, offset);
+    }
+  }
+  return token;
```
msujew: Suggestion: Can we generalize this into a generic `ExecStatement` already? I don't think it's necessary to do this only for CICS, right?

msujew (commented):
Thanks looks quite nice! I have a few improvements that we should incorporate before merging this.

packages/language/src/parser/parser-state.ts
Lotes: Thanks, that helped a lot!

msujew (approved):
Thanks, looks very nice 👍

Issue #631, opened 2026-03-09
Provide pgroups names in pgm_conf.json


asatklichov:
Currently  "pgroup" values are manually inserted, let's resolve them from` proc_grps.json#name`  and provide as a drop-down value 
in for "pgroup" in  `pgm_conf.json`.

PR #707, opened 2026-05-11
Completion for pgroup names

Wagner-Laranjeiras:
Issue: #631 

Suggest registered process-group names from proc_grps.json when the cursor is inside the value of a "pgroup" property in
.pliplugin/pgm_conf.json. The feature is wired through a JSONC-aware completion request that runs alongside the existing PL/I completion path, with trigger-character cross-cancellation so quote-typing in PL/I files (and dot/percent in config files) doesn't fire the wrong pipeline.


packages/language/test/lsp/completion-plugin-configuration.test.ts
```diff
@@ -0,0 +1,206 @@
...
+            ]
+        }`;
+    const { text, offset } = fromMarker(mockPgmConf);
+    const result = configCompletionRequest(
+      pluginConfig,
+      text,
+      offset,
+      pgmConfigUri(),
+    );
+    const expected = pluginConfig.getProcessGroupNames();
```
msujew: Suggestion: This way of testing features is prone to breakage. Since both the test **and** the implementation rely on `getProcessGroupNames`, an error in this function will still pass the test, but fail during the actual execution. Be more explicit about what you expect to see in this test. The same applies to the others. No test should call `getProcessGroupNames` to set the expectation.

packages/language/src/language-server/connection-handler.ts
```diff
@@ -201,8 +203,43 @@ export function startLanguageServer(
...
+    );
+    const triggerLang = params.context?.triggerCharacter
+      ? TRIGGER_CHAR_LANG[params.context.triggerCharacter]
+      : undefined;
+    if (
+      (triggerLang === "config" && !isConfigDocument) ||
+      (triggerLang === "pli" && isConfigDocument)
+    ) {
+      return [];
+    }
```
msujew, line 227: Suggestion: Can we extract this into a `selectCompletionMode` function?

packages/language/src/language-server/completion/completion-plugin-configuration.ts
```diff
@@ -0,0 +1,88 @@
...
+  uri: URI,
+): CompletionItem[] {
+  const rootNode = jsoncParseTree(documentText);
+  if (!rootNode) {
+    return [];
+  }
+  const currentNode = jsoncFindNodeAtOffset(rootNode, offset, true);
+  if (!currentNode) {
+    return [];
+  }
```
msujew: Question: Could `getLocation` work better here? As far as I can tell, this only works if we are within the string, but not if the cursor is right after the colon, right? See [this code from the npm extension](https://github.com/microsoft/vscode/blob/main/extensions/npm/src/features/jsonContributions.ts).

msujew (commented):
Works quite well, good job!

I noticed that when placing the cursor just before the string, the result is malformed JSON:

```ts
// before
"pgroup":  <|>"default"
// after
"pgroup":  default"
```

Also, when placing the cursor behind the `:`, no completion appears. See the question below for a potentially better approach to this.

msujew (approved):
Thank you, works very well!

PR #708, opened 2026-05-11
Support compiler testing

msujew:
Adds functionality to run our own fourslash tests against the compiler. Adds ~120 tests to this special `@compiler` testing suite, but does not implement all of the testing functionality - this can be done as a follow up. Similarly, we can add more test cases in the long run to this suite.

A detailed description of how to run the tests is outlined in the `TESTING.md` file.


packages/language/test/list-file-parser.ts
```diff
@@ -0,0 +1,237 @@
...
+  compilerMessages: CompilerMessage[];
+  macroMessages: CompilerMessage[];
+  fileTable: FileTableEntry[];
+}
+
+function parsePosition(text: string | undefined): ListFilePosition | undefined {
+  if (!text) return undefined;
+  const match = text.match(/^(\d+)\.(\d+)$/);
+  if (!match) return undefined;
+  return { line: parseInt(match[1], 10), file: parseInt(match[2], 10) };
```
Lotes, line 57: Interesting that the second number is a file index

packages/language/test/compiler-test-builder.ts
```diff
@@ -0,0 +1,300 @@
...
+    for (let i = 0; i < expectedTokens.length; i++) {
+      if (expectedTokens[i] !== actualTokens[i]) {
+        throw new Error(
+          `Token mismatch at index ${i}: expected '${expectedTokens[i]}', but got '${actualTokens[i]}'.`,
+        );
+      }
+    }
+  }
+
+  expectLinks(): void {
```
Lotes, line 86: Is it doable to have a base class for our test builders? It seems that this logic (for loops around indices and ranges) is duplicated. Correct me if I am wrong.

I was just wondering that this logic appears here. I thought that could be moved out then.

packages/language/test/fourslash/preprocessor/implicit-declaration-can-be-activated.ts
```diff
@@ -11,6 +11,7 @@
 
 /// <reference path="../framework.ts" />
 
+// @compiler: true
```
Lotes, line 14: What makes you decision for adding this compiler tag?
When does a test does not get the tag?

packages/language/test/fourslash-harness/extractor.ts
```diff
@@ -0,0 +1,70 @@
...
+      wrappers,
+    });
+    if (testFile.tags["compiler"] !== "true") {
+      if (testFile.tags["compiler"] === "fail") {
+        failures.push(file);
+      }
+      continue; // Only extract compiler tests
+    }
+    // Target path for the generated workspace - based on the test path/name
+    let testTargetPath = path.join(targetDir, file);
```
Lotes: The test file a/b/c/test.ts will end up with its workspace at output/compiler/a/b/c/test/. Are the workspaces marked as standalone? Or can they interact with each other by accident? (I have not built the config part, but I had the impression it marks the workspace as such)

packages/language/test/fourslash-harness/extractor.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+    const entries = Array.from(testFile.files.entries());
+    const [[firstFileUri]] = entries;
+    if (!firstFileUri) {
+      console.warn(`No files found in test ${file}`);
+      continue;
+    }
+    const firstFilePath = UriUtils.toFilePath(firstFileUri);
+    const relativeDir = path.dirname(firstFilePath);
+    for (const [fileUri, file] of testFile.files) {
```
Lotes, line 50: `file` was defined twice in this scope. Can be confusing. Maybe vary the name of the variable.

packages/language/test/fourslash/preprocessor/implicit-declaration-can-be-activated.ts
msujew: I've just added this for now to most preprocessor tests and to the linking tests. A lot of other tests won't run against the compiler, because they use unimplemented parts of the testing harness interface.

packages/language/test/fourslash-harness/extractor.ts
msujew: Exactly. The workspaces are supposed to be standalone, however I still need to work out how we properly do tests with include - all current `@compiler` tests don't use any include statements. That's something for a separate PR.

packages/language/test/compiler-test-builder.ts
msujew: Yep, there is some overlap. Abstracted this away into a `AbstractTestBuilder`.

Lotes (approved):
Looks good in my eyes. If you need more feedback, you could reach out to Steven.

ssmifi:
Thank you. I will have a look on Friday.

docs/TESTING.md
```diff
@@ -0,0 +1,24 @@
...
+
+Running all tests is as simple as running `pnpm test` in your terminal. Additional testing scripts can be found in the `package.json` of the root project. Running the fourslash tests only can be done via the `pnpm test:fourslash` script.
+
+## Running the compiler tests
+
+An additional benefit to the fourslash infrastructure is the ability to run the tests against the compiler. This enables developers to confirm that the behavior of the language server aligns to the behavior of the compiler.
+
+The compiler tests are run in three separate phases:
+1. The extraction phase is triggered via the `pnpm extract-compiler-tests` script. It expects an additional path as an argument like `pnpm extract-compiler-tests ../output/compiler`, where it will place all of the test files.
+2. The second phase requires use of a PL/I mainframe compiler. All output files in the aforementioned `../output/compiler` directory need to be compiled. This should provide a `.list` file for each test case, using the same structure as the input test files. The listings should be placed somewhere relative to this repository (e.g. `../output/listings`).
```
ssmifi, line 17: _Suggestion_: Create a pinned note in our intranet with instructions on how to compile all tests on the mainframe.

packages/language/test/fourslash/builtins/pli/assignability-conversion-failed.ts
```diff
@@ -11,8 +11,11 @@
 
 /// <reference path="../../framework.ts" />
 
+// The actual compiler does not throw here at all
+// @compiler: fail
```
ssmifi: _Suggestion_: Document the `@compiler` tag including expectation in the `TESTING.md`. (It would probably be good to document all tags there and/or the whole fourslash framework there). Also, if understood correctly, wouldn't `skip` be the canonical name for the purpose?

packages/language/test/fourslash-harness/implementation/compiler-test-builder.ts
```diff
@@ -0,0 +1,107 @@
...
+  };
+  return {
+    Syntax: SyntaxKind,
+    linker: {
+      expectLinks: () => testBuilder.expectLinks(),
+      expectNoLinksAt: notImplemented("linker.expectNoLinksAt"),
+    },
+    testAPI: {
+      // Stub, will not be used during testing
+      testBuilder: await TestBuilder.create("", {}),
```
ssmifi: _Minor Suggestion_: Could make this explicit, eg, `get testBuilder(): TestBuilder { throw new Error('testAPI.testBuilder is not available in compiler tests. ' ...`

packages/language/test/fourslash-harness/extractor.ts
```diff
@@ -0,0 +1,70 @@
...
+import { fourslashPath, getTestFiles } from "./utils";
+import { UriUtils } from "../../src/utils/uri";
+import { replaceNamedIndices } from "../utils";
+
+export async function extractCompilerTestFiles(
+  targetDir: string,
+): Promise<void> {
+  const wrappers = getWrappers();
+  const files = getTestFiles();
+  const failures: string[] = [];
```
ssmifi: _Minor_: Related to the comment above. For me, it looks like deliberately skipped and not really test failures. Could be misleading: In other test frameworks, "fail" is still tested.

packages/language/test/fourslash-harness/extractor.ts
```diff
@@ -0,0 +1,70 @@
...
+    const relativeDir = path.dirname(firstFilePath);
+    for (const [fileUri, file] of testFile.files) {
+      const fileContent = replaceNamedIndices(file.content).output;
+      // Generate relative path within the test files
+      const currentFilePath = UriUtils.toFilePath(fileUri);
+      const relativePath = path.relative(relativeDir, currentFilePath);
+      // Write the file to the target directory, preserving the relative structure
+      const targetPath = path.join(testTargetPath, relativePath);
+      const targetDirPath = path.dirname(targetPath);
+      await fs.promises.mkdir(targetDirPath, { recursive: true });
```
ssmifi, line 58: File operations can fail.

packages/language/test/fourslash-harness/utils.ts
```diff
@@ -9,7 +9,55 @@
...
+import path from "path";
+import { HarnessTest, UnnamedFile } from "./types";
+import { PliTestFile } from "../utils";
+
+export const frameworkFileName = "framework.ts";
+export const testsPath = "packages/language/test/fourslash";
+
+const url = new URL(import.meta.url);
+const __dirname =
+  process.platform === "win32"
```
ssmifi: _Question_: This gives me minor url refactoring flashbacks. :)
Do we need to have a manual platform check? Doesn't `fileURLToPath` work here?

packages/language/test/fourslash-harness/utils.ts
```diff
@@ -9,7 +9,55 @@
...
+/**
+ * The path to the `fourslash` directory.
+ *
+ * Important: Assume that the test files exist in the `packages/language/test/fourslash` directory.
+ */
+export const fourslashPath = path.join(__dirname, "../fourslash");
+
+export const harnessPath = __dirname;
+
+export function getTestFiles() {
```
ssmifi: _Minor_: Add return type for type-safety.

packages/language/test/fourslash-harness/utils.ts
```diff
@@ -9,7 +9,55 @@
...
+  process.platform === "win32"
+    ? path.dirname(url.pathname).slice(1)
+    : path.dirname(url.pathname);
+
+/**
+ * The root of the project.
+ *
+ * Important: Assume that the test files exist in the `packages/language/test/fourslash` directory.
+ */
+export const projectRoot = path.join(__dirname, "../../../..");
```
ssmifi, line 26: _Minor Suggestion_: Validate at test startup (somewhere?) that these paths really exist and if not, give a readable error message. For the root, perhaps just check for the `package.json`.

packages/language/test/abstract-test-builder.ts
```diff
@@ -0,0 +1,185 @@
...
+ * Base class for test builders that share the same marker-based file/index/range setup.
+ *
+ * Subclasses hold backend-specific state (e.g. a `CompilationUnit` or a compiler listing)
+ * but inherit the shared scaffolding for: file/marker initialization, label resolution,
+ * and the diagnostic matcher used by `expectDiagnosticsAt`-style assertions.
+ */
+export abstract class AbstractTestBuilder {
+  protected files: Map<string, TestFile> = new Map();
+  protected indices!: Record<string, TestIndex[]>;
+  protected ranges!: Record<string, TestRange[]>;
```
ssmifi: There is no guarantee that `initMarkerState` will be called.

packages/language/test/compiler-test-builder.ts
```diff
@@ -0,0 +1,259 @@
...
+    }
+    return {
+      file: listingFile,
+      line: this.offsetToLine(sourceFile, offset),
+    };
+  }
+
+  private findListingFile(uri: string): File | undefined {
+    const index = uri.lastIndexOf("/");
+    const baseName = index >= 0 ? uri.slice(index) : uri;
```
ssmifi, line 148: `Init` used `first.uri.slice(lastIndex + 1)` for the `baseName`. Is the `+1` discrepancy correct?

packages/language/test/compiler-test-builder.ts
```diff
@@ -0,0 +1,259 @@
...
+export class CompilerTestBuilder extends AbstractTestBuilder {
+  private listing!: ListFile;
+
+  private async init(files: PliTestFile[], outputDir: string): Promise<void> {
+    // The path of the listing is always the path to the main .pli file but with a .list extension
+    const first = files[0];
+    const lastIndex = first.uri.lastIndexOf("/");
+    const baseName =
+      lastIndex >= 0 ? first.uri.slice(lastIndex + 1) : first.uri;
+    const outputListing = `${outputDir}/${baseName}.list`;
```
ssmifi: _Question_: Does the manual path construction and parsing work on all fs?

packages/language/test/compiler-test-builder.ts
```diff
@@ -0,0 +1,259 @@
...
+    } else {
+      expectedTokens = tokenize(tokens, undefined).tokens.map((e) => e.image);
+    }
+    const actualTokens = tokenize(
+      this.listing.sourceCode,
+      undefined,
+    ).tokens.map((t) => t.image);
+    if (expectedTokens.length !== actualTokens.length) {
+      throw new Error(
+        `Expected ${expectedTokens.length} tokens, but got ${actualTokens.length}.`,
```
ssmifi: _Suggestions_: Should we print out the actual tokens for debugging?

packages/language/test/list-file-parser.test.ts
```diff
@@ -0,0 +1,157 @@
...
+  test("extracts source lines", () => {
+    expect(result.compilerSource.length).toBeGreaterThan(0);
+  });
+
+  test("main procedure is found at the correct position", () => {
+    const mainProc = result.compilerSource.find((l) =>
+      l.text.includes("AVERAGE: PROCEDURE"),
+    );
+    expect(mainProc).toBeDefined();
+    expect(mainProc!.position).toEqual({ line: 4, file: 0 });
```
ssmifi, line 72: _Minor Suggestion_: This file will probably stay the same, I guess, but the expectations would be more robust using our/a marker <|...|> system.

packages/language/test/list-file-parser.test.ts
```diff
@@ -0,0 +1,157 @@
...
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { describe, expect, test } from "vitest";
+import { parseListFile } from "./list-file-parser.js";
```
ssmifi: _Minor_: Do we need the `.js` here?

packages/language/test/list-file.ts
```diff
@@ -0,0 +1,131 @@
...
+  listFile.diagnostics.push(
+    ...parsed.macroMessages.map((msg) =>
+      compilerMessageToDiagnostic(listFile, msg),
+    ),
+  );
+  listFile.definitions = parsed.xrefTable.map((entry) => ({
+    position: entry.definedAt
+      ? {
+          line: entry.definedAt.line,
+          file: listFile.getFile(entry.definedAt.file)!,
```
ssmifi: _Question_ Are these Non-Nulls always safe?

packages/language/test/list-file.ts
```diff
@@ -0,0 +1,131 @@
...
+    })),
+  }));
+  return listFile;
+}
+
+function compilerMessageToDiagnostic(
+  listFile: ListFile,
+  msg: CompilerMessage,
+): Diagnostic {
+  let severity: Severity = Severity.I;
```
ssmifi, line 108: In the list-file-parser, there is also a severity `L`. Should `L` also be handled here?

ssmifi (commented):
Looks nice, thanks. I only have a couple of minor suggestions/questions.

packages/language/test/fourslash-harness/extractor.ts
msujew: Yes, but that's alright, because that will crash the script with the respective error message.

packages/language/test/compiler-test-builder.ts
msujew: Yes, because we want to capture the leading `/` to use the `endsWith` function. Otherwise, it will accept all file names that end with the expected name.

I'll add a comment there.

packages/language/test/list-file-parser.test.ts
msujew: I'm not sure how that would help here actually.

ssmifi: I was just thinking about potential future additions to the test, but as I said, probably not necessary. Just leave it as it is.

ssmifi (approved):
Looks nice. Thanks for the effort!

PR #709, opened 2026-05-12
Fix LAXDCL in linking

ssmifi:
Fixes https://github.com/zowe/zowe-pli-language-support/issues/702

Small bug, was testing `laxDef` instead of `laxDcl`. Added a test.

### DIFF
--- a/packages/language/test/fourslash/linker/implicit-declaration-laxdcl.ts
+++ b/packages/language/test/fourslash/linker/implicit-declaration-laxdcl.ts
@@ -1,0 +1,20 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


packages/language/test/fourslash/linker/implicit-declaration-laxdcl.ts
```diff
@@ -0,0 +1,21 @@
...
+/// <reference path="../framework.ts" />
+
+// @wrap: process
+////*PROCESS RULES(LAXDCL);
+//// A: PROCEDURE() OPTIONS (MAIN);
+////     <|1:J|> = 72;
+//// END A;
+
+verify.noDiagnostics(1, code.Error.IBM1373I);
+linker.expectLinks();
```
msujew: Suggestion: The `expectLinks` isn't doing anything here as far as I can tell.

msujew (approved):
Looks good. One minor issue in the test.

PR #711, opened 2026-05-12
Fix escaping < and > in markdown comments for signature help and hover

### LINKED ISSUE #687: Invalid escaping on `<>` in documentation code blocks
While `<>` need to be escaped in normal markdown, they shouldn't be escaped within code blocks. This leads to this behavior in our language server:

Image

Lotes:
Closes #687

### DIFF
--- a/packages/language/src/documentation/jsdoc.ts
+++ b/packages/language/src/documentation/jsdoc.ts
@@ -571,36 +571,34 @@ class JSDocCommentImpl implements JSDocComment {
     return this.escapeNonVerbatimCode(value.trim());
   }
 
-  private escapeNonVerbatimCode(text: string): string {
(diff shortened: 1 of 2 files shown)


packages/language/src/documentation/jsdoc.ts
```diff
@@ -568,7 +568,41 @@ class JSDocCommentImpl implements JSDocComment {
         value += fillNewlines(value) + text;
       }
     }
-    return value.trim().replace(/</g, "&lt;").replace(/>/g, "&gt;");
+    return this.escapeNonVerbatimCode(value.trim());
+  }
+
+  private escapeNonVerbatimCode(text: string): string {
```
Lotes: I can fix this code duplication here...

msujew (approved):
Looks good, thanks!

PR #714, opened 2026-05-13
Add support for multi-line process directives

ssmifi:
Fixes https://github.com/zowe/zowe-pli-language-support/issues/699
Fixes https://github.com/zowe/zowe-pli-language-support/issues/698

This PR is two-fold, it first
- adds the remaining `F` and `S` shortcuts to the `FULL` and `SHORT` length parameters, which is not documented but confirmed on the mainframe for the options affected in this PR
- adds the `OF` and `NOF` abbreviations to the compiler options, which are not documented but confirmed on the mainframe
- Also removed some now unnecessary code in the compiler option translation 

and second, 
- refactors the compiler option extraction to allow multi-line process directives and keep that code maintainable. It is a bit aligned to the recently refactored lexer code, so I hope you approve the refactoring. Please let me know what you think. 
  - Therefore, the `;` is now mandatory which corresponds to the behavior on the mainframe
  - Also note that the compiler option parser did not change and handles the multi-line segment well. This is only about the extraction and keeping the positions correct.
- adds a test for the multi-line directive

_Edit_: 
- Also adds the M, NM abbreviations to the compiler options


packages/language/src/preprocessor/compiler-options-processor.ts
```diff
@@ -133,77 +131,178 @@ export class CompilerOptionsProcessor {
...
+    let inString: number = 0; //charCode of the opening quote or 0
+
+    let processStart = 0;
+    let processStartLine = 0;
+
+    const la = (n: number): number =>
+      index + n < text.length ? text.charCodeAt(index + n) : -1;
+
+    const consumeLine = () => {
+      while (index < text.length && text.charCodeAt(index) != 10) {
```
msujew: Question: What is `10` as a char code?

packages/language/src/preprocessor/compiler-options-processor.ts
```diff
@@ -133,77 +131,178 @@ export class CompilerOptionsProcessor {
...
+              start: processStart,
+              end: index,
+            });
+            inDirective = false;
+            continue;
+          }
+        }
+      }
+
+      if (code == 10) {
```
msujew: Same here

msujew (commented):
Looks mostly quite good. Just a question on the use of magic numbers.

packages/language/src/preprocessor/compiler-options-processor.ts
ssmifi: Ah, sry, that is the LF code. I'll add a constant.

msujew (approved):
Thanks, looks good to me 👍

PR #719, opened 2026-05-13
Semantic highlighting for CICS dialect

Lotes:
Reintroduces semantic highlighting for the CICS dialect.


packages/dialect-cics/src/engine/parsing.ts
```diff
@@ -26,16 +26,24 @@ import {
...
 import { CicsWordContext } from "../generated/CICSParser";
 
-export interface Identifier {
-  name: string;
+export enum SemanticsKind {
+  Identifier,
+  Keyword,
+  String,
+  Comment,
+}
```
msujew, line 41: Minor: Now that I see this, `Number` probably also makes sense.

packages/dialect-cics/src/engine/preprocessor.ts
```diff
@@ -58,12 +61,46 @@ export class CICSPreprocessor {
...
+          idIndex < identifierTokens.length &&
+          token.start === identifierTokens[idIndex].startOffset
+        ) {
+          return identifierTokens[idIndex++];
+        } else if (token.channel === COMMENTS) {
+          semanticsKind = SemanticsKind.Comment;
+        } else if (keywordPattern.test(token.text!)) {
+          semanticsKind = SemanticsKind.Keyword;
+        } else if (stringPattern.test(token.text!)) {
+          semanticsKind = SemanticsKind.String;
```
msujew: Suggestion: AFAIK there are multiple string-like tokens in the grammar, essentially all `NONNUMERICLITERAL` terminals. We should check for those instead of using a regexp. Using a regexp is also rather slow.

packages/dialect-cics/src/engine/preprocessor.ts
```diff
@@ -58,12 +61,46 @@ export class CICSPreprocessor {
...
+        let semanticsKind: SemanticsKind;
+        if (
+          idIndex < identifierTokens.length &&
+          token.start === identifierTokens[idIndex].startOffset
+        ) {
+          return identifierTokens[idIndex++];
+        } else if (token.channel === COMMENTS) {
+          semanticsKind = SemanticsKind.Comment;
+        } else if (keywordPattern.test(token.text!)) {
+          semanticsKind = SemanticsKind.Keyword;
```
msujew, line 84: Suggestion: It's probably enough to check whether the text starts with an ASCII letter. Also move it after the string check. Using regexp here is rather slow.

packages/language/src/syntax-tree/ast.ts
```diff
@@ -4360,14 +4361,16 @@ export function createCicsResponseStatement(): CicsResponseStatement {
 
 export interface CicsExecStatement extends AstNode {
   kind: SyntaxKind.CicsExecStatement;
-  hostVariables: Token[];
+  tokens: Token[];
+  semanticsTokenTypes: SemanticTokenTypes[];
```
msujew: Suggestion: Can we merge this into one object? Having dependent, but disjointed data here feels wrong. Something like:

```ts
interface DialectToken {
  token: Token;
  semanticType: SemanticTokenTypes;
}
```

That would also make it easier to add more data (if necessary) here in the future.

packages/dialect-cics/src/engine/test/positives.test.ts
msujew, line 1: Suggestion: This file shouldn't be part of the PR, right? Something might have gone wrong during your last rebase.

packages/dialect-cics/src/engine/test/positives.txt
msujew, line 1: Same here.

msujew (approved):
Nice, that works pretty well. Aside from the two issues below (the files can probably just be removed?) this looks good 👍

packages/dialect-cics/src/engine/test/positives.test.ts
Lotes: The rebase was fine, I moved these two file somewhen inbetween.

PR #720, opened 2026-05-13
Implement diagnostic and quick fix for unknown "pgroup" entries.

Wagner-Laranjeiras:
Follow up for PR #707 . Issue #631.


msujew (commented):
Thanks, the code itself makes sense. However, I'm getting a bit concerned with the amount of testing code that we have to maintain. Do you think there is some way to either:
1. Integrate this part into the fourslash framework.
2. Write some other testing infrastructure to abstract away some of that complexity.
We're now at >1300 LoC in the `apply-quick-fixes.test.ts` and it becomes difficult to understand what each test is doing due to the amount of boilerplate in there.

packages/language/test/test-builder.ts
```diff
@@ -250,24 +269,45 @@ export class TestBuilder extends AbstractTestBuilder {
...
+        procGrpsUri = uri;
+        procGrpsOutput = file.output;
       }
     }
+    if (pgmConfUri && procGrpsUri) {
+      await defaultTestWorkspace().config.init(
+        UriUtils.toUri(
+          pgmConfUri.replace(PluginConfiguration.PROGRAM_FILE_PATH, ""),
+        ),
+      );
```
msujew: Suggestion: Using `replace` doesn't really explain what this code is _supposed_ to do.
```suggestion
      const configUri = UriUtils.dirname(UriUtils.toUri(pgmConfUri));
      await defaultTestWorkspace().config.init(configUri);
```

packages/language/test/test-builder.ts
```diff
@@ -250,24 +269,45 @@ export class TestBuilder extends AbstractTestBuilder {
...
+        pgmConfOutput,
+      )
+    );
+    let hasProcessGroupConfig = false;
+    if (procGrpsOutput) {
+      await defaultTestWorkspace().config.parseProcessGroupConfigs(
+        procGrpsOutput,
+      );
+      hasProcessGroupConfig = true;
+    }
```
msujew: Suggestion: I feel like this is becoming rather convoluted. Instead, can we simply write the default config files into the in-memory filesystem, if they are not already provided? That completely eliminates the code from line 297-328.

packages/language/test/workspace/unknown-process-group.test.ts
msujew, line 1: Suggestion: Except for the tests calling `reloadConfigurations`, I believe that these tests can be moved into the fourslash framework, right?

packages/language/test/lsp/apply-quick-fixes.test.ts
msujew, line 1: Suggestion: I believe it should be possible to cover the new tests in this file via the fourslash framework, right?

packages/language/test/fourslash/validate/unknown-pgroup-single-bad-ref.ts
```diff
@@ -0,0 +1,54 @@
...
+// The diagnostic targets the offending pgroup value in pgm_conf.json,
+// carrying the COPC04E code and the offender in its message.
+verify.expectExclusiveDiagnosticsAt("bad", {
+  code: code.LSP.PluginConfiguration.UnknownProcessGroup.code,
+  message: `Unknown process group 'doesnotexist'.`,
+  severity: constants.Severity.E,
+});
+
+// COPC04E is the only diagnostic in the whole compilation unit.
+verify.noDiagnosticsExcept([code.LSP.PluginConfiguration.UnknownProcessGroup]);
```
msujew: Suggestion: I don't think we need this one. Using `noDiagnosticsExcept` is always very fickle once we implement more diagnostics.

packages/language/test/fourslash/validate/unknown-pgroup-multiple.ts
```diff
@@ -0,0 +1,55 @@
...
+  message: `Unknown process group 'ghost'.`,
+});
+verify.expectExclusiveDiagnosticsAt("ghost2", {
+  message: `Unknown process group 'ghost'.`,
+});
+verify.expectExclusiveDiagnosticsAt("other", {
+  message: `Unknown process group 'other-ghost'.`,
+});
+
+verify.noDiagnosticsExcept([code.LSP.PluginConfiguration.UnknownProcessGroup]);
```
msujew: Same here.

packages/language/test/fourslash/validate/unknown-pgroup-empty-value.ts
```diff
@@ -0,0 +1,41 @@
...
+////       "libs": []
+////     }
+////   ]
+//// }
+
+// @filename: main.pli
+// @wrap: main
+//// DCL A CHAR(8);
+
+verify.noDiagnosticsExcept([code.LSP.PluginConfiguration.UnknownProcessGroup]);
```
msujew: Suggestion: Where does this diagnostic appear? The test should flag the position using a label.

packages/language/test/test-builder.ts
```diff
@@ -195,7 +195,26 @@ export class TestBuilder extends AbstractTestBuilder {
...
+          textDocument: file.textDocument,
+        });
+      }
+    }
     this.diagnostics = this.unit.diagnostics.getAll();
+    const configDiagnostics =
+      defaultTestWorkspace().config.getConfigInternalDiagnostics();
+    for (const diagnostic of configDiagnostics) {
+      this.diagnostics.push(...diagnostic[1]);
+    }
```
msujew:
```suggestion
    for (const diagnostics of configDiagnostics.entries()) {
      this.diagnostics.push(...diagnostics);
    }
```

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -308,24 +330,46 @@ export class PluginConfigurationProvider {
...
+      unknownProcessGroupsDiagnostic = validatePgroupReferences(
+        pgroupReferences,
+        pgroupNames,
+        document.uri,
+      );
+      for (const diagnostic of unknownProcessGroupsDiagnostic) {
+        const lspDiag = toLspDiagnostic(diagnostic, document);
+        programConfigDiagnostics.push(lspDiag);
+      }
+    }
```
msujew, line 383: Suggestion: This is too specific for being in the general `loadConfigurations` method. Outsource this into some `validatePluginConfig()` function.

packages/language/src/config/cross-validation.ts
```diff
@@ -0,0 +1,50 @@
...
+  Diagnostic,
+  diagnosticFromCodeAtRange,
+  offsetLengthToRange,
+} from "../language-server/types";
+import { LspCodes } from "../validation/lsp-codes";
+
+/**
+ * Check for unknown process groups references under "pgm_conf.json"
+ * by comparing with existing process group names at "proc_grps.json"
+ * If process group names existis and an unknown process is found,
```
msujew: Typo. Also what does this mean?
> If process group names existis and an unknown process is found,

msujew (commented):
Thanks, already looking much better!

packages/language/src/config/cross-validation.ts
Wagner-Laranjeiras: Thanks for catching it! This is the function that find the potential diagnostics. I meant that "If we in fact have process names to compare and the comparison finds one or more mismatches, then we throw the diag (instead of returning early if there are no values to compare). Now that you pointed out I can see it could be better phrased. 😅 Working on that

Wagner-Laranjeiras: Wrote like this:

> /**
>  * Check for unknown process groups references under "pgm_conf.json"
>  * by comparing with existing process group names at "proc_grps.json"
>  * If there are values to compare and a mismatch between them is found,
>  * throw UnknownProcessGroup(COPC04).
>  */

packages/language/test/test-builder.ts
Wagner-Laranjeiras: Got it, I'm currently writing like the following, because using only one `dirname` maintains the `.pliplugin` folder in the URI.

```
const configUri = UriUtils.dirname(UriUtils.dirname(UriUtils.toUri(pgmConfUri)))
```

packages/language/test/fourslash/validate/unknown-pgroup-empty-value.ts
Wagner-Laranjeiras: I removed the `verify.noDiagnosticsExcept([code.LSP.PluginConfiguration.UnknownProcessGroup]);` and added a label, it worked. Btw, there's a similar test on the quick-fixes folder, but I didn't delete this one because it tests the diagnostic, not the quick fix. WDYT?

packages/language/test/test-builder.ts
```diff
@@ -250,42 +277,37 @@ export class TestBuilder extends AbstractTestBuilder {
...
-          pgroup:
-            PluginConfiguration.DEFAULT_PROGRAM_FILE_CONTENT.pgms[0].pgroup,
-        }),
-      ]);
+    if (pgmConfUri && procGrpsUri) {
+      const configUri = UriUtils.dirname(
+        UriUtils.dirname(UriUtils.toUri(pgmConfUri)),
+      );
+      await defaultTestWorkspace().config.init(configUri);
+      return;
```
msujew: Suggestion: We only need one `init` call.

msujew (approved):
Thanks, looks much better. Just one more suggestion - then this is good to go 👍

packages/language/test/test-builder.ts
Wagner-Laranjeiras: Yep, that makes sense 😅 this code block was inside the loop before, so we could init the config and return early before having to necessarily read every file. It makes no sense to maintain it outside the for loop.

PR #721, opened 2026-05-13
Improve skipped code eval

ssmifi:
Related to https://github.com/zowe/zowe-pli-language-support/issues/716

Currently, the skipped code indicator only works at `IF` if there is an else branch.
If the branch is missing and the (true) branch is missing from the evaluation results, it is not clear, if the condition could not be evaluated or if the result was false (which is displayed differently w.r.t. the skipped code indicator). 

To fix this, this PR also stores `undefined` in the evaluation results to indicate when the condition (resp. case values) couldn't be evaluation and in which case it should not be treated as skipped code.

### DIFF
--- a/packages/language/test/fourslash/preprocessor/skipped-code-if-no-else-unknown.ts
+++ b/packages/language/test/fourslash/preprocessor/skipped-code-if-no-else-unknown.ts
@@ -1,0 +1,20 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+// @wrap: main
+//// %DECLARE ARR(5) CHARACTER;
+//// %IF ARR %THEN <|notSkipped:%DO;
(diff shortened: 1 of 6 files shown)


packages/language/test/fourslash/preprocessor/skipped-code-if-no-else-unknown.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+// @wrap: main
+//// %DECLARE ARR(5) CHARACTER;
+//// %IF ARR %THEN <|notSkipped:%DO;
+////   WHAT = 123;
+//// %END;|>
```
msujew, line 18: Question: I'm not entirely sure about this. Our implementation will skip this code, doesn't it? It doesn't feel weird not displaying it as skipped.

msujew (commented):
Looks mostly good - I'm just not entirely sure when to show and when to skip certain sections. See the question below.

packages/language/test/fourslash/preprocessor/skipped-code-if-no-else-unknown.ts
ssmifi: Only if there is an `%ELSE` currently. I would not favor any branch if the evaluation does not succeed. This should probably instead give a diagnostic at the condition instead. wdyt?

msujew (approved):
Let's keep it like this for now and get back to this once it's clearer how to handle this 👍

PR #725, opened 2026-05-18
Parse multiple attributes in allocate statement

### LINKED ISSUE #717: Go Green - on all source samples
We have still some sample codes not processed fully, there are still some minor issues. 
Let's keep the source code analysis green.

msujew:
Part of an effort to resolve https://github.com/zowe/zowe-pli-language-support/issues/717.

Our `ALLOCATE` statement parser didn't allow for multiple attributes - only one at most. This change simply changes this to a loop and stores the results in an array.

### DIFF
--- a/packages/language/test/fourslash/parser/allocate-multiple-attributes.ts
+++ b/packages/language/test/fourslash/parser/allocate-multiple-attributes.ts
@@ -1,0 +1,17 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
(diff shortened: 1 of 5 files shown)


packages/language/test/fourslash/parser/allocate-multiple-attributes.ts
```diff
@@ -9,5 +9,9 @@
  *
  */
 
-export * from "./engine/preprocessor";
-export * from "./engine/parsing";
+/// <reference path="../framework.ts" />
+
+// @wrap: main
+//// ALLOCATE TEST CHAR(10) INIT("VALUE");
```
Lotes, line 15: May add another test where LEVEL is used. Just to cover the syntax as well.

Does TEST need to be declared first? Or is this a declaration?

Lotes (approved):
Look good. I was just wondering if a declaration is missing.

packages/language/test/fourslash/parser/allocate-multiple-attributes.ts
msujew: It should be declared first, but that has no impact on the actual parsing of the statement.

PR #726, opened 2026-05-18
Fix empty EXEC statement results

### LINKED ISSUE #717: Go Green - on all source samples

msujew:
Part of an effort to resolve https://github.com/zowe/zowe-pli-language-support/issues/717.

In our test files, we have a bunch of `EXEC CICS RETURN;` statements. Currently, our EXEC statements don't emit anything. This ensures that as a temporary fix, they always emit an empty DO statement to ensure that the real PLI parser has something to grab onto.

This is essentially a simplification of the real preprocessor logic.


Lotes (approved):
Nice idea. Thanks.

PR #728, opened 2026-05-20
Reintroduce Db2Sql using ANTLR grammar

Lotes:
* Renamed "dialect-*" to "preprocessor-*", since "dialect" is the wrong terminology here.
* I removed everything referencing to the old SQL and CICS logic (tokens, tokenizers, ...)
* include logic for SQL is inside the preprocessor, have a look if you like the exchange format/type between preprocssor and PL/I


packages/language/src/preprocessor/instruction-generator.ts
```diff
@@ -209,13 +209,18 @@ function generateInstructionForStatement(
...
 
-function generateExecInstruction(
-  stmt: ast.ExecStatement,
-): inst.ExecInstruction {
+function generateExecInstruction(stmt: ast.ExecStatement): inst.Instruction {
+  if (stmt.replacement) {
+    if (typeof stmt.replacement === "object") {
+      return generateIncludeInstruction(stmt.replacement)!;
+    }
+  }
```
msujew: Suggestion: Merge into one `if` condition.

packages/preprocessor-api/src/types.ts
```diff
@@ -0,0 +1,53 @@
...
+export type PreprocessorReplacement =
+  | {
+      type: "text";
+      text: string;
+    }
+  | {
+      type: "include";
+      filePath: string;
+      token: Token;
+    };
```
msujew, line 43: Suggestion: The definition provider should use this token to return the expected definition link.

packages/preprocessor-db2/src/antlr/message-service-parser.ts
msujew, line 1: Suggestion: Add a comment that we need this for backwards compatibility with the COBOL DB2 implementation.

msujew (commented):
Nice, works pretty well. Just a few minor improvements and this should be good to go :)

msujew (approved):
Looks good to me, thanks 👍

PR #729, opened 2026-05-20
Properly support INITIAL attribute content

msujew:
Implements a larger refactoring of the parser code related to expressions & the `INITIAL` attribute:

- Init attributes can now feature arbitrary repetitions as part of the contained expressions. I.e. `(VAR)((2)("HELLO"), "WORLD)`
- The preprocessor can now accept those repetitions and populate the initial declaration of an array variable as expected.
- Wildcards (`*`) are now part of the expressions (and need to be handled in the type system, later on) to support this.


packages/language/test/fourslash/parser/initial-wildcard-empty.ts
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+// @wrap: main
+//// DCL TEST(3) CHAR(5) INIT("HELLO", *, "WORLD");
```
Lotes, line 15: May add a test API to get constant values?

packages/language/src/typesystem/descriptions.ts
```diff
@@ -78,7 +78,13 @@ export enum AttributeKind {
   /** @see https://www.ibm.com/docs/en/epfz/6.1.0?topic=control-hexadec-ieee-attributes */
   FloatFormat,
   /** @see https://www.ibm.com/docs/en/epfz/6.1.0?topic=control-initial-attribute */
+  InitAcross,
```
Lotes, line 84: Add a doc line so that one can distinguish the meanings.

packages/language/src/syntax-tree/ast.ts
```diff
@@ -3120,19 +3106,32 @@ export function createLinkageOptionsItem(): LinkageOptionsItem {
   };
 }
 
-export interface Literal extends AstNode {
-  kind: SyntaxKind.Literal;
-  multiplier: Parenthesis | null;
-  value: LiteralValue | null;
+/**
+ * Repeated expressions are used in two ways:
```
Lotes, line 3110: Crazy PL/I + preprocessor

packages/language/src/parser/parser.ts
```diff
@@ -4498,9 +4499,95 @@ const writeStatementOption = rule(
...
+      tokens.CloseParen,
+    );
+    return element;
+  },
+);
+
 const initialAttribute = rule(
-  choice(sequence(tokens.INITIAL), sequence(tokens.INITACROSS)),
-  (state: ParserState): ast.InitialAttribute => {
+  choice(sequence(tokens.INITIAL)),
```
Lotes, line 4576: remove outer choice(...)

Lotes: ah I see... a manual lookahead... you could use a placeholder instead... I think I have something like this

Lotes: throwHasManualLookahead

packages/language/src/parser/parser.ts
```diff
@@ -6392,23 +6357,44 @@ const parenthesizedExpression = rule(
...
     return element;
   },
 );
 
+function repeatedExpressionLookahead(state: ParserState): boolean {
+  // First check for the syntax that indicates a repeated expression
+  if (!state.canConsumeFirst(expression.first())) {
+    return false;
+  }
+  // Next, prevent ambuities with wildcard expressions (*)
```
Lotes: ambiguities

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -1328,9 +1365,112 @@ function evaluateExpression(
...
+function unrollArrayValue(value: Value): Value[] {
+  if (isScalarValue(value)) {
+    return [value];
+  }
+  const result: Value[] = [];
+  for (const item of value.array) {
+    if (isScalarValue(item)) {
+      result.push(item);
+    } else {
+      result.push(...unrollArrayValue(item));
```
Lotes, line 1475: Can expressions be nested in an arbitrarily deep way?

(...)(4)(3)(2)0

This looked like there are only two levels of repetition...

Or is this meant to be:

((1,2,3),(4,5,6),(789))

? In order to define matrices..

msujew: Yes, you can do something like `(2)((3)"a", (4)"b")` and it will unroll to a flat array `[a, a, a, b, b, b, b, a, a, a, b, b, b, b]`. You can define matrices that way, because an array like `X(4, 4)` is the same (in memory) as `X(16)`.

packages/language/test/fourslash/parser/initial-wildcard-empty.ts
msujew: What do you mean exactly here?

Lotes: I think this can be done by the type system more or less: checking the constant values of an INITIAL (evaluated values as far as possible).
Could be useful for calculated values that are based on other constant values.

Maybe in a separate PR, if we want this.

packages/language/src/preprocessor/instruction-interpreter.ts
Lotes: Are there limits of nesting? Just asking with validations in mind. Can we allow any nesting or are there also forbidden combinations?
(2)(2)(2)(2)(2)(2)(2)(2)0 == array with 256 zeros?

You demonstrated the repetition operator. Here is what I remembered:
(8)0 == 8 zeros
(2)(8)0 == 2 arrays with 8 zeros each
(4)(2)(8)0 == ??? array of array of 8 zeros?
Or does it depend on the type of the variable declaration?

packages/language/test/fourslash/parser/initial-wildcard-empty.ts
msujew: Yeah, let's tackle this in a separate PR, if it makes sense.

packages/language/src/preprocessor/instruction-interpreter.ts
msujew: There is no real array nesting in PL/I, everything is unrolled in memory. Meaning that all are just flat arrays, independent of the syntax and nesting of the repetitions.

packages/language/src/typesystem/descriptions.ts
Lotes: I also meant the other three INITIAL attributes because they all have the same link.

Lotes (approved):
Thanks, looks good.

PR #650, opened 2026-03-19
Fix ambiguity quick fix action

Lotes:
Fixes #636 

There are two cases according to the nesting of a declare statement:

Case 1:

```
DCL X CHAR(300);
 DCL CT BIN FIXED(31);
 DCL 1 A BASED(ADDR(X)),
       5 B BIN FIXED(31),
       5 C (CT REFER(B)),
         10* CHAR(1);
 DCL 1 AA BASED(ADDR(X)),  //splitted into two or more statements
       5 B1,
         10 B BIN FIXED(31),
       5 B2,
         10 B BIN FIXED(31), 
        5 C (CT REFER(B));
```

Case 2:

```
 DCL X CHAR(300);
 DCL CT BIN FIXED(31);
 DCL 1 A BASED(ADDR(X)), //everything in one statement
       5 B BIN FIXED(31),
       5 C (CT REFER(B)),
         10* CHAR(1),
     1 AA BASED(ADDR(X)),
       5 B1,
         10 B BIN FIXED(31),
       5 B2,
         10 B BIN FIXED(31), 
        5 C (CT REFER(B));
```

In case 2 we need to take care of the original structure, where the quickfix action or reference is coming from.

PR #730, opened 2026-05-21
Remove ambiguity on DECLARE...REFER

Lotes:
The [documentation](https://www.ibm.com/docs/en/epfz/6.1.0?topic=attribute-refer-option-self-defining-data) says that the member expression is limited to the same STRUCTURE/UNION.

I added a special filter for this edge case.

Closes #696

### DIFF
--- a/packages/language/src/linking/resolver.ts
+++ b/packages/language/src/linking/resolver.ts
@@ -12,6 +12,7 @@
 import { Location, tokenToRange } from "../language-server/types";
 import { Token } from "../parser/tokens";
 import {
+  DeclaredItem,
   getContainer,
   iterateReferenceNodes,
   MemberCall,
@@ -281,7 +282,7 @@ function getMatchingSymbols(
 
   const getFullName = () => qualifiedName.toReversed().join(".");
 
-  const explicitlyDeclaredSymbols = scope
+  let explicitlyDeclaredSymbols = scope
     .getExplicitSymbols(qualifiedName, {
       /**
        * If the symbol is a procedure parameter, it is only permitted to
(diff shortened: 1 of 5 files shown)


msujew (commented):
I'm not actually sure this is the correct fix. I believe we're in general too eager to say that a reference is ambiguous. For example, in the following, the compiler will also not complain about ambiguous references:

```
 TEST: PROCEDURE;
   DCL DUPLICATE CHAR(10) INIT("HELLO");
   DCL 1 A,
      2 DUPLICATE CHAR(10) INIT("WORLD"),
      2 C CHAR(10) INIT(DUPLICATE); 
 END;
```

In the compiler output, I can see that the `INIT(DUPLICATE)` is actually pointing to the `2 DUPLICATE` definition (which is "more local"), instead of the `DCL DUPLICATE` definition.

Therefore, this does not only apply to the `REFER` part of the syntax. Can we try to build a more general fix for this?

Edit: This doesn't seem to be quite correct. The compiler is doing something weird...

packages/language/src/linking/resolver.ts
```diff
@@ -517,27 +527,55 @@ export function getTokenAt(unit: CompilationUnit, uri: URI, offset: number) {
...
+  if (declaredItem?.level !== 1) {
+    return null;
+  }
+  return declaredItem;
+}
+
+function tryExtractRootStructureIfBasedMember(
+  node: SyntaxNode,
+): DeclaredItem | null {
+  const referNode = getContainer(node, SyntaxKind.Bound);
```
msujew: Suggestion: This is on a hot-path which is called for **every** reference. Calling `getContainer` here is pretty slow, since the default case will be one in which the node is not part of a `Bound`, which means that is has to run the full container chain, until it simply returns `null`. This should be replaced with an optimized function based on the possible container types.

msujew (commented):
I believe we longer need the workaround introduced for https://github.com/zowe/zowe-pli-language-support/issues/636, right? That was specifically needed for `REFER`. See https://github.com/zowe/zowe-pli-language-support/pull/650

Also, please add some comments to the PR. `getRootCompositeNode` is rather difficult to parse.

packages/language/src/linking/resolver.ts
```diff
@@ -291,6 +295,15 @@ function getMatchingSymbols(
...
     .filter((symbol) => !symbol.isRedeclared); // Don't resolve reference to redeclared symbols.
 
+  if (reference.owner) {
+    const rootComposite = tryExtractRootStructureIfBasedMember(reference.owner);
+    if (rootComposite) {
+      explicitlyDeclaredSymbols = explicitlyDeclaredSymbols.filter((symbol) =>
+        isMemberOfComposite(symbol.node, rootComposite),
+      );
+    }
+  }
```
msujew, line 306: Suggestion: Add a comment here that this is for the `REFER` part of a declaration bound - I don't think this is obvious.

msujew (approved):
Thanks, looks good to me 👍

PR #731, opened 2026-05-21
Improve PL/I source file detection

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/697

Tries to narrow down the heuristic a bit, by first trying to exclude cases which are very likely not PL/I code.

Also improves the detection by running immediately when the extension starts. This ensures that we catch all open files as well.

### DIFF
--- a/packages/vscode-extension/src/extension/document-identification.ts
+++ b/packages/vscode-extension/src/extension/document-identification.ts
@@ -15,24 +15,71 @@ import { BaseLanguageClient } from "vscode-languageclient";
 
 export function registerPliDocumentIdentifier(lc: BaseLanguageClient) {
   const proposedFiles = new Set<string>();
+  // Some files might be opened before the extension starts up
+  for (const document of vscode.workspace.textDocuments) {
+    void checkFileType(proposedFiles, document, lc);
+  }
   return vscode.workspace.onDidOpenTextDocument(async (document) => {
-    if (!Settings.getInstance().autoDetect) {
-      // Auto-detection is disabled, so do not propose to set language for any file.
-      return;
-    }
-    if (proposedFiles.has(document.uri.toString())) {
-      // We've already proposed to change the language for this document.
-      return;
-    }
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/extension/document-identification.ts
```diff
@@ -15,26 +15,73 @@ import { BaseLanguageClient } from "vscode-languageclient";
 
 export function registerPliDocumentIdentifier(lc: BaseLanguageClient) {
   const proposedFiles = new Set<string>();
+  // Some files might be opened before the extension starts up
+  for (const document of vscode.workspace.textDocuments) {
+    void checkFileType(proposedFiles, document, lc);
```
Lotes: What is the effect of void here? Never seen this so far.

packages/vscode-extension/src/extension/document-identification.ts
```diff
@@ -15,26 +15,73 @@ import { BaseLanguageClient } from "vscode-languageclient";
...
+  // Listing files might contain PL/I code, but they are not PL/I source files
+  "lst",
+  "list",
+  // The compiler also outputs .xml files
+  "xml",
+]);
+
+function isNotPliDocument(document: vscode.TextDocument): boolean {
+  // Look for file types that are likely not PL/I source files
+  const baseName = document.uri.path.split("/").pop()?.toLowerCase() ?? "";
```
Lotes: Is there no UriUtils for the EXT extraction?

msujew: Good point, there is 👍

packages/vscode-extension/src/extension/document-identification.ts
msujew: It will _void_ the returned promise from `checkFileType`. We don't want to actually await the promise - voiding it makes that explicit. Though it's actually unnecessary here. I've removed it again.

Lotes (approved):

PR #733, opened 2026-05-28
Add recompile flag for compiler options and more fixes

ssmifi:
Several fixes related to https://github.com/zowe/zowe-pli-language-support/issues/716

This one looks larger than it is because of some reformatting. 

The fixes include 
- the addition of a `recompile` flag to the compiler options, which is used to calculate a fingerprint for the instruction cache. If one of these options change, the cache is cleared. 
  - Added `recompile` to a list of options that can/could influence compilation.
- changes to the plugin configuration also refreshes semantic tokens. (Otherwise the syntax highlighting is not updated correctly. You can test this by decreasing the right margin to truncate comment delimeters. The comment should then bleed into the next line.)
- Indicators do not send information for virtual files any longer. This resulted in additional "ghost" rulers.
- Use `onDidChangeVisibleTextEditors` instead of tracking open documents with `open` and `close` which does not necessarily correspond to the actual open tabs. Therefore, rulers can be shown for all open tabs and also remove ruler for tabs that are not open. However, note that rulers are set on language level... so still, if you have multiple tabs open with different margin settings, more than 2 rulers are shown.
- Use `workspace` settings for ruler settings. This is a suggestion to create more transparency. The client would only chance workspace settings and the settings that are changed (automatically) can actually be observed in `.vscode/settings.json`.
- Added some preliminary tests for the cache and virtual files. I was thinking about more sophisticated tests using the fourslash framework to check directly for cache hit/missed but decided that this is out-of-scope. This could be something for upcoming test framework changes though.

### DIFF
--- a/packages/language/src/language-server/connection-handler.ts
+++ b/packages/language/src/language-server/connection-handler.ts
@@ -424,6 +424,9 @@ export function startLanguageServer(
 
       // reindex reachable compilation units
       await compilationUnitHandler.reindex(connection, CancellationToken.None);
+
+      // refresh semantic tokens so syntax coloring updates immediately
+      connection.languages.semanticTokens.refresh();
     },
   );
   connection.onRequest(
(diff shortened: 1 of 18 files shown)


msujew (commented):
Thanks, makes sense to use a fingerprint instead of relying on the update of specific compiler options 👍

msujew (approved):

PR #735, opened 2026-05-29
Improve compiler option handling

ssmifi:
Additions and fixes w.r.t. the compiler option tests mention in https://github.com/zowe/zowe-pli-language-support/issues/734.

Summary of the fixes/additions
- NOAGGREGATE should not accept parameters.
- "Binary" should be "Hexadec" in AGGREGATE.
- **CHECK** Added check for duplicated abbreviations.
- **CODEPAGE** leading zero is not mandatory, but note that is is still stored as `CODEPAGE( 01140 )` in the compiler.
- **DEFAULT**
  - "NORECURSIVE" should be "NONRECURSIVE".
  - Added abbreviations.
- **RULES**
  - Fixed mutex options `COMPLEX` vs `NOCOMPLEX`.
  - Single suboptions do only expect one parameter. (confirmed on mainframe.) Note that suboptions with lists, such as `NOLAXINOUT`, allow arbitrary many entries in their lists, so `*PROCESS RULES(NOLAXINOUT(ALL LOOSE STRICT SOURCE));` is valid here.
  - Fixed `NOUNSET`, but note that even sub-options have defaults and are valid without option on the mainframe. so `*PROCESS RULES(NOUNSET);` is valid and will  result in `NOUNSET( ALL )`. Note that `*PROCESS RULES(NOUNSET());` is NOT valid.
  - Fixed `UNREFFILE` should be `UNREFDEFFILE`.

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/codes.ts
+++ b/packages/language/src/preprocessor/compiler-options/codes.ts
@@ -138,7 +138,7 @@ export const CompilerOptionsCodes = {
       code: "COAG01",
       severity: Severity.E,
       message: (value: string) =>
-        `Expected "DECIMAL" or "BINARY", but received '${value}'.`,
+        `Expected "DECIMAL" or "HEXADEC", but received '${value}'.`,
     },
   },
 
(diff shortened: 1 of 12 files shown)


msujew (approved):
Thanks, looks good to me 👍

Issue #668, opened 2026-04-09
Configuration Management Enhancement


asatklichov:
This enhancement is required to be completed before  ['Remote resolution of include members (copybooks)'](https://github.com/zowe/zowe-pli-language-support/issues/629)

Currently we have .pliplugin folder with two config files. It is created (via asking pop-up dialogue or quick fix suggestion) during a user interaction (opening fresh new file, or later using a %include ABC) copybook. 

We keep it as it is. 

Why we need an enhancement?  Because source files opened beside workspace (externally from disk or remotely by Zowe Explorer API) can not find configuration files. 

**Solution:**

- [x]  1)   Create  (via asking pop-up dialogue or quick fix suggestion) a VSCode user settings with same content as in .pliplugin folder files as in below code snippet, based  on user interaction.  So, if user opens a source file in VSCode from  disk or remotely by Zowe Explorer it should create a configuration in VSCode user settings.  

 ```
"pli.pgm_conf": {
  "pgms": [
    {
      "program": "test.pli",
      "pgroup": "default"
    }
  ]
},
"pli.proc_grps": {
{
  "pgroups": [
    {
      "name": "default",
      "compiler-options": [],
      "libs": [
                 "inc", 
                "C:\\copybooks\\ca11\\AZAT.PL1.APP.LIB",
                "zowe-ds:/profilename/AZAT.PL1.ZAPP.LIB"
      ],
      "member-name-validation": false,
      "include-extensions": [
        ".pli",
        ".pl1",
        ".inc"
      ],
      "lsp-options": {
        "check-margins": true
      }
    }
  ]
}
```

- [x]  2) If user opens a source in workspace logic stays same (currently .pliplugin folder created with two configs). 
But, still configuration should be merged in memory (config from .pliplugin folders and frokm VSCode user settings)

- [x]  3) For multiple workspace case, each workspace has its own .pliplugin folder  with two config files. 
And still configuration should be merged in memory (config from .pliplugin folders and frokm VSCode user settings)

- [x]  4) Update current .pliplugin  proc_grps file for "libs" values, so we can have a 3 examples  

- [x]  5)  Note that remote member resolution should follow the URI rule which as below.
`"zowe-ds:/activezoweprofile/dataset-qualified-name"    `

- [x]  6)  Provide below error message on existing  `"zowe-ds:" ` schema
 
"Plugin Configuration failed to resolve library entry 'zowe-ds:/azat/IBMUSER.SRC2'.  
[...]

PR #736, opened 2026-06-01
Support pliplugin config from vscode settings

### LINKED ISSUE #690: Improve typing of custom LS messages
We currently have a few custom language server<->client messages that carry arguments/return types. Those aren't properly annotated yet. We should improve that.

msujew:
Related to (1) and (2) of https://github.com/zowe/zowe-pli-language-support/issues/668.

Closes https://github.com/zowe/zowe-pli-language-support/issues/690, since I needed those to properly implement the message handling.

Enables the language server to query vscode settings to populate the plugin settings. While we could simply use the language server configuration object, this change performs this more in depth:

Instead of simply reading the vscode settings, our language server reads the actual settings files of vscode. This is done so that the language server can then provide the diagnostics on the vscode settings. Note that this obviously only works in vscode-like editors. Other lsp clients can only use the pliplugin directory.

Some of the tests become easier to handle once https://github.com/zowe/zowe-pli-language-support/pull/720 is merged - I'll keep the PR open (?) or provide a follow up PR to change the tests into the fourslash format.


scripts/generate-config-schema.mts
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env npx tsx
```
Lotes: A different approach, that I have used in free-time projects, is to generate the JSON schemas from a Zod type definition.

This has the advantage that you can immediately use the type definition for validations. The Schema files you normally use in VS Code so that the editor serves nice auto-completion. There are different generation methods depending on which Schema version you need.

You do not have to do so, just wanted to mention it. Maybe it makes sense in your eyes or not.

packages/vscode-extension/src/extension/messages.ts
Lotes, line 1: Question: Why is this necessary? It redirects to the original method. Now the call looks a bit different. What is the benefit now?

packages/vscode-extension/src/extension/config-loader.ts
```diff
@@ -0,0 +1,153 @@
...
+ * Computes the user-scope `settings.json` URI from an extension's
+ * `globalStorageUri`. VS Code's `globalStorageUri` always resolves to
+ * `<userdata>/User/globalStorage/<extensionId>` (on every platform and
+ * in remote/web), so going up two levels gives us `<userdata>/User`,
+ * and that's where `settings.json` lives.
+ */
+export function deriveUserSettingsUri(
+  globalStorageUri: vscode.Uri,
+): vscode.Uri {
+  return vscode.Uri.joinPath(globalStorageUri, "..", "..", "settings.json");
```
Lotes, line 120: Would it make sense to have a single file PATHS.TS where we collect such system/extension paths?

packages/language/test/lsp/apply-quick-fixes.test.ts
```diff
@@ -459,6 +458,7 @@ describe("applyQuickFixes", () => {
     const result = await applyQuickFixes.applyQuickFixes(
       diagnostics,
       workspace,
+      "",
```
Lotes, line 461: Is it really ok to pass an empty string?
It makes one think that this is a normal usage.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -304,66 +370,210 @@ export class PluginConfigurationProvider {
...
-      UriUtils.joinPath(workspaceUri, ".pliplugin", "pgm_conf.json"),
+
+    this.programConfigs.clear();
+    this.processGroupConfigs.clear();
+    this.procGrpsSnapshots.clear();
+    this.knownPgmConfUris.clear();
+    this.knownProcGrpsUris.clear();
+
+    const diagnostics: PluginConfigDiagnostics = new Map();
+    const ensureList = (uri: string): LspDiagnostic[] => {
```
Lotes: Proposal: Rename to "ensureDiagnosticsList"

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -304,66 +370,210 @@ export class PluginConfigurationProvider {
...
+    ensureList(plipluginProcGrpsUri.toString());
+
+    // Collect sources in PRECEDENCE-LOWEST-FIRST order. Map-based merge
+    // means later writes overwrite earlier writes, so listing
+    // .pliplugin/ second makes it the winning source on key collisions.
+    const global = await this.fetchGlobalSettings();
+    const pgmSources: ConfigSource[] = [];
+    const procGrpsSources: ConfigSource[] = [];
+
+    if (global?.pgmConf) {
```
Lotes: You could split this function. It is quite long.
Like having Setup/Gathering/Merge phases for example.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -304,66 +370,210 @@ export class PluginConfigurationProvider {
...
+        undefined,
+      );
+    } catch (err) {
+      console.error("Failed to fetch global plugin configuration:", err);
+      return undefined;
+    }
+  }
+
+  /** Reads & wraps a settings-side {@link Messages.GlobalConfigEntry}. */
+  private async readSource(
```
Lotes: Rename to readConfigSource? Or something more specific :)

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -304,66 +370,210 @@ export class PluginConfigurationProvider {
...
+    this.libFileMatchers = undefined;
+
+    // Make sure URIs we published last time but are no longer sources
+    // come back with an empty list, so the editor clears those diagnostics.
+    for (const uri of this.previouslyPublishedUris) {
+      ensureList(uri);
+    }
+    this.previouslyPublishedUris = new Set(diagnostics.keys());
+
+    cancel();
```
Lotes: Why the cancel at the end? A bit confusing.

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -304,66 +370,210 @@ export class PluginConfigurationProvider {
...
+    }
+    const plipluginPgmSource =
+      await this.readPlipluginSource(plipluginPgmConfUri);
+    if (plipluginPgmSource) pgmSources.push(plipluginPgmSource);
+    const plipluginProcSource =
+      await this.readPlipluginSource(plipluginProcGrpsUri);
+    if (plipluginProcSource) procGrpsSources.push(plipluginProcSource);
+
+    // Parse pgm_conf sources and merge by resolved program URI.
+    const mergedPrograms = new Map<string, ProgramConfig>();
```
Lotes, line 469: Is it doable to split the two merges in two different functions? Or to interleave them by creating steps of functions:

init1
init2
next1
next2...

This would improve the readability here, I hope!

scripts/generate-config-schema.mts
msujew: Good point - though I would consider this in a separate PR 👍

packages/vscode-extension/src/extension/messages.ts
msujew: Type safety, see also https://github.com/zowe/zowe-pli-language-support/issues/690. Doing this ensures that the parameters and return types of the `NotificationType` and `RequestType` are properly respected by the caller and receiver side.

packages/vscode-extension/src/extension/config-loader.ts
msujew: Where would we place this? I'm not entirely sure, since some of these paths are very specific to the issue (in this case - we need to find the user settings by traversing the file system).

packages/language/test/lsp/apply-quick-fixes.test.ts
msujew: It is a test - it would tell us, if it's not ok ;)

packages/vscode-extension/src/extension/config-loader.ts
Lotes: Somewhere at the source root of the vscode-extension package.

If it is very problem-specific, we can avoid it. A "settings.json" sounds quite important.

Again, it was more a habit from a free-time project to collect paths (and maybe the file JSON type) at a central point. Does not need to make sense here. Just thinking loud...

packages/language/test/lsp/apply-quick-fixes.test.ts
Lotes: Tests "can" also document usage.

Having an empty string as last argument seems like the parameter is or should be optional, or does not always need to have a meaningful usage internally.

msujew: The last parameter used to be optional, but for "real usage" cases (i.e. those that produce meaningful output), it was actually mandatory.

packages/language/src/workspace/plugin-configuration-provider.ts
msujew: It's really hard to split this up - too many side effects (i.e. diagnostics collecting, URI collecting, etc.). Splitting it up actually leads to more code. This requires a more throughout refactoring of the code to be honest. I could do this as part of this PR. But doing so right now would lead to a lot of merge conflicts with other PRs that are related to this code (mainly #720).

I would do this in a separate PR. The diagnostics collection of the plugin config should be refactored anyway.

packages/language/src/workspace/plugin-configuration-provider.ts
msujew: We cancel at the end ... because we have to cancel at the end. We call `const cancel = startLongRunningOperation(...)` earlier in order to notify the user that a (potentially) long running, blocking operation is running. This call cancels the notification and tells the user that the language server is ready for more input.

packages/language/src/workspace/plugin-configuration-provider.ts
msujew: Good point, I was able to reduce the complexity here a bit 👍

packages/language/src/workspace/plugin-configuration-provider.ts
Lotes: Ah interesting, never seen so far. I would assume a STOP with regards to the terminology. CANCEL sounds like a ROLLBACK instead of a commit xD. But ok, when that is the pattern here, it is like this..

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -421,37 +414,38 @@ export class PluginConfigurationProvider {
...
-    const plipluginPgmSource =
-      await this.readPlipluginSource(plipluginPgmConfUri);
-    if (plipluginPgmSource) pgmSources.push(plipluginPgmSource);
-    const plipluginProcSource =
-      await this.readPlipluginSource(plipluginProcGrpsUri);
-    if (plipluginProcSource) procGrpsSources.push(plipluginProcSource);
+    // Read every source concurrently. Precedence (settings < .pliplugin/)
+    // is established by the order we assemble each array below, not by the
+    // order the reads resolve, so reading in parallel is safe.
+    const [globalPgm, globalProc, plipluginPgm, plipluginProc] =
```
Lotes, line 453: Much better here, thanks.

Wagner-Laranjeiras (approved):

package.json
```diff
@@ -24,7 +24,8 @@
     "watch": "tsc -b tsconfig.build.json --watch",
     "license:check": "tsx ./scripts/check-license-headers.mts",
     "license:fix": "tsx ./scripts/check-license-headers.mts --fix",
-    "extract-compiler-tests": "tsx ./scripts/extract-compiler-tests.mts"
+    "extract-compiler-tests": "tsx ./scripts/extract-compiler-tests.mts",
+    "generate-config-schema": "tsx ./scripts/generate-config-schema.mts"
```
Lotes: Include the script in `build:full`?

packages/language/src/utils/messages.ts
```diff
@@ -9,21 +9,76 @@
...
   export const WorkspaceDidChangePluginConfigNotification =
-    "workspace/didChangePluginConfig";
+    createNotificationType<void>("workspace/didChangePluginConfig");
 
   /**
    * Request sent to the LS to check if a file is already present in the workspace (i.e. included in a program)
    */
-  export const ExistingFileRequest = "pli/existingFileRequest";
+  export const ExistingFile = createRequestType<string, boolean>(
+    "pli/existingFile",
```
Lotes, line 48: You have to be careful with renaming: `pli/existingFileREQUEST` is still present in `packages/vscode-extension/src/extension/document-identification.ts#L102`.

packages/vscode-extension/package.json
```diff
@@ -122,6 +126,115 @@
           ],
           "default": "automatic",
           "description": "Controls the rulers for the margin compiler option. 'off' disables rulers, 'default' shows the PL/I default rulers [2, 72], 'automatic' adds rulers based on margin compiler options."
+        },
+        "pli.pgm_conf": {
```
Lotes, line 130: Generating the `package.json` feels (a bit) wrong. At least because we then reformat the whole file on every generation.
Also comments would be stripped out (if we would use them, but as far as I know, comments are actually not allowed in JSON files... there are some exceptions like the tsconfig, it seems...).
I do not know an easy fix. Why do we need to put the whole schema in here? Can we not put a schema in here that refers to another schema file? Then I could sleep better ^^*

Lotes: You normally do a `pnpm install` which in our case renews one of our `package.json`, which is somehow the source of dependencies for `pnpm install` (actually it is the lockfile for CI). The JSON file changes do not affect the install really, but would like to avoid such cycles. A `package.json` should not be a (half-)generated file.

msujew: > Why do we need to put the whole schema in here? Can we not put a schema in here that refers to another schema file? Then I could sleep better ^^*

I would love to, but VSCode doesn't allow that.

msujew: Nevermind, there seems to be a workaround - although a fairly weird one:

https://github.com/eclipse-che4z/che-che4z-lsp-for-hlasm/blob/432e72a500c16cd6fded8e6477552c8519b67f5e/clients/vscode-hlasmplugin/package.json#L451-L466

Lotes: Oh nice!(?) I hope that is applicable to the current file system...

package.json
msujew: Removed the script - no longer required.

packages/language/src/utils/messages.ts
msujew: Thanks 👍

packages/vscode-extension/package.json
msujew: I was able to implement it. It looks a bit weird, but works pretty well 👍

Lotes: Great!

Lotes (approved):
Looks great, thanks!

PR #737, opened 2026-06-01
Experimental: Fourslash LS integration

ssmifi:
This is the branch where I experimented with integrating the LS into the fourslash framework some time ago, mainly to address coverage concerns as mentioned in https://github.com/zowe/zowe-pli-language-support/pull/720. 
Following the recent changes to the test framework, the implementation would need some additional polishing.
For now, I am sharing it so that we can discuss the approach internally and decide whether it is worth pursuing further.

The core idea is probably best illustrated by [`packages/language/test/fourslash/server/lsp-code-actions-ambiguous.ts`](https://github.com/zowe/zowe-pli-language-support/pull/737/changes#diff-c018d4a883b069b5122b3805c63eea95d3368f80639cff5cc2e8d9be079c9fd1).
Essentially, the fourslash interface gains a new `server` category whose commands are routed through the LS, while keeping the changes to the `test-builder` relatively small.
On the branch, I experimented with a broad range of requests, but for an initial adoption we could focus on specific use cases, such as code actions. 

Feel free to ping me if you'd like to discuss it further.

### DIFF
--- a/packages/language/test/fourslash-harness/execute.test.ts
+++ b/packages/language/test/fourslash-harness/execute.test.ts
@@ -199,7 +199,14 @@ function runSingleHarnessTest(filePath: string, timeout = 10_000) {
       const implementation =
         createTestBuilderHarnessImplementation(testBuilder);
 
-      await runHarnessTest(testFile, implementation);
+      try {
+        await runHarnessTest(testFile, implementation);
+      } finally {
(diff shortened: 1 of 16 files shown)


ssmifi:
We won't implement this during this milestone. It's still documented here.

PR #738, opened 2026-06-01
Migrate CICS options validation

Lotes:
Migrated 104 CICS options validation files from [here](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/tree/development/server/engine/src/main/java/org/eclipse/lsp/cobol/implicitDialects/cics/utility) using an AI agent.

Reviewing all "check-*" and test files might be too much. The task is very repeating. There are some unique files like "base.ts" or "options-registry.ts", that are more interesting.

Note: Some test just check the counts of diagnostics (0 or 1). These are repetitive tests to achieve a better test coverage.
Note II: This is not the entire Java logic. There will be a follow up PR that handles the remaining parts.


packages/preprocessor-cics/src/checks/base.ts
```diff
@@ -0,0 +1,772 @@
...
+   * Helper method to collect analysis errors if the rule context does not contain mandatory options
+   *
+   * @param rules Generic list of rules to check. Will either be a collection of ParserRuleContext
+   *     or TerminalNode
+   * @param ctx Context to extrapolate locality against
+   * @param options Options checked to insert into error message
+   * @return true if mandatory option found
+   */
+  protected checkHasMandatoryOptions(
+    rules: any[],
```
Lotes: Try to use better types

packages/preprocessor-cics/src/engine/collect-semantic-errors.ts
```diff
@@ -0,0 +1,368 @@
...
+  constructor() {
+    super();
+    this.cicsOptionsCheckUtilityParams = this.getCheckParams();
+    this.cicsOptionsCheckUtility = new OptionsRegistry(
+      this.errors,
+      this.cicsOptionsCheckUtilityParams,
+    );
+  }
+
+  // override visitQualifiedDataName(ctx: QualifiedDataNameContext): ParseTree[] {
```
Lotes: Check commented out code

packages/preprocessor-cics/src/engine/collect-syntax-errors.ts
```diff
@@ -27,14 +27,11 @@ import {
...
-  ParseTree,
 } from "antlr4ng";
-import { CICSParserVisitor } from "../generated/CICSParserVisitor";
-import { CicsWordContext } from "../generated/CICSParser";
-import { ParseError, SemanticsKind, Token } from "preprocessor-api";
+import { Diagnostic, Severity } from "preprocessor-api";
 
-export class CollectingErrorListener extends BaseErrorListener {
-  public readonly errors: ParseError[] = [];
+export class CollectingSyntaxErrorListener extends BaseErrorListener {
```
Lotes, line 33: Repeating listener code. May move to API package?

.gitignore
```diff
@@ -19,4 +19,6 @@ coverage/
 **/pli.merged.json
 MISSING-TOKENS.md
 packages/preprocessor-cics/src/generated/
-packages/preprocessor-db2/src/generated/
\ No newline at end of file
+packages/preprocessor-db2/src/generated/
+packages/preprocessor-cics/src/antlr/.antlr
```
Lotes, line 22: Add an ignore for db2 ANTLR files

Lotes: This is for ignoring the auto generated Java files (enabled by some VSCode extensions).

packages/preprocessor-cics/src/engine/collect-syntax-errors.ts
```diff
@@ -48,26 +45,8 @@ export class CollectingErrorListener extends BaseErrorListener {
       message: msg,
       startOffset: offendingSymbol?.start || 0,
       endOffset: offendingSymbol?.stop || 0,
+      severity: Severity.Error,
+      code: "cics.syntax.error",
```
msujew, line 49: Issue: I noticed that this code isn't propagated. This is because [handleDiagnostics](https://github.com/zowe/zowe-pli-language-support/blob/bf50e1104e6002fd63cc65bc300e3aa60d3e4db5/packages/language/src/parser/exec-parser.ts#L135-L140) calls `ParserState#error` and not `ParserState#diagnostics.push`. `ParserState#error` will omit a bunch of data and set the `error = true` state on the parser state, which isn't correct. Please use that the other one instead.

This also prevents the `startOffset` and `endOffset` to correctly propagate, thus showing the error always on the `;` after the `EXEC` statement.

msujew (commented):
Looks good for the most part - I've noticed an issue in #728 which results in incorrect diagnostic propagation.

msujew (approved):
Thanks, works well!

PR #741, opened 2026-06-03
Remove IBM1059 select validation

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/700

As far as I could tell by testing on the compiler, the compiler does never show this message. Since we cannot look into the compiler code to find the actual trigger code, we have to remove it for now.


ssmifi (approved):
Sounds reasonable. 
Should we add a brief comment to the `IBM1059` code about the findings that links to this issue, so that it will not be added again later by accident?

PR #742, opened 2026-06-03
Harmonize playground and web extension

### LINKED ISSUE #370: Harmonize Playground + Web Extension
We should harmonize the implementation of the playground & the web extension, so that ideally the playground is utilizing the web extension directly or leveraging its exact LC + LS implementation. This will help ensure we can keep the two implementations in sync, and remove the duplicate code that we have in place for both.

Between the two we should take the web-extension state being updated in #360, and use that in the current playground.

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/370

Updates our playground to use version `^10` of the `monaco-languageclient` library. Since we can now load the PL/I vscode extension directly in the frontend, this change removes a lot of duplication that was previously necessary to support the same feature set in vscode+playground.

This change can be tested by using the generated playground available over at [here](https://zowe.github.io/zowe-pli-language-support/pr/pr-742/).


packages/playground/src/app.ts
```diff
@@ -10,87 +10,47 @@
...
-} from "./file-system-provider.js";
-import { Builtins, BuiltinsUri } from "pli-language";
-
-let wrapper: MonacoEditorLanguageClientWrapper | undefined;
 
 export async function startClient() {
   try {
     redirectOutlineCancelReporting();
+    registerButtons();
     const config = await configure(
```
ssmifi, line 30: _Minor_: `config` is not used anylonger, right?

packages/playground/src/config.ts
```diff
@@ -15,161 +15,92 @@ import getKeybindingsServiceOverride from "@codingame/monaco-vscode-keybindings-
...
-      },
+    advanced: {
+      enableExtHostWorker: true,
     },
+    monacoWorkerFactory: configureDefaultWorkerFactory,
   };
 
   return {
-    wrapperConfig,
+    vscodeApiConfig: vscodeApiConfig,
```
ssmifi: _Minor_: Could just use the shorthand.

packages/playground/src/workspace.ts
```diff
@@ -57,7 +78,7 @@ export function registerButtons() {
  * @param content - The content to share is either a plain string or an array of workspace files.
  * @param options - The options to share. clearWorkspace is used to clear the workspace before sharing.
  */
-async function share(content: string | WorkspaceFile[]): Promise<void> {
+async function share(content: string | SharedWorkspace): Promise<void> {
```
ssmifi, line 106: _Minor_: JSDoc seems to be outdated.

packages/playground/src/workspace.ts
```diff
@@ -98,33 +123,45 @@ export async function handleSharedWorkspace(
         ?.replace(/[\\\/]/g, "")
         .replace(/^\.+/g, "") ?? "example.pli";
     const content = decompressFromEncodedURIComponent(encodedContent);
-    return await fileSystemProvider.addFileToWorkspace(
-      `/workspace/${filename}`,
-      content,
+    const uri = vscode.Uri.file(`/workspace/${filename}`);
+    await fileSystemProvider.writeFile(
+      vscode.Uri.file(`/workspace/${filename}`),
```
ssmifi: _Minor_: Just use `uri` here.

packages/playground/src/workspace.ts
```diff
@@ -98,33 +123,45 @@ export async function handleSharedWorkspace(
...
+        path: uri.path.substring(0, uri.path.lastIndexOf("/")),
+      });
+      try {
+        await fileSystemProvider.mkdir(parentUri);
+      } catch {
+        // Ignore if the directory already exists.
+      }
+      await fileSystemProvider.writeFile(
+        uri,
+        new TextEncoder().encode(file.content),
```
ssmifi: _Minor_: `encoder` was declared above and is only used in `loadDefaultWorkspace`. Can we also use it here? Or make encoder local to `loadDefaultWorkspace`.

packages/playground/src/workspace.ts
```diff
@@ -183,3 +228,73 @@ export function redirectOutlineCancelReporting() {
     }
   });
 }
+
+import helloWorld from "../workspace/hello-world.pli?raw";
+import includeExample from "../workspace/include.pli?raw";
+import includedExample from "../workspace/lib/included.pli?raw";
+import pgmconf from "../workspace/.pliplugin/pgm_conf.json?raw";
+import procgrps from "../workspace/.pliplugin/proc_grps.json?raw";
```
ssmifi: Move imports to the top of the file?

packages/playground/src/workspace.ts
```diff
@@ -98,33 +123,45 @@ export async function handleSharedWorkspace(
...
+      fileOptions,
     );
+    return uri;
   }
 
   // Load the workspace files.
   const encodedWorkspace = url.searchParams.get("workspace");
   if (encodedWorkspace) {
-    const workspaceFiles: WorkspaceFile[] = JSON.parse(
+    const workspaceFiles: SharedWorkspace = JSON.parse(
```
ssmifi: _Question_: Can this fail? Should we guard this?

packages/playground/src/workspace.ts
```diff
@@ -98,33 +123,45 @@ export async function handleSharedWorkspace(
...
-        file.uri,
-        file.content,
+      const uri = vscode.Uri.file(file.uri);
+      const parentUri = uri.with({
+        path: uri.path.substring(0, uri.path.lastIndexOf("/")),
+      });
+      try {
+        await fileSystemProvider.mkdir(parentUri);
+      } catch {
+        // Ignore if the directory already exists.
```
ssmifi, line 177: _Question_: This catches all errors. Could there be an error, where we don't want to write the file afterwards?

packages/vscode-extension/src/extension/main-browser.ts
```diff
@@ -32,7 +44,9 @@ export function deactivate(): Thenable<void> | undefined {
   return undefined;
```
ssmifi: _Question_: Should we also dispose `settings` here similar to the node version? Return type looks also different: should/could this be `export async function deactivate(): Promise<void>`?

packages/vscode-extension/src/extension/main-browser.ts
```diff
@@ -48,7 +62,8 @@ function startLanguageClient(context: vscode.ExtensionContext): LanguageClient {
   const client = new LanguageClient("pli", "PL/I", clientOptions, worker);
   registerFileSystemProvider(client);
   registerProgressReporter(client);
+  registerCustomDecorators(client, settings);
   // Start the client. This will also launch the server
-  client.start();
+  await client.start();
```
ssmifi, line 69: _Question_: Should we also subscribe the resources here?

ssmifi (commented):
Nice consolidation. Thanks. I just have some minor questions.

packages/playground/src/workspace.ts
msujew: I don't think the in-memory file system provider can error in any other way.

ssmifi (approved):
Thank you!

PR #745, opened 2026-06-08
Add exec statements in macro preprocessor

ssmifi:
Fixes https://github.com/zowe/zowe-pli-language-support/issues/716 second part and allows the preprocessors to be run inside the macro preprocessor.

The change basically moves `parseExecStatement` into the preprocessor-parser and makes the required functions async.

### DIFF
--- a/packages/language/src/parser/parser-state.ts
+++ b/packages/language/src/parser/parser-state.ts
@@ -29,8 +29,7 @@ import {
 } from "../validation/pli-codes";
 import { tokenIdxToClass } from "./token-type-factory";
 import * as environment from "../workspace/environment";
-
-import { CompilerOptions } from "../preprocessor/compiler-options/options-pli";
+import { CompilerOptions } from "../preprocessor/compiler-options/options";
 
 export enum RecoveryResult {
   /**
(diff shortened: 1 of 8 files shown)


packages/language/src/parser/parser-state.ts
```diff
@@ -53,15 +53,21 @@ export class ParserState {
   readonly tokens: t.Token[];
   readonly diagnostics: Diagnostic[];
   readonly compilerOptions?: CompilerOptions;
+  readonly textDocument?: TextDocument;
```
msujew: Refer to https://github.com/zowe/zowe-pli-language-support/pull/692#discussion_r3235672452 on why this is not a good idea.

msujew (commented):
Theoretically, we can solve the immediate problem like this.

However, the more general problem is actually that our current architecture does not support something like this:

```
*PROCESS PP(MACRO SQL);

%DCL SOME_SQL_CODE INIT("INCLUDE SQLCA");
EXEC SQL SOME_SQL_CODE;
```

(Note: This works on the compiler)

In this sense, I would like the actual fix to split the parser preprocessor phase into those separate PP phases. Do you think this is feasible?

msujew (approved):
I believe it's alright to merge this in the current state. Can you create an issue about the splitting of the macro phases?

PR #746, opened 2026-06-08
Add preprocessor highlighting to the cics/db2 block

ssmifi:
Add the preprocessor highlighting to the whole block.


Lotes (approved):
LGTM

PR #748, opened 2026-06-09
Safeguard macro interpreter

msujew:
Improves our macro interpreter by safeguarding against unreasonable user-inputs:
1. Hard caps the amount of times an instruction can run at 50,000.
2. Prevents out-of-memory issues by using a huge `COPY` call by limiting output size to 100_000 characters.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -365,7 +365,7 @@ export interface InterpreterOptions {
   marginsProcessor: MarginsProcessor;
 }
 
-export const MAX_INSTRUCTION_COUNTER = 5000;
+export const DEFAULT_INSTRUCTION_LIMIT = 5000;
 export const MAX_INSTRUCTION_LIMIT = 50000;
 
 export async function runInstructions(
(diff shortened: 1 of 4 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -366,13 +366,24 @@ export interface InterpreterOptions {
 }
 
 export const MAX_INSTRUCTION_COUNTER = 5000;
+export const MAX_INSTRUCTION_LIMIT = 50000;
```
ssmifi, line 369: _Suggestion_: I find the naming somewhat confusing. What do you think about `DEFAULT_INSTRUCTION_LIMIT` and `MAX_INSTRUCTION_LIMIT`?

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -2536,11 +2545,17 @@ function copy(
...
     return defaultEmptyValue;
   }
   const repeatCount = valueToNumber(repetitions, 0) + plus;
-  if (repeatCount === 0) {
+  if (
+    // Invalid repeat count, emit empty
+    repeatCount <= 0 ||
+    // Safeguard against large outputs
+    // Could cause long processing times
+    value.value.length * repeatCount > 100_000
```
ssmifi: _Suggestion_: Add another constant for this.

ssmifi (approved):
Looks good. Just 2 small suggestions.

PR #749, opened 2026-06-11
Proposal: Split pp calls into individual phases

ssmifi:
This PR is a follow-up to https://github.com/zowe/zowe-pli-language-support/pull/745 and proposes to split the pp calls into individual phases. 

The PR mainly alters the way `pli-lexer` invokes the pps so that PP directives such as `*PROCESS PP(MACRO MACRO);` become possible. A pp phase strictly becomes a `Token[]` to `Token[]` transformation.
The pp phases still all use the same interpreter code, which currently seems to be more maintainable. 

This is probably not the last call here, but I'd appreciate it if we could talk about the direction.

### DIFF
--- a/packages/language/src/preprocessor/macro-phase.ts
+++ b/packages/language/src/preprocessor/macro-phase.ts
@@ -1,0 +1,171 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { TextDocument } from "vscode-languageserver-textdocument";
+import { preprocessorParse, StatementParser } from "../parser/parser-entry";
+import { ParserState } from "../parser/parser-state";
+import { includeAltStatement, statement } from "../parser/preprocessor-parser";
+import * as t from "../parser/tokens";
+import * as ast from "../syntax-tree/ast";
+import { CstNodeKind } from "../syntax-tree/cst";
+import {
+  CompilerOptionResult,
+  CompilerOptions as PliCompilerOptions,
+} from "./compiler-options/options";
+import { CompilerOptions } from "./compiler-options/options-pli";
+import { generateInstructions } from "./instruction-generator";
+import { runInstructions } from "./instruction-interpreter";
+import { createIncludeInstruction, InstructionNode } from "./instructions";
+import { MarginsProcessor } from "./pli-margins-processor";
+import { PhaseInput, PhaseResult, PreprocessorPhase } from "./pp-phase";
+
+/**
+ * Recognizes `%` preprocessor statements.
+ */
+function createPreprocessorHandler(
+  textDocument: TextDocument,
+): StatementParser {
+  return async (state) => {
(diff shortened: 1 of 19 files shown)


packages/language/test/fourslash/preprocessor/pp-phases/macro-generates-macro-twice.ts
```diff
@@ -0,0 +1,27 @@
...
+/// <reference path="../../framework.ts" />
+
+// PP(MACRO MACRO) runs the macro preprocessor twice, with the output tokens of the first
+// pass becoming the input tokens of the second (Token[] -> MACRO -> Token[] -> MACRO -> ...).
+//
+// First pass:  TEST  ->  % DCL Y FIXED ; % Y = 7 ; Y
+// Second pass: the generated macro code runs  ->  7
+
+////*PROCESS PP(MACRO MACRO);
+//// %TEST: PROC RETURNS (CHAR);
```
Lotes, line 21: Crazy, did not know that this is possible!

How about `*PROCESS PP(MACRO MACRO MACRO);`? Can we execute it 3 or more times? I am just curious :D...

Lotes: Nevermind I saw the other test.. very cool

packages/language/src/preprocessor/pli-lexer.ts
```diff
@@ -70,105 +77,180 @@ export class PliLexer {
...
+
+  for (const item of pp.items) {
+    switch (item.name) {
+      case PliCompilerOptions.PPItemName.MACRO:
+        // Each explicit MACRO item is its own pass, so e.g. PP(MACRO MACRO) runs the
+        // macro preprocessor twice. A second pass can expand macro code that the first
+        // pass generated.
+        hasMacroPhase = true;
+        phases.push(
+          new MacroPreprocessorPhase(compilerOptionsResult, marginsProcessor),
```
Lotes, line 226: Is the margins processor served in every phase? Is this wanted? Or can we even omit it? (I have no idea! Just a question!)

packages/language/src/preprocessor/macro-phase.ts
```diff
@@ -0,0 +1,171 @@
...
+    const output = await runInstructions(unit, uri, instructionResult, {
+      compilerOptions: this.compilerOptionsResult,
+      marginsProcessor: this.marginsProcessor,
+      createParseHandlers: (includeDocument) =>
+        createMacroHandlers(opts, includeDocument),
+    });
+
+    return {
+      tokens: output.all,
+      statements: [...statements, ...output.statements],
```
Lotes, line 142: statements, diagnostics and references are always collected in dedicated arrays.. is the end goal not the compilation unit itself?

Lotes (commented):
I have some questions:

msujew (approved):
Thanks, I think that works well 👍

packages/language/src/preprocessor/pli-lexer.ts
ssmifi: Since all PPs can include files, we cannot omit it right now. There are all using the same mechanism, right? But if this changes or I am mistaken for SQL or CICS, then we could restrict this.

packages/language/src/preprocessor/macro-phase.ts
ssmifi: True. The design here is to keep the phases mostly isolated. But we could maybe optimize for performance here if this becomes a bottleneck.

PR #751, opened 2026-06-11
Add warnings for dup/mutex suboptions in rules directive

ssmifi:
Addition to https://github.com/zowe/zowe-pli-language-support/issues/734

Adds duplication and mutex warnings to suboptions of the RULES directive.


msujew (approved):
Thanks, looks good to me 👍

PR #752, opened 2026-06-11
Add more CICS positive unit tests

Lotes:
Progress 347/347 commands covered (due to the internal Excel file). The first 100 I noted down every option at least once. The remaining commands are tested in at least 2 test rows as requested by Azat.

Also: the first 100 rows have additional fourslash tests that can be run against the mainframe. I tested and debugged them all until all warnings and errors were gone.


vitest.config.ts
```diff
@@ -20,6 +20,14 @@ export default defineConfig({
     deps: {
       interopDefault: true,
     },
+    watchTriggerPatterns: [
+      {
+        pattern: /packages\/preprocessor-cics\/src\/test\/positives\.txt$/,
+        testsToRun: () => {
+          return ["packages/preprocessor-cics/src/test/positives.test.ts"];
+        },
+      },
```
msujew, line 30: Minor: Can you also add the `fourslash/**/*.ts` files to rerun the fourslash test suite? Right now, we don't rerun those when the test files change.

msujew (approved):
Thanks, makes sense 👍

vitest.config.ts
Lotes: Yup, makes also sense. I will give it a try

Lotes: Done, thanks for the idea.

PR #755, opened 2026-06-16
Add builtins LP int types

### LINKED ISSUE #53: Respect LP(32/64) Flags for __SIGNED_INT, __UNSIGNED_INT
The `__SIGNED_INT` & `__UNSIGNED_INT` built-ins are parametric in PL/I, following the LP(32) or LP(64) compiler flags (default is LP(32)). These will need to be setup to change the built-in based on the compiler options, whilst still falling back to the LP(32) default.

ssmifi:
Closes https://github.com/zowe/zowe-pli-language-support/issues/53

Add the builtin int type definitions according to the LP compiler option setting. 
Since the type definition can only be added after the compiler option extraction, I moved the whole builtin unit addition out of the initialization. (In theory it can happen directly after the tokenizing step,  but I thought closer to the symbol table is actually more appropriate. But we can change that if desired.)
Reuses the scope parent mechanism to load the correct builtins into the compilation unit.

### DIFF
--- a/packages/language/src/workspace/builtins.ts
+++ b/packages/language/src/workspace/builtins.ts
@@ -12,6 +12,7 @@
 import { TextDocument } from "vscode-languageserver-textdocument";
 import { UriUtils } from "../utils/uri";
 import { BuiltinsUriSchema } from "./builtins-constants";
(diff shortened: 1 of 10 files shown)


msujew (approved):
Thanks, looks good to me 👍

PR #756, opened 2026-06-17
Improve and extend documentation

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/743
Closes https://github.com/zowe/zowe-pli-language-support/issues/722

Adds documentation for the big, important parts of the language server. If you find anything wrong with it, please let me know. These are living documents, and they will be expanded as times goes on.


docs/CICS-KEYWORDS.md
Lotes, line 1: Thx.

ssmifi (approved):
LGTM

docs/ARCHITECTURE.md
Lotes, line 1: Some pictures would be nice, like UML diagrams, especially class diagrams for concepts and how they are wired together. Just saying this, because reviewing so much text in a short time is exhausting.
And: a picture says more than thousand words.

ssmifi: Good point. You could use AI to create and add mermaids to the markdown that depict the flow already described in the text.

msujew: I've added some diagrams. Thanks for the suggestions 👍

PR #757, opened 2026-06-17
[Improvement] Wildcard support for libs - expand Fourslash

### LINKED ISSUE #628: `libs` in proc_grps.json does not accept wildcards
neither of the following seems to work

```json
{
  "pgroups": [
    {
      "name": "default",
      "libs": [
        "inc",
        "*/inc",
      ],
    }
  ]
}
```

```json
{
  "pgroups": [
    {
      "name": "default",
      "libs": [
        "inc",
        "**/inc",
      ],
    }
  ]
}
```

Wagner-Laranjeiras:
## Support wildcards in `proc_grps.json` `libs`

Closes #628.

`libs` entries now accept glob patterns (`*/inc`, `**/inc`, `example/*/inc`)
in addition to literal paths. Each pattern is expanded against the workspace
into the concrete directories it matches, which then flow through the existing
lib-resolution logic unchanged.

### Notes
- `*` matches a single path segment; `**` matches any depth (including zero).
- Matching is case-insensitive, consistent with the rest of include resolution.
- A pattern that matches nothing is reported as an unresolved entry (COPC01E),
  same as a literal lib that doesn't exist.

### Tests
Adds fourslash coverage for single-level, globstar, literal-prefix, overlap
with a literal lib, case-insensitivity, and unresolved/partial-match
diagnostics.


README.md
msujew, line 1: Suggestion: I wouldn't touch this file, this is managed by the technical writers.

msujew (approved):
Thanks, looks good! Just one comment, see below.

packages/language/test/test-builder.ts
```diff
@@ -289,6 +307,7 @@ export class TestBuilder extends AbstractTestBuilder {
         procGrpsUri = true;
       }
     }
+    this.suppliedProcGrps = procGrpsUri;
```
Lotes: Sound like ProcGroups that were supplied, so a set of objects. Not a boolean.
Rename to (are|were)ProcGrpsSupplied.

packages/language/test/fourslash/preprocessor/include/include-resolution-order.ts
Lotes, line 57: crazy edge case!

packages/language/test/fourslash/preprocessor/include/ddmember-include-recursive.ts
```diff
@@ -9,10 +9,25 @@
...
+////         {
+////             "name": "default",
+////             "libs": [
+////                 "cpy/**"
+////             ]
+////         }
+////     ]
+//// }
+
 // @filename: cpy/MYLIB(m1)
```
Lotes, line 31: what function do MYLIB and A.B.C have here?

packages/language/src/config/lib-expander.ts
```diff
@@ -188,44 +217,89 @@ export async function expandLib(
...
+  pattern: string,
   fs: FileSystemProvider,
-): Promise<LibsDirEntry[]> {
-  rootLib = UriUtils.normalizePath(rootLib);
-  const entries: LibsDirEntry[] = [];
-  const queue: { lib: string; uri: URI }[] = [{ lib: rootLib, uri: rootUri }];
+  workspace: URI,
+): Promise<string[]> {
+  let basePrefix: string | undefined;
+  const normalizedPattern = pattern.replace(/\\/g, "/");
```
Lotes: This normalization step is somewhere given under UriUtils. Can you reuse it?

packages/language/src/config/lib-expander.ts
```diff
@@ -188,44 +217,89 @@ export async function expandLib(
...
+ * workspace-relative directory paths it matches. `*` matches a single path
+ * segment; `**` matches any depth. A pattern ending in `/**` also matches its
+ * own base directory (e.g. `copybooks/**` yields `copybooks` itself), since
+ * `minimatch` alone would only match the descendants. Matching is
+ * case-insensitive. Each returned path is later indexed by {@link expandLib}.
  */
-async function expandDirectoryTree(
-  rootLib: string,
-  rootUri: URI,
+async function expandWildcardLib(
```
Lotes, line 325: Just a foundational question, because I am not so deep in this part of the code:
The package „glob“ can already do such a search, as far as I know from my experience. Why did we need to implement our own thing? What is the key difference here?

msujew: `glob` is only for `file:/` uri search, but we need it for all URI schemes (even virtual) ones. That's why we roll our own.

packages/language/test/test-builder.ts
Lotes: Same goes for the local variable: Reading the name does not feel like a boolean.
Plus, you have two variables here that end with "Uri". They seem to be similar, but they are actually different types.

msujew (approved):
Thanks for the additional review @Lotes. I think this is good to go now 👍

PR #759, opened 2026-06-18
Cascading include file diagnostics up to the including file

Lotes:
Closes #718

If there are any diagnostics inside the included files, cascasde them to the importing file.


packages/language/src/workspace/lifecycle.ts
```diff
@@ -46,6 +50,9 @@ export async function lifecycle(
   await interruptAndCheck(cancellation);
   validate(compilationUnit);
   await interruptAndCheck(cancellation);
+  const includes = extractIncludeDirectives(compilationUnit);
+  markErroneousIncludes(compilationUnit, includes);
+  await interruptAndCheck(cancellation);
```
msujew: Suggestion: I don't like that this is a separate lifecycle phase. Can we make this a part of the validation?

packages/language/src/workspace/include-validations.ts
```diff
@@ -0,0 +1,126 @@
...
+ */
+import { Severity, diagnostic } from "../language-server/types";
+import { IncludeDirective, SyntaxKind, IncludeItem } from "../syntax-tree/ast";
+import { traverseAllNodes } from "../syntax-tree/ast-iterator";
+import { DiagnosticCategory } from "../validation/diagnostics-store";
+import { CompilationUnit } from "./compilation-unit";
+
+export function extractIncludeDirectives(compilationUnit: CompilationUnit) {
+  const directives: IncludeDirective[] = [];
+  traverseAllNodes(compilationUnit.preprocessorAst, (node) => {
```
msujew: Suggestion: `traverseAllNodes` is a rather expensive process for large ASTs. Can we integrate this into the validation (which already does this)?

packages/language/src/workspace/include-validations.ts
```diff
@@ -0,0 +1,126 @@
...
+      severitiesByUri.set(
+        uri,
+        Math.max(severitiesByUri.get(uri)!, severity) as Severity,
+      );
+    }
+  }
+  for (const diagnostic of compilationUnit.diagnostics.getAll()) {
+    if (!diagnostic.uri) continue;
+    addSeverity(diagnostic.uri, diagnostic.severity);
+  }
```
msujew: This goes hand in hand with the comments above: Should we limit this to lexing/parser errors? While semantic errors are still interesting, the important part for users is that they want to know whether this file was parsed correctly. If we also include semantic errors, we really need to do this in a separate phase.

packages/language/src/workspace/include-validations.ts
```diff
@@ -0,0 +1,126 @@
...
+function renderErrorsOnIncludeItems(
+  handledIncludingFiles: Map<IncludeItem, Severity>,
+  compilationUnit: CompilationUnit,
+) {
+  for (const [item, severity] of handledIncludingFiles.entries()) {
+    const message = `Included file '${item.relativeFilePath}' contains ${SeverityToString[severity]}.`;
+    compilationUnit.diagnostics.add(
+      DiagnosticCategory.Validation,
+      diagnostic(severity, message, item.token),
+    );
```
msujew: With the comment above in mind, simply say "x parser errors"?

msujew (commented):
Works well, but I'm concerned about the additional lifecycle step for this. See comments below. WDYT?

packages/language/src/workspace/include-validations.ts
Lotes: Have a different approach now, thanks

packages/language/src/validation/pp-validator.ts
```diff
@@ -30,10 +37,37 @@ export function registerPreprocessorValidationChecks(): ValidationChecks {
...
+        item.filePath!,
+      );
+      if (errors.length > 0) {
+        acceptor({
+          uri: item.token?.uri?.toString(),
+          range: item.range ?? undefined,
+          message: `Included file '${item.relativeFilePath}' contains ${errors.length} ${DiagnosticCategoryToString[category]}.`,
+          severity: Severity.E,
+        });
+      }
```
msujew: Suggestion: This would generate three separate diagnostics in the worst case, right? Can we cut this down to one?

msujew (commented):
Thanks, I think that's better. Just one more caveat: I don't think we should produce multiple diagnostics, just because the categories are different. In fact, I don't think the user cares at all whether something is a lexer, compiler options or a parser error - it's just an internal splitting so we can test each of them individually.

msujew (approved):
Perfect, thank you!

PR #763, opened 2026-06-23
CICS: distinguish PL/I from COBOL comments

Lotes:
There is a difference between PL/I and COBOL comments when using CICS.

I added two derived classes, one for COBOL, one for PL/I. I also thought about extension via composition, but thought it would blow up the test changes (having the constructor with extra parameters).

We could add more validations if needed.


msujew (commented):
Thanks, that works well. We could merge it as it is, but I've got a misgiving with the inheritance structure: With the idea in mind that the `preprocessor-cics` and `preprocessor-db2` libraries are to be reused in separate vscode extensions (which probably simply get the name of the host language as input?) I would rather have an enum `COBOL/PLI` that switches the behavior. Internally, we can maybe implement this as a strategy pattern? I think it makes sense to favor composition over inheritance in this case here.

I.e. pass the enum to `CICSPreprocessor` and let it internally select which strategy to pick.

packages/language/src/parser/exec-parser.ts
```diff
@@ -29,6 +29,7 @@ import {
   Severity as LSSeverity,
   Diagnostic as LSDiagnostic,
 } from "../language-server/types";
+import { HostLanguageType } from "../../../preprocessor-cics/out/engine/host-languages";
```
msujew: Suggestion: Can we export/import this from `preprocessor-cics`?

msujew (approved):
Thanks, looks good. Please just fix the issue below before merging 👍

packages/language/src/parser/exec-parser.ts
Lotes: Ah damnit, I auto-imported this without looking at the result ^^*....

PR #764, opened 2026-06-24
Prevent includes outside of library directories

msujew:
Does as the title says. On current `development` the test in the PR fails, because the include resolution is able to "escape" the lib folder it was using to resolve the path. This change simply ensures that we never escape the lib that we've used to compute the directory.


ssmifi (approved):
LGTM

PR #765, opened 2026-06-24
Update lib quick fix to use text edit

msujew:
Does a few things:

1. Adjusts the quick fix that adds libs to the plugin config to use a text edit (instead of a command).
2. Adjusts the quick fix that removes libs from the plugin config to use a text edit
3. Removes a bunch of tests that are no longer required.


ssmifi (approved):
LGTM 👍

PR #766, opened 2026-06-29
Use local vsce, bumb version

msujew:
`pnx`/`pnpm dlx` downloads the package if it can't find it globally installed - however, we want to actually use the locally downloaded vsce package. This change makes sure that we use the local one.

Also removes the minimatch override, since the vsce bumb is now using the latest version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
   },
   "overrides": {
     "isexe": "3.1.1",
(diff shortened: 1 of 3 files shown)


Lotes (approved):
Looks like it works already using the GH action... so...

🎶 Du bist gut genuuug...! 🎵

PR #767, opened 2026-06-29
Fix comment highlighting above compiler directives

ssmifi:
Fixes https://github.com/zowe/zowe-pli-language-support/issues/694

Our assumption is that compiler options are always located at the beginning of the file. This still holds semantically, but directives can be preceded by comments and the semantic token information should not overlap. 
The fix uses the actual `startLine` of the token range. Added a test for this.

### DIFF
--- a/packages/language/test/fourslash/semantic-tokens/comments-before-process.ts
+++ b/packages/language/test/fourslash/semantic-tokens/comments-before-process.ts
@@ -1,0 +1,23 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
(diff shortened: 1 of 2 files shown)


msujew (approved):
Thanks, makes sense 👍

packages/language/test/fourslash/semantic-tokens/comments-before-process.ts
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+//// <|comment:/* a comment that mentions happens and with before the directive|>
+////    and a second line of that very same comment here */
```
msujew: Minor: Can we test that the comment semantic token continues on the second line?

PR #768, opened 2026-06-29
Propagate errors from INCAFTER compiler option

Lotes:
Closes #718 

If the is any lexer/parser/compiler options error in the INCAFTER file, the include gets assigned a diagnostic.

### DIFF
--- a/packages/language/src/workspace/lifecycle.ts
+++ b/packages/language/src/workspace/lifecycle.ts
@@ -12,7 +12,7 @@
 import { ReferencesCache, resolveReferences } from "../linking/resolver";
 import { iterateSymbols } from "../linking/symbol-table";
 import { CompilationUnit, addBuiltinUnits } from "./compilation-unit";
-import { Program } from "../syntax-tree/ast";
+import { Program, SyntaxKind } from "../syntax-tree/ast";
 import {
   generatePliValidationDiagnostics,
   generatePreprocessorValidationDiagnostics,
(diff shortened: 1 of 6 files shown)


packages/language/test/utils.ts
```diff
@@ -279,12 +279,12 @@ export async function parseAndLink(
   const document = TextDocument.create(uri.toString(), "pli", 0, text);
   const unit = await createCompilationUnit(uri, defaultTestWorkspace());
 
-  await lifecycle.tokenize(unit, document);
+  const { compilerOptions } = await lifecycle.tokenize(unit, document);
```
msujew: Suggestion: Carrying this through the lifecycle seems like a workaround for me. Do you think it makes sense to store the whole object on the compilation unit? I think that would be preferable.

packages/language/src/workspace/lifecycle.ts
```diff
@@ -106,6 +108,21 @@ export function validate(compilationUnit: CompilationUnit): void {
...
+    const item = compilerOptionsResult.result.options.incAfter.token.element;
+    if (item && item.kind === SyntaxKind.IncludeItemFile) {
+      PropagateIncludeItemErrors(
+        item,
+        (d) =>
+          compilationUnit.diagnostics.add(DiagnosticCategory.Validation, d),
+        compilationUnit,
+      );
+    }
+  }
```
msujew: Minor: With the above in mind, can we move this into `generatePreprocessorValidationDiagnostics`?

msujew (commented):
Thanks, works well 👍

packages/language/test/utils.ts
Lotes: I can do so. I was just wondering: Is it actually possible to set the compiler options in different sections of the same file?
I was assuming that one could switch the arithmetic mode between modes IBM and IEEE multiple times.

Lotes: Refering to [this here](https://www.ibm.com/docs/en/epfz/6.1.0?topic=sd-process-directive) it is not the case... but was it not possible somehow else?

msujew: The spec says the compiler options have to appear on the first byte of the file. However, the compiler is much more lenient than that. We currently don't support switching compiler options mid-file.

msujew (approved):

PR #770, opened 2026-06-29
Watch changes to workspace in language server

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/753
Closes https://github.com/zowe/zowe-pli-language-support/issues/754

Instead of running the watcher in the vscode extension side (which currently only watches for changes in the pliplugin directory), we simply watch all files in the workspaces. Core logic is the same in general, but now we can also correctly update compilation units whose included files no longer exist or were just created.

Only testable (right now) via the vscode extension or the playground. There are a few manual test cases:
1. Delete a file in the vscode workspace (that was previously included somewhere. Ensure that the include no longer resolves.
2. Restore the same file - it should now resolve again.
3. Same as 1/2 but with whole directories.
4. Delete the `.pliplugin` folder (or perform changes to it) - it should behave as expected.


ssmifi (commented):
Code looks good btw. @Lotes please do the approval as soon as you are satisfied with the test cases.

Lotes (commented):
* Rename file ✅
* Rename folder ✅
* Move file ✅
* Delete file ✅
* Delete folder ✅
* Create folder+file ✅

Thanks, I am happy with the playground. I will approve after I checked the code a last time.

packages/language/src/config/lib-expander.ts
```diff
@@ -211,6 +205,8 @@ async function expandDirectoryTree(
   rootUri: URI,
   fs: FileSystemProvider,
 ): Promise<LibsDirEntry[]> {
+  // Normalize path first
+  rootLib = rootLib.replace(/\\/g, "/");
```
Lotes: Don't we have a "normalize" function already?

Lotes (approved):
One last finding below.

Nice changes. Looks good to me.

packages/language/src/config/lib-expander.ts
msujew: Yes, but only for URIs. The `rootLib` isn't actually a URI, just a path.

Lotes: On `development` there are 11 occurrences of the `/\\/g` replacement. Can be an opportunity to reduce redundancy. Even the comment would be obsolete when the line would be `rootLib = normalizePath(rootLib)`.

msujew: I've introduced a `UriUtils.normalizePath` function 👍

PR #772, opened 2026-07-02
Fix ruler bounce

ssmifi:
Fixes an issue where switching between tabs actually toggles the display of the margin rulers. 

`onDidChangeVisibleTextEditors` is actually called twice per tab switch (one time for the closing/opening tab each). Also, the margin notification from the language server might arrive after the tab switch notification. 
Both notifications now use the same debounced update mechanism. Last update wins. 

Also reduced updates/ruler flickering by not deleting the current ruler settings _for PLI_, when switching to a non-PLI file in automatic ruler mode.

You can test this on current dev by opening 3 pli files and just switching tabs between them in any order. The rulers should consistently toggle on/off. The PR should fix this.

### DIFF
--- a/packages/vscode-extension/src/extension/decorators.ts
+++ b/packages/vscode-extension/src/extension/decorators.ts
@@ -110,6 +110,35 @@ export namespace MarginIndicatorDecorator {
     string,
     { m: number; n: number; active: boolean }
   >();
+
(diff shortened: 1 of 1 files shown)


msujew (approved):
Thanks, works much better!

PR #775, opened 2026-07-08
Colorize EXEC CICS/SQL keywords: using semantic tokens

### LINKED ISSUE #674: Coloring EXEC CICS and SQL
Currently `EXEC CICS and EXEC SQL ` are colored same as like PL/I program statements.  

Image

Let's make both `EXEC CICS and EXEC SQL `  colored differently as below. 

Image

Lotes:
Closes #674

### DIFF
--- a/packages/language/src/language-server/semantic-tokens.ts
+++ b/packages/language/src/language-server/semantic-tokens.ts
@@ -24,6 +24,7 @@ import {
 import { CstNodeKind } from "../syntax-tree/cst";
 import {
   controlTokens,
+  EXEC,
   modifierTokens,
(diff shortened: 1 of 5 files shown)


msujew (commented):
Now that I look at the implementation over at the COBOL side, maybe we should just use the same token type they're using? I.e. just make it a string. See:

https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol/blob/87cfdc478079c3cb02e549ca75095a6f6f85122a/clients/cobol-lsp-vscode-extension/syntaxes/COBOL.tmLanguage.json#L125-L130

PR #777, opened 2026-07-09
Remove dead command code

msujew:
This is a follow up on https://github.com/zowe/zowe-pli-language-support/pull/765. Removes some dead code that was no longer called anywhere.


ssmifi (approved):
Makes sense.

PR #778, opened 2026-07-09
Add more safeguards for macro interpreter

msujew:
Follow up on #748, adds additional safeguards for array generation.

### DIFF
--- a/packages/language/src/preprocessor/instruction-interpreter.ts
+++ b/packages/language/src/preprocessor/instruction-interpreter.ts
@@ -79,6 +79,10 @@ interface ArrayValue {
   readonly upper: number;
 }
 
+// Safeguard against unbounded allocation from source-controlled array
+// dimensions/repetition counts (e.g. %DCL X(999999999) or (999999999)(expr)).
+const MAX_ARRAY_SIZE = 100_000;
+
 /**
  * Creates a simple array value with the given elements and default lower bound of 1.
  * (PL/I arrays are 1-based by default)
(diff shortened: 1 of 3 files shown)


packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -234,6 +242,19 @@ function generateVariableValue(
         upper = valueToNumber(evaluatedUpper, 1);
       }
       length = upper - lower + 1;
+      if (fullLength === 0) {
+        fullLength = length;
+      } else {
+        fullLength *= length;
+      }
+      if (fullLength >= MAX_ARRAY_SIZE) {
```
ssmifi: _Minor_: Later tests use `>`, which is closer to what the comment describes (if you want to be pedantic :)).

ssmifi (approved):
LGTM, just one tiny remark.

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -79,6 +79,10 @@ interface ArrayValue {
   readonly upper: number;
 }
 
+// Safeguard against unbounded allocation from source-controlled array
+// dimensions/repetition counts (e.g. %DCL X(999999999) or (999999999)(expr)).
+const MAX_ARRAY_SIZE = 100_000;
```
Lotes: Maybe `MAX_ARRAY_ELEMENT_COUNT`?
When I am reading `SIZE` or `LENGTH, I am wondering if one meant "byte count".

packages/language/src/preprocessor/instruction-interpreter.ts
```diff
@@ -234,6 +242,19 @@ function generateVariableValue(
...
+        fullLength = length;
+      } else {
+        fullLength *= length;
+      }
+      if (fullLength >= MAX_ARRAY_SIZE) {
+        // Too many elements, return empty array
+        return {
+          array: [],
+          lower: 1,
+          upper: 0,
```
Lotes, line 255: `upper: 0` when `lower` is `1`?
Looks weird and I think their is some validation on a different level against such situation.
Is it not enought that the array field is empty?

msujew: `{ lower:1, upper:1 }` implies one element in the array. This is empty though.

PR #779, opened 2026-07-10
Add ppinclude option effect

### LINKED ISSUE #300: Extend PPINCLUDE && PP(INCLUDE(...)) for INCLUDE PP Logic
An extension of the work for #290, which focuses on the following additional behaviors of the INCLUDE pre-processor logic. Both ways of passing options to the INCLUDE pre-processor need to be handled, and should respect each other, but currently we're only handling the `PP(INCLUDE)` invocation approach (which can configure & invoke).

- `PPINCLUDE('ID(-inc)')` by itself should _not_ activate the alternative, it also requires `PP(INCLUDE)` (no args)
- `PPINCLUDE('ID(-inc)')` should be equivalent to `PP(INCLUDE('ID(-inc)'))`
- `PPINCLUDE('ID(-inc)') PP(INCLUDE('ID(++inc)'))` is equiv to `PP(INCLUDE('ID(-inc) ID(++inc)'))`, which reduces to `PP(INCLUDE('ID(++inc)'))`
- Options str is limited to 1k characters
- `NOPPINCLUDE` should disable the alternative, even when it was defined using `PP`

Additionally, it seems that`NOPP` shouldn't clear options information information, but does indicate that we won't invoke a pre-processor.

ssmifi:
Closes https://github.com/zowe/zowe-pli-language-support/issues/300

Adds the logic for the `PPINCLUDE` (also see https://github.com/zowe/zowe-pli-language-support/issues/300) and provides a helper function `getEffectiveIncludeAlt` that can be used to retrieve the final active configuration.
Added several tests to cover the different cases.

### DIFF
--- a/packages/language/test/fourslash/preprocessor/compiler-options/options-pli/ppinclude-length-exceeded.ts
+++ b/packages/language/test/fourslash/preprocessor/compiler-options/options-pli/ppinclude-length-exceeded.ts
@@ -1,0 +1,22 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../../../framework.ts" />
+
+// @wrap: process
+//// // PPINCLUDE configuration must not exceed 1000 characters in length.
+////*PROCESS PPINCLUDE(<|1:'\$1k'|>);
+
+verify.expectDiagnosticsAt(1, {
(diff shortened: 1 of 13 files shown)


packages/language/test/fourslash/preprocessor/compiler-options/options-pli/ppinclude-length-exceeded.ts
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../../../framework.ts" />
+
+// @wrap: process
+////*PROCESS PPINCLUDE(<|1:'\$1k'|>);
```
Lotes, line 16: For me these cases look very cryptic. What do you think about adding a short description? Or may add a link to documentation?

msujew (approved):
Thanks, looks good to me 👍

PR #782, opened 2026-07-14
Fix missing pp exec when plugin config is missing

ssmifi:
When no `.pliplugin` config files exist AND no `*PROCESS` directive is present, EXEC SQL/CICS statements fail with parser errors, because the PPs are not executed. 

`CompilerOptionTranslator` initializes with empty objects `{}` instead of the default compiler options. 
Any config (even an empty one) triggers `translateCompilerOptions()`, which copies defaults from `translator.options`. 
However, if no config is present, `translateCompilerOptions()` is never called and `result.options` stays empty, and hence, gives an empty `PP` option to the pipeline.

The fix initializes `result.options` with the correct defaults from the start.

I also added a `@noDefaultConfig` tag to the test framework, because the test builder always creates a default plugin configuration, which prevents a proper fourslash test.

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/translate.ts
+++ b/packages/language/src/preprocessor/compiler-options/translate.ts
@@ -9,7 +9,11 @@
  *
  */
 
-import { CompilerOptionResult, CompilerOptionsPP } from "./options";
+import {
+  CompilerOptionResult,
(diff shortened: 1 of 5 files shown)


msujew (approved):
Thanks!

PR #784, opened 2026-07-15
Add support for multiple workspaces

Lotes:
### Done

* introducing multiple `WorkspaceContexts` under `CompilationUnitHandler`
* each `WorkspaceContext` manages its `CompilationUnits`
* `CompilationUnitHandler` is the place to initialize new workspaces
* getting the right `CompilationUnit` is still hidden by `CompilationUnitHandler` which delegates to the specific `WorkspaceContext`
* it is actually very simple :-)
* missing configuration is handled apart from this mechanism
* quick fixes reach out to the specific `WorkspaceContext`

### TODO for further PRs

- lazy loading of libs
- automatically test features like this

### DIFF
--- a/packages/language/src/language-server/connection-handler.ts
+++ b/packages/language/src/language-server/connection-handler.ts
@@ -55,7 +55,12 @@ import { applySourceActions } from "./code-actions/apply-source-actions";
 import { commandCreateConfig } from "./commands";
 import { Commands } from "./constants";
 import { signatureHelpRequest } from "./signature-help-request";
-import { GlobalConfigLoader, Messages, NotificationType, RequestType } from "../utils/messages";
+import {
+  GlobalConfigLoader,
+  Messages,
+  NotificationType,
+  RequestType,
+} from "../utils/messages";
 import { configCompletionRequest } from "./completion/completion-plugin-configuration";
 import { JsonItemMeta } from "../config/schema";
 import { assertType } from "../preprocessor/util";
@@ -70,7 +75,11 @@ export function startLanguageServer(
   // Wire the on-demand file loader for include URIs to the workspace's fs
   // so the document store doesn't reach for any module-level singleton.
   resetDocumentProviders(fs);
-  const compilationUnitHandler = new CompilationUnitHandler(fs, globalConfigLoader, new LongRunningOperationImpl(connection));
+  const compilationUnitHandler = new CompilationUnitHandler(
+    fs,
+    globalConfigLoader,
+    new LongRunningOperationImpl(connection),
+  );
   compilationUnitHandler.listen(connection);
   let folders: WorkspaceFolder[] = [];
 
@@ -152,12 +161,14 @@ export function startLanguageServer(
         },
       ],
     });
-    const promises = folders.map(async (folder) =>
-      compilationUnitHandler.initializeWorkspaceFolder(folder.uri),
-    ).concat(
-      //add more default schemes here if needed
-      compilationUnitHandler.initializeFallbackFolderForScheme("file"),
-    );
+    const promises = folders
+      .map(async (folder) =>
+        compilationUnitHandler.initializeWorkspaceFolder(folder.uri),
+      )
+      .concat(
+        //add more default schemes here if needed
+        compilationUnitHandler.initializeFallbackFolder(),
+      );
     await Promise.all(promises);
     compilationUnitHandler.markReady();
   });
--- a/packages/language/src/language-server/connection-handler.ts
+++ b/packages/language/src/language-server/connection-handler.ts
@@ -152,6 +152,9 @@ export function startLanguageServer(
     });
     const promises = folders.map(async (folder) =>
       compilationUnitHandler.initializeWorkspaceFolder(folder.uri),
+    ).concat(
+      //add more default schemes here if needed
+      compilationUnitHandler.initializeFallbackFolderForScheme("file"),
     );
     await Promise.all(promises);
     compilationUnitHandler.markReady();
--- a/packages/language/src/language-server/connection-handler.ts
+++ b/packages/language/src/language-server/connection-handler.ts
@@ -55,20 +55,22 @@ import { applySourceActions } from "./code-actions/apply-source-actions";
 import { commandCreateConfig } from "./commands";
 import { Commands } from "./constants";
 import { signatureHelpRequest } from "./signature-help-request";
-import { Messages, NotificationType, RequestType } from "../utils/messages";
+import { GlobalConfigLoader, Messages, NotificationType, RequestType } from "../utils/messages";
 import { configCompletionRequest } from "./completion/completion-plugin-configuration";
 import { JsonItemMeta } from "../config/schema";
 import { assertType } from "../preprocessor/util";
+import { LongRunningOperationImpl } from "../utils/promises";
 export { PluginConfiguration, Commands } from "./constants";
 
 export function startLanguageServer(
   connection: Connection,
   fs: FileSystemProvider,
+  globalConfigLoader: GlobalConfigLoader,
 ): void {
   // Wire the on-demand file loader for include URIs to the workspace's fs
   // so the document store doesn't reach for any module-level singleton.
   resetDocumentProviders(fs);
-  const compilationUnitHandler = new CompilationUnitHandler(fs);
+  const compilationUnitHandler = new CompilationUnitHandler(fs, globalConfigLoader, new LongRunningOperationImpl(connection));
   compilationUnitHandler.listen(connection);
   let folders: WorkspaceFolder[] = [];
 
--- a/packages/language/src/language-server/connection-handler.ts
+++ b/packages/language/src/language-server/connection-handler.ts
@@ -12,15 +12,12 @@
 import {
   CompilationUnit,
   CompilationUnitHandler,
-  publishPluginConfigDiagnostics,
+  pluginConfigChanged,
 } from "../workspace/compilation-unit";
 import {
-  CancellationToken,
   Connection,
   DidChangeWatchedFilesNotification,
   DocumentHighlight,
-  FileChangeType,
-  FileEvent,
   TextDocumentSyncKind,
 } from "vscode-languageserver";
 import { URI, UriUtils } from "../utils/uri";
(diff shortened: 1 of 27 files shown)


msujew (commented):
This seems to work pretty well when it comes to files that live inside of actual workspace folders, but it breaks down once we open files outside of that.

There are two main use cases for this in particular:
1. A user has no workspace open, and simply opens a file from their disk.
2. A user opens a file from the Zowe Explorer file system. (i.e. a file from the `zowe-ds:/` URI scheme).

In both cases, the VSCode config should be used (or a default one, if not provided) to provide language support to those files.

packages/vscode-extension/src/extension/config-loader.ts
```diff
@@ -51,21 +51,20 @@ export function registerConfigLoader(
 export async function getGlobalConfig(
   userSettingsUri: vscode.Uri,
 ): Promise<Messages.GlobalConfig> {
-  // TODO: support multi-root workspaces here as well
-  const folder = vscode.workspace.workspaceFolders?.[0];
+  const workspaceFolder = locateWorkspaceFolder(userSettingsUri);
```
Lotes: seems to be wrong

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -626,15 +626,15 @@ export class PluginConfigurationProvider {
...
+  private async fetchGlobalSettings(
+    workspaceUri: URI,
+  ): Promise<Messages.GlobalConfig | undefined> {
     if (!this.connection) return undefined;
     try {
       return await sendRequest(
         this.connection,
         Messages.GetGlobalConfig,
-        undefined,
+        workspaceUri.fsPath,
```
msujew: Suggestion: Send the whole stringified URI. We can't guarantee that the settings file lives in a directory with the `file` scheme. Generally, using `fsPath` loses information and should be avoided ;)

msujew (commented):
Still doesn't seem to work on files outside of the current workspace. It creates compilation units for them (it did not do so in a previous version of the PR), but it doesn't load any configurations for those (from the vscode settings). You can open the `code_samples/RXGIM.pli` file for example, and see the `Could not resolve include directive. Plugin configuration is missing` error on the include. When debugging, I could see that the `pluginConfigs` inside of the `PluginConfigurationProvider` is empty.

AFAIK, there are no unit tests for this scenario either, right?

packages/language/src/workspace/compilation-unit.ts
```diff
@@ -357,16 +358,26 @@ export class CompilationUnitHandler extends WorkspaceFolderTree<WorkspaceContext
...
+  async initializeFallbackFolderForScheme(uriScheme: string) {
+    const uri = UriUtils.toUri(`${uriScheme}:///`);
+    const workspace = new WorkspaceContext(this.fs, this.configLoader, this.longRunningOperation);
+    this.fallbackWorkspaces[uriScheme] = workspace;
+    const diagnosticsByUri = await workspace.config.init(uri);
+    if (this.connection) {
+      publishPluginConfigDiagnostics(this.connection, diagnosticsByUri);
+    }
+    return workspace;
+  }
```
msujew: Question: I don't quite follow this logic. Why do we need a dedicated workspace context per URI scheme? In my mind, just one should be enough?

packages/vscode-extension/src/extension/config-loader.ts
```diff
@@ -9,31 +9,25 @@
...
+  static register(client: BaseLanguageClient, context: vscode.ExtensionContext): void {
+    client.onRequest(Messages.GetUserSettingsUriRequest.method, async () => {
+      return deriveUserSettingsUri(context.globalStorageUri).toString();
+    });
+  }
+  constructor(private readonly connection: Connection) {}
+  async loadGlobalConfig(workspaceUri: vscode.Uri): Promise<Messages.GlobalConfig> {
+    const userSettingsUri = await sendRequest(this.connection, Messages.GetUserSettingsUriRequest, undefined);
+    return getGlobalConfig(workspaceUri, UriUtils.toUri(userSettingsUri));
+  }
```
msujew: Suggestion: Please don't mix language server and vscode types in one implementation. I.e. `loadGlobalConfig` receives a `import(vscode-uri).URI`, this implementation uses `vscode.Uri` (which is not the same!).

I was initially very confused, because I thought the `VscodeGlobalConfigLoader` lives in the context of the vscode extension host. Move this to the `vscode-extension/src/language/` directory instead.

msujew (commented):
I know this isn't done yet, but I wanted to give some quick feedback about this before leaving for the weekend.

packages/vscode-extension/src/extension/config-loader.ts
msujew: Nevermind, already resolved on the latest commit (that GitHub didn't want to show me).

packages/language/test/workspace/multi-workspace.test.ts
```diff
@@ -0,0 +1,350 @@
...
+    const outsideProgram = UriUtils.toUri("file:///outside/test2.pli");
+
+    await fs.writeFile(
+      settingsUri,
+      JSON.stringify(
+        {
+          "pli.pgm_conf": {
+            pgms: [
+              {
+                program: "**/*.pli",
```
Lotes: @msujew If I change this to `"*.pli"`, the test will fail.

Is this the correct behaviour? The config is there, but not matching the file pattern. Our fallback workspace context has its root in the folder `file:///`.

msujew: I believe this is correct. `*.pli` only works "as expected" if we have some kind of workspace available, from which we can do relative lookups.

packages/language/src/language-server/connection-handler.ts
```diff
@@ -421,130 +423,40 @@ export function startLanguageServer(
...
-    return false;
-  }
   onNotification(
     connection,
     Messages.OnDidChangePluginConfigSettingsNotification,
     async () => {
       // Handle changes to the pli.pgm_conf and pli.proc_grps settings in vscode
-      await pluginConfigChanged();
+      const promises = compilationUnitHandler
+        .getAllWorkspaceFolders()
```
msujew, line 439: Bug: This `getAllWorkspaceFolders` here does not update the fallback workspace folder. Therefore, files outside of the current workspace never are notified that the plugin config has changed.

Suggestion: Add a test for `getAllWorkspaceFolders` that ensures that all the fallback workspace context is also returned here.

packages/language/src/workspace/compilation-unit.ts
```diff
@@ -403,8 +355,90 @@ export class CompilationUnitHandler {
...
+  /** must be initialized at least once */
+  async initializeFallbackFolder() {
+    const uri = UriUtils.toUri(`file:///`);
+    const workspace = new WorkspaceContext(
+      this.fs,
+      this.configLoader,
+      this.longRunningOperation,
+    );
+    this.fallbackWorkspace = workspace;
+    const diagnosticsByUri = await workspace.config.init(uri);
```
msujew, line 385: Suggestion: I'm not a big fan that `init` accepts a placeholder URI here. Theoretically, `file:///` is a valid workspace folder (although a rather weird one). Can we make the URI here optional?

msujew (commented):
I had some initial issues getting this to work (since it doesn't listen on vscode settings changes if we're in the fallback context). However, after getting this to run, this looks rather nice. Thanks!

packages/language/src/workspace/compilation-unit.ts
Lotes: As discussed: instead of a `file:` scheme, I will use the builtin schema.

msujew (commented):
The `programMatchRank` function takes the URI scheme into account when attempting to match the URI to the given pattern:

Image

Image

Image

If the URI schemes differ (which they always will!), no plugin config will get loaded. You can test this by opening any PLI outside of your workspace and placing a breakpoing in `PluginConfigurationProvider.getProgramConfig`.

packages/language/src/workspace/compilation-unit.ts
```diff
@@ -293,8 +294,8 @@ export async function addBuiltinUnits(
   unit.rootPreprocessorScope = Scope.createChild(macroScope);
 }
 
-export class CompilationUnitHandler {
-  private compilationUnits: Map<string, CompilationUnit> = new Map();
+export class CompilationUnitHandler extends WorkspaceFolderTree<WorkspaceContext> {
+  private fs: FileSystemProvider;
```
msujew: Suggestion: Favor composition over inheritance. That way, you also don't need to use the `override`, which makes reasoning difficult.

packages/language/test/workspace/multi-workspace.test.ts
```diff
@@ -0,0 +1,353 @@
...
+    )!;
+    expect(
+      firstGroup.compilerOptions.map((co) => co.value).includes("AGGREGATE"),
+    ).toBeTruthy();
+
+    const workspaceOutside = ch.getWorkspaceFolderOf(outsideProgram);
+    expect(workspaceOutside).toBeDefined();
+    expect(
+      workspaceOutside!.config.hasProgramConfig(outsideProgram),
+    ).toBeFalsy();
```
msujew: Bug: This should be `toBeTruthy()`. It should use the global config.

msujew (commented):
Ah, I was wondering why the test is passing, even though the feature doesn't work as expected. Turns out, it asserts incorrectly. See below. 👍

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -1073,8 +1061,9 @@ export class PluginConfigurationProvider {
     // No PL/I-extension filtering here: callers only ever pass files already
     // identified as PL/I (client language id / auto-detect), so glob matching
     // the path alone is sufficient.
-    // Note that we need to decode the URI
-    const uri = program.toString(true);
+    // Note that we just need the path of the URI in order to match against
+    // the program config patterns
+    const uri = program.path;
```
msujew: Minor: Not a URI.

packages/language/src/workspace/plugin-configuration-provider.ts
msujew: Minor: Not a URI

msujew (commented):
I just noticed that the `pli-builtin` URI scheme breaks absolute (local) lib imports. It will attempt to load the directories from the same URI scheme that was used to generate the workspace:

Image

https://github.com/zowe/zowe-pli-language-support/blob/9afda719c5a06f600cd0c01038d71e16f6be980d/packages/language/src/config/path-resolver.ts#L35-L37

Since we still need this behavior for non-file URI scheme workspaces, I would actually recommend to move the fallback URI back to `file:///`. It's not a 100% solution, but it's better than most things.

Everything else works great!

msujew (approved):
Thanks, looks good to me 👍

PR #796, opened 2026-07-16
Add go to configuration

ssmifi:
Closes https://github.com/zowe/zowe-pli-language-support/issues/789

Adds two commands to jump to the configuration files. 
Not sure if I got all edge cases, but it works for the settings.json as well.


msujew (commented):
This doesn't seem to work on files that are included? They should return the same results (i.e. jump to the same configs) as their including files.

msujew (approved):
Thanks, works well!

PR #800, opened 2026-07-16
Fix various PP option issues

### LINKED ISSUE #786: Retain PP settings for other PPs
The PP default is `PP(MACRO SQL CICS)`. Currently, if some settings to the MACRO PP are entered, the other PPs are removed from the pipeline. This should not be the case.

ssmifi:
Closes https://github.com/zowe/zowe-pli-language-support/issues/786

This PR fixes several issues in how the PP(...) compiler option is handled.

- **Fix pp compiler option configuration**
Fixed a bug where specifying `PP(...)` in more than one `*PROCESS` directive would reset the preprocessor pipeline instead of accumulating it. E.g., `PP(SQL CICS); PP(MACRO);` would incorrectly end up as just `PP(MACRO)` instead of `PP(SQL CICS MACRO)`.

- **Handle MACRO + PP**
Implemented the documented rule that the standalone `MACRO` option, combined with `PP(...)`, adds `MACRO` to the front of the preprocessor list unless it's already first. Added a new diagnostic reported when this implicit addition happens.

- **Fix PPXXX configuration**
Fixed `PPSQL`/`PPMACRO`/`PPCICS` one-time settings being incorrectly re-applied (and re-validated) on every subsequent `*PROCESS` directive in the same file, which caused spurious duplicate-option diagnostics.

- **Add dedicated compiler option post-processing step**
Implemented the remaining documented `PP(...)` limits: a maximum of 31 preprocessor steps, CICS invoked at most once, and SQL invoked at most twice (only if the first invocation specifies `INCONLY`). Introduced an explicit `postProcessCompilerOptions()` step that runs once, after all directives have been translated, so these whole-pipeline checks don't need to guard against re-reporting on every directive.
This is generalized even more in https://github.com/zowe/zowe-pli-language-support/pull/801

- **Fix nested exec statement**
The `EXEC CICS`/`EXEC SQL` statements nested inside a `%DO`/`%IF` macro-conditional block were always processed by the MACRO phase's internal statement walk, completely bypassing `PP(CICS)`/`PP(SQL)`. Hence, disabling `CICS`/`SQL` had no effect on macro-nested `EXEC` statements. The fix split the exec-statement parsing so only the dedicated `PP(CICS)`/`PP(SQL)` phase ever invokes the real preprocessor engine; any other caller now just preserves the statement's tokens (as a plain token statement) and defers to that dedicated phase.

- **Add user friendly error message on missing PP**
Added a new pipeline step `UnresolvedExecPhase` that runs once, after every configured `PP()` phase, and detects any `EXEC CICS`/`EXEC SQL` statement that no dedicated phase ended up processing (i.e., the preprocessor was never enabled). Reports a clear diagnostic (`"EXEC CICS requires the CICS preprocessor to be enabled via PP(CICS)"`) and replaces the statement so the generic parser error no longer also appears.

Note that changing the way the pipeline is configured now can have unintended effects with old process group configs when configuring PPs via `PP(MACRO('CASE(UPPER)')` for example, but this is intended and according to spec. See the documentation for more info. Add the other PPs when needed or use `PPMACRO` etc for the configuration.

```
If you specify the PP option more than once, the compiler effectively concatenates them. So
specifying PP(SQL) PP(CICS) is the same as specifying PP(SQL CICS). This also means that if you 
specifies PP(MACRO SQL('CCSID0')) and PP(MACRO SQL('CCSID0 DATE(ISO)')), the resulting PP option
is PP( MACRO SQL('CCSID0') MACRO SQL('CCSID0 DATE(ISO)')), and both the MACRO preprocessor and 
the SQL preprocessor will be invoked twice, and the second invocation of the SQL preprocessor will be in 
error. If you do this to override the earlier SQL options, it might be better not to specify the preprocessor 
options in the PP option, but rather to specify them through the PPSQL option, that is, specify PP(MACRO 
SQL) PPSQL('CCSID0 DATE(ISO)').
```


msujew (approved):
Looks good to me 👍

Issue #780, opened 2026-07-13
Implement cross-option mutual exclusivity


ssmifi:
The current compiler option translator only handles positive/negative mutex within a single option and suboptions. 
However, there are several options (also in the pre-processors) that are cross-option mutually exclusive, e.g.,  EXCI is mutually exclusive with CICS and with DLI in the CICS pre-processor. 
Therefore, extend the translator to handle cross-option mutual exclusivity and modify the rules accordingly.

PR #801, opened 2026-07-20
Generalize compiler option post processing

ssmifi:
Fixes https://github.com/zowe/zowe-pli-language-support/issues/787 and lays a foundation for https://github.com/zowe/zowe-pli-language-support/issues/780 in generalizing the post compiler option processing from https://github.com/zowe/zowe-pli-language-support/pull/800
PR is based on https://github.com/zowe/zowe-pli-language-support/pull/800.

- Adds some convenience in declaring compiler option rules by chaining. 
- Compiler option rules can add post processing individually which is executed after all directives have been processed. Note that the mechanism does not differ from https://github.com/zowe/zowe-pli-language-support/pull/800, but the instead of hardcoding the post effects into one method, they can be categorized by rule and added to the rule definition.
- Added the PPINCLUDE validation for https://github.com/zowe/zowe-pli-language-support/pull/787


msujew (approved):
Looks good to me, thanks 👍

PR #803, opened 2026-07-21
project .pliplugin program config must win over user-scope pli.pgm_conf

Wagner-Laranjeiras:
**Solves #802** 

Files were binding to the wrong process group when a user-scope `pli.pgm_conf` entry (glob or exact path) also matched a file, causing `%INCLUDE` to fail `(IBM1848I)` even though the project's process group listed the right lib.

Now project `.pliplugin program entries take precedence over user-scope settings entries at lookup time. Adds fourslash regression tests for the glob and exact-path cases.

**Changes:**

 1. **packages/language/src/config/schema.ts** — Added a ProgramOrigin type and an optional origin field on ProgramRecord (absent = "project"), so records know whether they came from the project or user settings.

 2. **packages/language/src/workspace/plugin-configuration-provider.ts** —
     - The config merge now tags each program entry with its origin (project vs. settings) and keeps project entries at highest precedence.
     - getProgramConfig now picks the best-ranked match instead of the first hit, using an encapsulated ProgramMatchRank constant + programMatchRank helper. Precedence: project beats settings, then exact path beats glob, then insertion order. This fixes both the broad-glob and exact-path settings cases, while still letting settings apply when no project entry matches.

3. **Tests** — Added two fourslash regression tests covering the glob case (program-config-precedence-over-settings.ts) and the exact-path case (program-config-precedence-exact-settings.ts).

### DIFF
--- a/packages/language/src/config/schema.ts
+++ b/packages/language/src/config/schema.ts
@@ -106,6 +106,13 @@ export interface GroupRecord extends ProcessGroup {
 }
 
 /**
+ * Where a {@link ProgramRecord} was loaded from. Project (`.pliplugin/`)
+ * entries take precedence over user-scope `settings.json` entries when a file
+ * matches program configs from both sources — see `getProgramConfig`.
+ */
+export type ProgramOrigin = "project" | "settings";
+
+/**
  * A {@link ProgramConfig} together with its merged compiler options. The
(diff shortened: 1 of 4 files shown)


msujew (approved):
Looks good, thank you!

PR #805, opened 2026-07-22
Compile program entries that live in a lib directory, and warn on the ambiguity

Wagner-Laranjeiras:
### Summary
A file can be both a **program entry point** (`pgm_conf.json`) and sit inside a directory configured as a **lib** (`proc_grps.json`) using an extension listed in that lib's `include-extensions`. This branch makes such files compile as standalone programs, and adds a config-level warning so users can spot and resolve the ambiguity intentionally.

### Changes
- **Program entries in lib dirs now get their own compilation unit** (`CompilationUnitHandler.getOrCreateCompilationUnit`). A registered program entry point always heads its own unit, even when its directory is also a lib — program status takes precedence over lib membership. Previously these files could be treated purely as library members and never compiled standalone.
- **New warning `COPC05` (`AmbiguousProgramLibOverlap`)**. Fires when a real file in a lib directory has an extension present in that group's `include-extensions` **and** is matched by a program entry (exact or glob). It's a **warning only** — it never blocks compilation. Emitted at most once per `(program entry, lib directory)` pair and attributed to the program entry in `pgm_conf.json`.

### Why
The compilation-unit change is correct but means files a user intended purely as copybooks may now compile standalone and emit surprising diagnostics. Rather than silently changing behavior, the warning surfaces the ambiguous config shape and suggests fixes (use a separate include directory, or drop the extension from `include-extensions`).

### Testing
- Unit tests for the compilation-unit gate and for the provider (`program-lib-overlap.test.ts`): glob overlap (deduped to one), exact-entry overlap, non-include extension (no warning), no matching files (no warning).
- Fourslash tests: positive (`program-lib-overlap-diagnostic.ts`) and negative (`program-lib-overlap-no-diagnostic.ts`).

### DIFF
--- a/packages/language/src/workspace/plugin-configuration-provider.ts
+++ b/packages/language/src/workspace/plugin-configuration-provider.ts
@@ -148,21 +148,6 @@ export function deserializeProcessGroup(
   };
 }
 
-function validatePluginConfig(
-  configs: Map<string, ProgramRecord>,
-  pGroups: Map<string, ProcessGroup>,
-): Diagnostic[] {
-  const diagnostics: Diagnostic[] = [];
-  // Check for unknown process groups references under "pgm_conf.json"
-  const pgroupReferences: Iterable<ProgramConfig> = configs.values();
-  const unknownProcessGroupsDiagnostic = validatePgroupReferences(
-    pgroupReferences,
-    new Set(pGroups.keys()),
-  );
-  diagnostics.push(...unknownProcessGroupsDiagnostic);
-  return diagnostics;
-}
-
 export type PluginConfigLspDiagnostics = MultiMap<string, LspDiagnostic>;
 export type PluginConfigDiagnostics = Diagnostic[];
 
@@ -561,16 +546,14 @@ export class PluginConfigurationProvider {
         computedLibs: [],
       });
     }
-    const validationResult = validatePluginConfig(
-      this.programConfigs,
-      mergedGroups,
-    );
-    diagnostics.push(...validationResult);
 
     this.postProcessProgramConfigs();
     const procGrpsDiagnostics =
       await this.postProcessProcessGroups(procGrpsDocuments);
     diagnostics.push(...procGrpsDiagnostics);
+    // Runs AFTER the post-processing above so the program/lib overlap check
+    // sees each group's expanded `computedLibs`.
+    diagnostics.push(...this.validatePluginConfig());
     this.libFileMatchers = undefined;
     this.configDiagnostics = diagnostics;
     const lspDiagnostics = await this.convertDiagnosticsToLsp();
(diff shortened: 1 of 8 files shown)


packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -859,6 +862,87 @@ export class PluginConfigurationProvider {
...
+              ext,
+              processGroup.name.value,
+            ),
+          );
+        }
+      }
+    }
+    return diagnostics;
+  }
+
```
Wagner-Laranjeiras, line 945: Should I move this into a dedicated file or is it good enough right now?

packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -859,6 +862,87 @@ export class PluginConfigurationProvider {
...
+   * The check is file-based: it iterates the real files each lib actually
+   * contains (populated in {@link postProcessProcessGroups} via
+   * {@link expandGroup}) and reuses {@link getProgramConfig} to decide whether
+   * a program pattern claims them, so it handles exact and glob program
+   * patterns uniformly and only fires when overlapping files really exist.
+   *
+   * Emits at most one diagnostic per (program entry, lib directory) pair,
+   * attributed to the program entry in `pgm_conf.json`.
+   */
+  private validateProgramLibOverlap(): Diagnostic[] {
```
msujew, line 887: Suggestion: We already have a `validatePluginConfig` function available. This method should be placed/live in there.

packages/language/test/workspace/program-lib-overlap.test.ts
msujew, line 1: Question: Isn't this just mirroring the tests already in the fourslash framework? Do we really need this?

packages/language/test/workspace/compilation-unit.test.ts
```diff
@@ -149,4 +153,109 @@ describe("Compilation Unit Tests", () => {
...
+    const libUnit = await ch.getOrCreateCompilationUnit(libOnlyUri);
+    expect(libUnit).toBeUndefined();
+  });
+
+  // Faithful reproduction of the pli-fs01 bug dynamics: with the folder listed
+  // both as the program-entry location AND as a lib, the entry file must be
+  // readable/compilable (before the fix, `getOrCreateCompilationUnit` returned
+  // `undefined` here, so the text could never be read). Removing the lib — the
+  // manual workaround from the bug report — must produce the same outcome.
+  test("Reads an entry point sharing its directory with a lib, with and without the lib", async () => {
```
msujew: Suggestion: I feel like this test goes way into the specifics of the bug. I believe the other test + the fourslash tests cover this behavior already.

msujew (commented):
Thanks, the fix looks good to me. I believe the tests are a bit overboard for this.

packages/language/src/workspace/plugin-configuration-provider.ts
Wagner-Laranjeiras: Hey Mark, again, thanks for the review. I was able to work on the test related items, but I couldn't find a solution for this: We need data that is only available after validatePluginConfig (and more) runs.

msujew: You mean we need the data from `this.postProcessProgramConfigs();` and `this.postProcessProcessGroups(procGrpsDocuments);`? Then why not move the `validatePluginConfig` call to after those calls? It doesn't matter whether we validate before or after the post-processing of the configs.

msujew (approved):
Thanks, looks good to me 👍

PR #806, opened 2026-07-23
Add completion request support for compiler options

ssmifi:
Provides completion support for compiler options on the top level. 
This is intentionally kept separate from `translator-pli` so that completion behavior can be declared without touching the translation logic. A future refactoring could derive these entries automatically from the translator rule definitions.

### DIFF
--- a/packages/language/src/preprocessor/compiler-options/completion-pli.ts
+++ b/packages/language/src/preprocessor/compiler-options/completion-pli.ts
@@ -1,0 +1,338 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/**
+ * Completion metadata for PLI compiler options
+ *
+ * This file is intentionally kept separate from translator-pli.ts so that
+ * completion behaviour can be declared without touching the translation logic.
+ * A future refactoring could derive these entries automatically from the
+ * translator rule definitions.
+ */
+export interface CompilerOptionCompletion {
+  /** Canonical option name (uppercase). */
+  name: string;
+  /**
+   * Mandatory number of parameters this option requires.
+   * `0`: no parentheses inserted (plain name).
+   * `>0`: a snippet with parentheses is offered: `NAME(${1:...})`.
+   */
+  mandatoryParams: number;
+  /**
+   * Candidate value(s) for the (typically single) parameter:
+   * - One entry: used as the snippet tab-stop default, e.g. `["2, 72"]` is
+   *   inserted as `MARGINS(${1:2, 72})`. May contain a single `<|>`
+   *   marker to indicate where the cursor should land after completion.
+   * - Multiple entries: rendered as an LSP snippet "choice", e.g.
+   *   `["DECIMAL", "HEXADEC"]` inserted as `AGGREGATE(${1|DECIMAL,HEXADEC|})`,
+   *   letting the user tab into the parentheses and cycle through the
+   *   offered values. These are also offered as completions when typing
+   *   inside the parentheses directly.
+   */
+  params?: readonly string[];
+}
+
+export const CURSOR_MARKER = "<|>";
+
+export function getValueAlternatives(
+  meta: CompilerOptionCompletion,
+): readonly string[] | undefined {
+  const values = meta.params?.filter((value) => !value.includes(CURSOR_MARKER));
+  return values && values.length > 0 ? values : undefined;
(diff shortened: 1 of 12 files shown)


packages/language/src/workspace/compilation-unit.ts
```diff
@@ -74,6 +74,7 @@ export interface CompilationUnit {
    */
   uri: URI;
   compilerOptions: CompilerOptions;
+  compilerOptionRanges: Range[];
```
msujew: Suggestion: Can we store this on the compiler options directly?

packages/language/src/preprocessor/compiler-options/completion-pli.ts
```diff
@@ -0,0 +1,279 @@
...
+
+  completion("BACKREG", 1, "5"), // 5 | 11
+  completion("BIFPREC", 1, "31"), // 31 | 15
+  completion("BLANK", 1), // single non-PLI char (string)
+  completion("BLKOFF", 0),
+  completion("NOBLKOFF", 0),
+  completion("BRACKETS", 1), // 2-char string, e.g. '[]'
+
+  completion("CASE", 1, "ASIS"), // UPPER | ASIS
+  completion("CASERULES", 1, "KEYWORD(ASIS)"), // KEYWORD(MIXED|UPPER|LOWER|START)
```
msujew: Bug: This completion is invalid? `ASIS` is never a correct value (and is marked as such).

packages/language/src/preprocessor/compiler-options/completion-pli.ts
```diff
@@ -0,0 +1,279 @@
...
+  completion("NOGONUMBER", 0),
+  completion("GRAPHIC", 0),
+  completion("NOGRAPHIC", 0),
+
+  completion("HEADER", 1, "ALL"), // ALL | FILE | FIRST | SOURCE
+  completion("HGPR", 1, "PRESERVE"), // PRESERVE | NOPRESERVE
+
+  completion("IGNORE", 1, "ASSERT"), // ASSERT | DISPLAY | PUT
+  completion("NOIGNORE", 0),
+  completion("INCAFTER", 1, "PROCESS('')"), // PROCESS('member-name')
```
msujew: Suggestion: As far as I know, a string is not valid for the `INCAFTER(PROCESS(...))` directive. Completing this, results in `Expected a plain text value.`.

packages/language/src/preprocessor/compiler-options/completion-pli.ts
```diff
@@ -0,0 +1,279 @@
...
+  completion("OFFSETSIZE", 1, "4"), // 4 | 8
+  completion("ONSNAP", 1, "STRINGRANGE"), // STRINGRANGE | STRINGSIZE
+  completion("NOONSNAP", 0),
+  completion("OPTIMIZE", 0), // optional: 0 | 2 | 3 | TIME
+  completion("NOOPTIMIZE", 0),
+  completion("OPTIONS", 0), // optional: DOC | ALL
+  completion("NOOPTIONS", 0),
+  completion("OR", 1), // string of up to 7 non-PLI chars
+
+  completion("PP", 1, "CICS('')"), // MACRO|SQL|CICS|INCLUDE with string arg
```
msujew: Suggestion: I would recommend
```suggestion
  completion("PP", 3, "MACRO, SQL, CICS"), // MACRO|SQL|CICS|INCLUDE
```

msujew (commented):
Works mostly well, and I like the API. However, some of the completion recommendations don't work as expected.

packages/language/src/preprocessor/compiler-options/completion-pli.ts
```diff
@@ -0,0 +1,279 @@
...
+  completion("SYSPARM", 1), // string up to 1023 chars
+  completion("SYSTEM", 1, "MVS"), // MVS | CICS | IMS | OS | TSO
+
+  completion("TERMINAL", 0),
+  completion("NOTERMINAL", 0),
+  completion("TEST", 0), // optional: ALL|BLOCK|NONE|PATH|STMT + HOOK|SYM…
+  completion("NOTEST", 0),
+
+  completion("UNROLL", 1, "AUTO"), // AUTO | NO
+  completion("USAGE", 1, "HEX(SIZE)"), // HEX|REGEX|ROUND|SUBSTR|UNSPEC|UUID|VALIDDATE
```
Lotes: Why the default value and the detailed comment?
If this is about completion, would it not be better to have all possible keywords suggested?
What is this good for?

ssmifi: True. Since the current compiler option definitions do not really allow for a complete completion, I did first just add the system for the options without params, but added the defaults afterwards. 
For simple options, I added the complete lists now. 
In other cases you will at least get the auto-complete and if possible the default.

msujew (approved):
Looks quite good, thanks 👍

PR #807, opened 2026-07-23
Preserve original path casing when creating config / adding INCLUDE libs

Wagner-Laranjeiras:
**Preserve original path casing when creating config / adding INCLUDE libs**

Fixes [#793](https://github.com/zowe/zowe-pli-language-support/issues/793).

### Problem
When we auto-generate a startup configuration (missing-config toast + "Create a startup configuration" quick fix) or suggest a folder for the "Add to INCLUDE libs" quick fix, the path was lower-cased before being shown and written into `pgm_conf.json` / `proc_grps.json`. 

### Change
- Added `UriUtils.workspaceRelativeEntryPath(workspace, entry)`: normalizes backslashes and strips the scheme via `toFilePath`, detects workspace membership **case-insensitively**, and returns the relative path with **original casing preserved** (or the absolute path when outside the workspace).
- Refactored `handleMissingConfig` and `quickFixCreateConfig` to use it, replacing two divergent hand-rolled copies.
- Rebuilt `computeWorkspaceRelativeParentFolder` on the same helper so the "Add to INCLUDE libs" quick fix keeps the folder's on-disk casing.

### Notes
- Downstream config resolution is case-insensitive (`minimatch nocase`), so correctly-cased values are safe and now match as *exact* rather than glob.
- Tests: added focused `UriUtils` unit tests (casing preservation, case-insensitive membership, Windows paths, outside-workspace fallback);

### DIFF
--- a/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
+++ b/packages/language/src/language-server/code-actions/apply-quick-fixes.ts
@@ -137,19 +137,10 @@ export async function quickFixCreateConfig(
   if (!workspaceUri || !entryUri) {
     return;
   }
-  const resolvedEntry = UriUtils.toFilePath(entryUri);
-  const workspacePath = UriUtils.toFilePath(workspaceUri);
-
-  const workspaceParts = UriUtils.parts(workspacePath);
-  const entryParts = UriUtils.parts(resolvedEntry);
-
-  const isInsideWorkspace = workspaceParts.every(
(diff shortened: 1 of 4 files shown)


msujew (commented):
Thanks, works well. Please do me a favor and cut down on the jsdoc comments a bit, they are quite wordy.

msujew (approved):
Oops, forgot to approve.

PR #808, opened 2026-07-27
Support dimensions on labels

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/790.

Can also be seen in the `code_samples/PDUMP/S99VFR1.pli` file.

Adds AST/parser and linking support for dimensions on labels. The interesting part: If a `DCL TEST_LABEL(X) LABEL` declaration exists, a label prefix like `TEST_LABEL(1): ...` actually points to that label, so that it can later be referenced using `GOTO TEST_LABEL(1)`.

### DIFF
--- a/packages/language/src/validation/language-server/label-prefix-syntax.ts
+++ b/packages/language/src/validation/language-server/label-prefix-syntax.ts
@@ -1,0 +1,50 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+import { diagnosticFromCode } from "../../language-server/types";
+import { PLICodes } from "../pli-codes";
+import * as AST from "../../syntax-tree/ast";
+import { CompilationUnit } from "../../workspace/compilation-unit";
+import { ValidationAcceptor } from "../validator";
+
+export function checkLabelPrefixSyntax(
+  node: AST.ReferenceItem,
+  acceptor: ValidationAcceptor,
+  compilationUnit: CompilationUnit,
+): void {
+  if (
+    node.dimensions.length === 0 ||
+    node.container?.kind !== AST.SyntaxKind.LabelPrefix
+  ) {
+    // No need to check anything if its not a label, or if there are no dimensions attached
+    return;
+  }
+  if (node.dimensions.length > 1) {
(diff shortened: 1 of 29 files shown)


packages/language/test/fourslash/linker/goto-label-array-prefix-forward.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+// Forward GOTO to an implicitly declared label array, with a dimensioned
+// label prefix directly on the END statement.
+
+//// AVERAGE: PROCEDURE OPTIONS (MAIN);
+////   GOTO <|L>L(2);
```
Lotes: Can you use a descriptive name for the label? :)

packages/language/test/fourslash/linker/goto-label-array-prefix-declared.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+// @wrap: main
+//// DCL <|TEST_LABEL|>(2) LABEL;
+//// <|TEST_LABEL>TEST_LABEL(1): PUT("HELLO");
+//// <|TEST_LABEL>TEST_LABEL(2): PUT("WORLD");
+//// GOTO <|TEST_LABEL>TEST_LABEL(1);
```
Lotes, line 18: Question: Can the array access be combined with a variable read?

GOTO LABELS(INDEX)

 or

LABELS(INDEX+1): statement

msujew: The first one, yes, the second one fails on the compiler, but we currently allow it. I'll adjust that.

packages/language/src/validation/language-server/label-prefix-syntax.ts
```diff
@@ -0,0 +1,50 @@
...
+  const dim = node.dimensions[0];
+  for (const entry of dim.dimensions) {
+    if (entry.lower !== null) {
+      // Can only have a direct reference to a label, not a range
+      acceptor(diagnosticFromCode(PLICodes.Severe.IBM3988I, entry.lower.token));
+      continue;
+    }
+    if (entry.upper !== null) {
+      const value = entry.upper.expression;
+      if (value && value.kind !== AST.SyntaxKind.NumberLiteral) {
```
Lotes: Unfortunately, what if we have negative indices? Then you need also Unaries with `MINUS` sign and a `NumberLiteral` as operand.

Do we have some helper function already for this?

msujew: Good point, seems to be accepted by the compiler - though most other syntax constructs are invalid. I've added a small helper function for this. We can move it somewhere later on if we find more use cases for it.

Lotes (approved):

PR #809, opened 2026-07-29
CICS: Fix lower casing and hovering of preprocessor identifiers

Lotes:
I have not found the ideal place for the upper casing yet.

The hovering was not requested, but I think it makes sense to add it as well.


packages/language/src/language-server/hover-request.ts
```diff
@@ -432,11 +433,28 @@ export function hoverRequest(
...
+      token.element.replacement.kind !== SyntaxKind.IncludeDirective)
+  ) {
+    const ppToken = token.element.preprocessorTokens.find(
+      (t) => t.token.startOffset <= offset && t.token.endOffset >= offset,
+    );
+    if (ppToken) {
+      token = ppToken.token;
+    }
+  }
+
```
msujew: Suggestion: While this makes sense, this is already fixed in my refactoring branch. To prevent merge conflicts, please focus on the lower casing issue in this PR.

See https://github.com/zowe/zowe-pli-language-support/pull/810.

msujew (commented):
Thanks, I think that makes sense. See one comment below.

packages/language/src/language-server/hover-request.ts
Lotes: Done.

msujew (approved):
Thanks, looks good. I'll integrate this directly into #810 then to make sure that the test runs as expected!

PR #810, opened 2026-08-03
Preprocessor refactoring

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/811
Closes https://github.com/zowe/zowe-pli-language-support/issues/771

This is a major refactoring of the whole preprocessor pipeline. It includes:

* Implement a clear separation between all preprocessor phases. Each phase is now `text -> {text, SourceMap}` (i.e. text in, new text+SourceMaps out). Handles translation of offsets from one phase to the next.
* Refactors external preprocessors to use the full input text to perform its preprocessor.
* Adds a lot of tests and benchmarks to ensure that this is working correctly and the performance is acceptable. Based on the benchmarks, this also includes a few performance improvements to lexing (since this is now a larger focus).
* Adds a command to PL/I files to look at the generated text output of the preprocessors. This allows users to more easily debug their preprocessors. However, the MACRO preprocessor isn't doing this perfectly, so the output doesn't look like the one in the compiler. Also, it strips all comments.

Note: I've left out the DFHRESP/SQL TYPE IS from the external preprocessors. They are handled in the correct step (i.e. only when `CICS`/`SQL` is set in the `PP` compiler option), but aren't part of the respective package yet.

Generally, how this works in the LSP context is that after all preprocessor passes, the original input tokens are annotated with the "post-preprocessing" information via source maps. This means that all LSP services only need to use the token binary search to find the correct input token (which is now annotated with the expected info) to do its work. No longer any need to do any special offset lookup inside of EXEC tokens or similar.


packages/preprocessor-db2/src/engine/parsing.ts
```diff
@@ -69,12 +69,17 @@ export class CollectingErrorListener extends BaseErrorListener {
...
 
 export class CollectingIdentifierVisitor extends Db2SqlExecParserVisitor<void> {
+  // Reused across parses: constructing a generated visitor initializes one
+  // per-instance field per grammar rule, which dominates the profile when done
+  // once per EXEC statement.
+  private static readonly instance = new CollectingIdentifierVisitor();
   static collect(tree: ParseTree): Token[] {
-    const visitor = new CollectingIdentifierVisitor();
+    const visitor = CollectingIdentifierVisitor.instance;
+    visitor.identifiers = [];
```
Lotes, line 79: Looks a bit like a code smell. This only works because the code is not async.
You can still call `tree.accept(CollectingIdentifierVisitor.instance)` afterwards to mutate the list.

Lotes: Same for the following visitor.

packages/preprocessor-db2/test/context-execution.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    expect(context.diagnostics).toHaveLength(0);
+    expect(context.edits).toHaveLength(1);
+    const [edit] = context.edits;
+    expect(text.slice(edit.range.start, edit.range.end)).toBe(
+      "EXEC SQL SELECT 1 INTO :HV1 FROM T;",
+    );
+    // The host variable appears verbatim in the replacement, in token order - the contract
+    // the host's embedded-image search relies on.
+    expect(edit.text).toContain("HV1");
```
Lotes, line 37: As far as I understand from this test: the `edit.text` content is `HV1; DO END;`.
What happens with:

```
IF condition
  THEN EXEC SQL ...;
  ELSE EXEC SQL ...;
```

? Would the result not end up in a dangling DO-END? With dangling I mean that `HV1` is consumed for THEN/ELSE and the DO-END is the follow-up statement that is outside the control flow hierarchy.

Also: I am not sure for SQL: Is there also a `RETURN` like in CICS?

packages/preprocessor-db2/test/context-execution.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    expect(edit.text).toBe("");
+    const member = edit.tokens.find(
+      (t) => t.semanticsKind === SemanticsKind.Identifier,
+    );
+    expect(member?.image).toBe("COPY1");
+    expect(member?.startOffset).toBe(text.indexOf("COPY1"));
+  });
+
+  test("an unterminated statement is parsed and diagnosed but not replaced - only annotated", async () => {
+    const text = "EXEC SQL SELECT 1 INTO :HV1 FROM T";
```
Lotes, line 68: Ah the semicolon is intentionally missing!
But we are still extracting identifiers... felt a bit weird in the first moment.

packages/language/src/language-server/preprocessed-text.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (cached !== compilationUnit.preprocessedText) {
+    compilationUnit.requestCaches.set(
+      "preprocessedText",
+      compilationUnit.preprocessedText,
+    );
+    const uris = new Set(compilationUnit.services.files.keys());
+    uris.add(compilationUnit.uri.toString());
+    // Direct send instead of the connection-handler helper to avoid an import
+    // cycle (compilation-unit -> here -> connection-handler -> compilation-unit).
+    connection.sendNotification(Messages.PreprocessedTextChanged.method, {
```
Lotes: Wasn't there a typed `sendNotification`?

scripts/check-license-headers.mts
```diff
@@ -15,10 +15,14 @@ import { glob } from "glob";
 
 const fixFlag = process.argv.includes("--fix");
 
-const header = await readFile("license-header.js", "utf-8");
+const header = (await readFile("license-header.js", "utf-8")).replace(
```
Lotes, line 18: Ah! Was this relevant for you on Windows?
For the GH action I used non-Windows.

packages/preprocessor-cics/test/context-execution.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    const context = new RecordingPreprocessorContext(text);
+    await preprocessor.execute(context);
+
+    expect(context.diagnostics).toHaveLength(0);
+    expect(context.edits).toHaveLength(1);
+    const [edit] = context.edits;
+    expect(text.slice(edit.range.start, edit.range.end)).toBe(
+      "EXEC CICS ABEND ABCODE(12);",
+    );
+    expect(edit.text.endsWith("DO; END;")).toBe(true);
```
Lotes: CICS also has a RETURN. We should have a test for this I believe...

packages/preprocessor-cics/src/engine/preprocessor.ts
```diff
@@ -46,6 +51,19 @@ import {
...
+/**
+ * Confirmed against `CICSLexer.g4`: `'...'`/`"..."` (each escaped by doubling its own quote,
+ * never spanning a line) cover every CICS string form here too - the `X`/`Z`/`G`/`N` prefixes
+ * on hex/null-terminated/DBCS literals are just an ordinary character preceding the quote.
+ * Comments: `*>`/`>>`/`//` run to end of line; `/* *\/` is the one construct that can span
+ * multiple lines.
+ */
+const CICS_DELIMITERS: Delimiters = {
+  quotes: ["'", '"'],
+  lineComments: ["*>", ">>", "//"],
```
Lotes, line 63: ">>" really? Never saw this O.O

packages/preprocessor-cics/src/engine/collect-semantic-errors.ts
```diff
@@ -20,6 +20,20 @@ import { Diagnostic } from "preprocessor-api";
...
+  // Reused across parses: the constructor builds an OptionsRegistry with ~100
+  // checker instances (each copying option maps), which dominates the profile
+  // when done once per EXEC statement. The checkers keep a reference to
+  // `errors`, so `collect` drains the shared array instead of replacing it.
+  private static readonly instance = new CollectingSemanticErrorVisitor();
+  static collect(tree: ParseTree): Diagnostic[] {
+    const visitor = CollectingSemanticErrorVisitor.instance;
+    // Reset at entry: a checker throwing mid-visit must not leak stale errors
+    // into the next parse (the array is shared with the registry's checkers).
+    visitor.errors.length = 0;
```
Lotes, line 34: This one looks more hackier than the other collectors O.O!
How big is the performance gain?

packages/preprocessor-api/src/recording-context.ts
```diff
@@ -0,0 +1,63 @@
...
+ * A `Preprocessor`'s complete observable output flows through the interface's members
+ * (edits with their classified token lists, diagnostics, include resolutions), so running
+ * `execute` against this recorder and asserting on what was recorded is a full conformance
+ * check of a preprocessor implementation - including a future external one talking over a
+ * serialized boundary. `resolveInclude` records the attempt and reports it unresolved
+ * (returns `undefined`); `insertContext` records the call.
+ */
+export class RecordingPreprocessorContext implements PreprocessorContext {
+  readonly diagnostics: Diagnostic[] = [];
+  readonly edits: RecordedEdit[] = [];
```
Lotes, line 39: Edits are used for replacing EXEC fragments with new code.
Insertions are used to insert other INCLUDED files?

packages/language/src/linking/resolver.ts
```diff
@@ -523,37 +521,7 @@ export function getReferenceLocations(
 }
 
 export function getTokenAt(unit: CompilationUnit, uri: URI, offset: number) {
-  let token = binaryTokenSearch(
```
Lotes, line 533: Nice removal

packages/language/test/test-builder.ts
```diff
@@ -1110,6 +1111,64 @@ Available code actions for label "${label}" and URI "${uri}": ${codeActions.map(
...
+   * every reference marker of the same index - across all test files.
+   * @returns This test builder
+   *
+   * @example
+   * ```ts
+   * new TestBuilder(`
+   *  DCL <|1:A|>;
+   *  PUT(<|1>A);
+   *  PUT(<|1>A);
+   * `).expectReferences(); // Passes: references at A's declaration yield 3 locations
```
Lotes, line 1127: What is the difference to `expectLinks`?
I guess it is just a different entry point into the LS, or?

Lotes (commented):
Some intermediate comments. Still 64 files to review.

packages/language/src/language-server/semantic-tokens.ts
```diff
@@ -82,46 +81,27 @@ export function semanticTokens(
...
-        index < token.element.preprocessorTokens.length;
-        index++
-      ) {
-        const subToken = token.element.preprocessorTokens[index].token;
-        const type = token.element.preprocessorTokens[index].semanticType;
-        push(subToken, subToken.image.length, type);
-      }
+    // Tokens a preprocessor phase classified itself (e.g. inside an `EXEC SQL`/`EXEC CICS`
+    // statement's body) carry their type directly - see `Token.semanticType`.
+    if (token.semanticType !== undefined) {
```
Lotes: Can we rename this to `ppSemanticType`? It is a it confusing to see this condition which is only valid for preprocessor-sourced tokens.

packages/language/src/preprocessor/comment-stripper.ts
```diff
@@ -0,0 +1,181 @@
...
+ * text exactly once.
+ *
+ * String literals are skipped whole (reusing the real tokenizer's own `STRING_TERM` pattern)
+ * so a comment-looking sequence inside a string is never mistaken for a real comment.
+ *
+ * Linear in the length of `text`: a single left-to-right pass, with blanking done once per
+ * comment span (never re-scanning already-processed text), matching the perf requirements of
+ * the tokenize hot path.
+ */
+export function stripComments(text: string): StripCommentsResult {
```
Lotes, line 67: Question: Maybe you know it... what about Unicode in comments? Will all bytes also be blanked? I am imagining a Unicode character with [multiple code points](https://gist.github.com/Aldaviva/f8980d70eb91dd16426333e73b04bacb) in it. My spontanous assumption was that they are not equally replaced, which would lead to an offset mismatch in the following phases.

packages/language/src/preprocessor/macro-phase.ts
```diff
@@ -98,10 +109,12 @@ export function createMacroHandlers(
 }
 
 /**
- * Macro preprocessor phase. A `Token[] -> Token[]` transformation that parses `%`
- * statements and includes from its input tokens, runs the macro instructions (variable
- * expansion, `%IF`/`%DO`, `%INCLUDE`, `REPLACE`, ...) and passes EXEC SQL/CICS tokens
- * through with macro variables expanded.
+ * Macro preprocessor phase. A `{text, sourceMap} -> {text, sourceMap}` transform: tokenizes
```
Lotes, line 112: Would it make sense to give each phase an own folder containing all relevant classes and helpers under "src/preprocessor/phases/<phase>"?

packages/language/src/preprocessor/macro-phase.ts
```diff
@@ -137,16 +156,65 @@ export class MacroPreprocessorPhase implements PreprocessorPhase {
...
+  text: string,
+  opts: PliCompilerOptions | undefined,
+): boolean {
+  if (opts?.incAfter?.process) {
+    return true;
+  }
+  if (text.includes("%")) {
+    return true;
+  }
+  const includeAlt = opts ? getEffectiveIncludeAlt(opts) : undefined;
```
Lotes, line 217: Can you explain the Regexp? What is it testing? What is `includeAlt`?

packages/language/src/preprocessor/source-map.ts
```diff
@@ -0,0 +1,409 @@
...
+  /**
+   * A map for text that has not been touched by any preprocessor: every offset maps to
+   * itself. Used to seed the pipeline before the first phase runs.
+   */
+  static identity(text: string, uri?: URI): SourceMap {
+    const length = text.length;
+    return new SourceMap([
+      {
+        origStart: 0,
+        origEnd: length,
```
Lotes, line 132: For `origEnd` and `genEnd` it would make sense to add a comment if they are inclusive or exclusive. I think they are exclusive, when they point to `length` here.

packages/language/src/preprocessor/source-map.ts
```diff
@@ -0,0 +1,409 @@
...
+
+  /**
+   * Composes two maps: `first` maps `text -> A`, `second` maps `A -> B`; the result maps
+   * `text -> B`. This is how per-phase maps accumulate into one map from the original
+   * document to the final preprocessed text.
+   *
+   * Implemented as a merge over `second`'s segments (in `A`/generated-of-first space),
+   * each positioned into `first` by binary search - `second`'s verbatim segments are NOT
+   * guaranteed to advance monotonically through `A` (a `%DO` re-emission rewinds; see
+   * `serializeTokens`): verbatim spans are subdivided against `first`'s segment boundaries so untouched
```
Lotes: Make this comment as a list of actions. It is hard to read for me.
Maybe too much? Because each case below is handled with comments as well?

packages/language/src/preprocessor/source-map.ts
```diff
@@ -0,0 +1,409 @@
...
+        // context's own non-verbatim edits are just as "foreign" as its verbatim text.
+        composed.push({ ...secondSegment });
+        continue;
+      }
+      if (!secondSegment.verbatim) {
+        composed.push(composeNonVerbatimSegment(first, secondSegment));
+        continue;
+      }
+
+      let a = secondSegment.origStart;
```
Lotes: Find a better name. Maybe `original<SOMETHING>`?

packages/language/src/preprocessor/source-map.ts
```diff
@@ -0,0 +1,409 @@
...
+      // below still advances linearly.
+      let firstIndex = rightmostIndexLE(firstSegments, a, segmentGenStart);
+      if (firstIndex === -1) {
+        // `a` lies before `first`'s first segment - outside `first`'s generated space,
+        // nothing to resolve through (mirrors running past the last segment below).
+        continue;
+      }
+
+      while (a < secondSegment.origEnd) {
+        const fs = firstSegments[firstIndex];
```
Lotes: `fs` can be a better name.

packages/language/src/preprocessor/source-map.ts
```diff
@@ -0,0 +1,409 @@
...
+   *
+   * Implemented as a merge over `second`'s segments (in `A`/generated-of-first space),
+   * each positioned into `first` by binary search - `second`'s verbatim segments are NOT
+   * guaranteed to advance monotonically through `A` (a `%DO` re-emission rewinds; see
+   * `serializeTokens`): verbatim spans are subdivided against `first`'s segment boundaries so untouched
+   * source keeps precise, offset-accurate positions across any number of phases; non-verbatim
+   * (replaced/inserted) spans stay atomic blocks, per the `Segment` contract - only their
+   * anchor is resolved through `first`.
+   */
+  static compose(first: SourceMap, second: SourceMap): SourceMap {
```
Lotes, line 203: Puh, `compose` hard to understand currently - for me at least.
I am not so sure about `foreign` and `verbatim`. It could help to extract condition and action helper functions. Like:

```
if(condition1(...)) action1(...);
else if(condition2(...)) action2(...);
else action3();
```

The name of the condition would describe the situation, the name of the action describes the fix.

On the other hand the code is not sooo big. From line 256 (`let a = secondSegment.origStart`) I am not sure how the code ist meant.

---

But the general purpose of the function is clear.

packages/language/test/preprocessor/source-map.test.ts
```diff
@@ -0,0 +1,555 @@
...
+        origEnd: 6,
+        genStart: 4,
+        genEnd: 10,
+        uri: uriA,
+        verbatim: false,
+        tokens: [
+          {
+            name: "A",
+            startOffset: 4,
+            endOffset: 9,
```
Lotes, line 155: There is some kind of inconsistency, right?

Token offsets end inclusively, segments end exclusively.

Can lead to confusion, I think.

packages/language/src/preprocessor/token-annotator.ts
```diff
@@ -0,0 +1,243 @@
...
+  );
+  // Every token that logically belongs to the same file must share the exact same `URI`
+  // object, not just an equal string - some existing consumers (e.g.
+  // `stringify.ts#extractDeclaration`, checking a multi-token declaration's start/end are
+  // "the same file") compare by reference. In the old pipeline that held for free (one
+  // `tokenize()` call per file handed every token the same `uri` instance); segments now
+  // assign `.uri` independently per span, so without interning, two tokens from the same
+  // included file could end up with two different (if `.toString()`-equal) `URI` objects.
+  const uriCache = new Map<string, URI>();
+  const internUri = (uri: URI | undefined): URI => {
```
Lotes, line 43: Is it not intern**al**Uri?

packages/language/test/fourslash/linker/host-variable-special-chars.fail.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+/// <reference path="../framework.ts" />
+
+/**
+ * Go-to-definition on an `EXEC SQL` host-variable whose name contains the PL/I extra
+ * identifier characters (`#`) resolves to its `DCL` - the embedded-image lookup must
+ * treat `#`/`@`/`$` as identifier characters, not word boundaries.
+ *
+ * Marked `.fail.`: the host-side embedded-image lookup (`findEmbeddedImage`) handles
```
Lotes: Mark with TODO or add an issue?

packages/language/test/fourslash/preprocessor/include/exec-sql-include.ts
```diff
@@ -12,12 +12,13 @@
 /// <reference path="../../framework.ts" />
 
 // @filename: cpy/lib.pli
-//// DECLARE LIB_VAR FIXED;
+////<|click:|> DECLARE LIB_VAR FIXED;
```
Lotes, line 15: Why is this important? It is unusual to use the reference of `click` like that.

packages/language/test/fourslash/preprocessor/pp-phases/exec-sql-empty-statement.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+
+/// <reference path="../../framework.ts" />
+
+/**
+ * An empty `EXEC SQL;` statement is still recognized by `scanExecFragments` (the tokenizer
+ * consumes it as an EXEC statement, so the scan must too) and replaced like any other
+ * statement - the raw `EXEC SQL` tokens must not leak to the final parse, where they would
+ * only produce generic grammar errors. The SQL engine may report its own diagnostic for the
```
Lotes, line 18: Can we check the diagnostic for the empty statement plz? :)

packages/language/test/parser/pli-lexer.test.ts
```diff
@@ -49,6 +49,29 @@ describe("PL/1 Lexer", () => {
...
+            dcl A fixed bin(31);
+            dcl C fixed bin(31);
+        `;
+    const document = TextDocument.create(uri.toString(), "pli", 0, text);
+    const unit = await createCompilationUnit(uri, defaultTestWorkspace());
+    const lexer = new PliLexer();
+    const result = await lexer.tokenize(unit, document, uri);
+    // The macro replaced A with B in the final text.
+    expect(result.preprocessedText).toContain("B");
+    expect(result.preprocessedText).not.toContain("%decl");
```
Lotes: Do you mean %DCL or %dcl?

packages/language/test/preprocessor/preprocessor-context.test.ts
```diff
@@ -0,0 +1,451 @@
...
+    expect(sourceMap.mapToOriginal(5)).toEqual({ uri, offset: 4 }); // "B" right after ONE
+    expect(sourceMap.mapToOriginal(11)).toEqual({ uri, offset: 9 }); // "C" right after TWO
+  });
+
+  test("an overlapping edit is dropped with a diagnostic instead of throwing", async () => {
+    const context = await createContext("ABCDEFGH");
+    context.replace({ start: 2, end: 5 }, "X");
+    context.insert(3, "Y"); // falls inside the previous replace's range
+    // Edits come from external preprocessor plugins - a throw here would kill
+    // tokenization of the whole document, so build() must degrade instead.
```
Lotes, line 145: Very interesting! But when does this happen in the real world? Do you have an example? Another question: The diagnostic goes from 3 to exclusively 3, so basically just a position instead of a range. Is this displayable and desired to have?

packages/language/src/preprocessor/exec-phase.ts
```diff
@@ -10,168 +10,879 @@
...
+    token.image,
+    t.ID,
+    token.startOffset,
+    token.endOffset,
+    uri,
+  );
+}
+
+function toSemanticType(kind: SemanticsKind): SemanticTokenTypes {
+  switch (kind) {
```
Lotes: Just because it pops up in my mind now: We could replace mappings like this with a constant record (`Record<SemanticsKind, SemanticTokenTypes>`).
Not sure if it is much faster, but it is more compact code ^^*. A `switch` looks like shooting with a cannon on birds, or? There is no real logic, it is just a mapping.

packages/language/src/preprocessor/exec-phase.ts
```diff
@@ -10,168 +10,879 @@
  */
 
 import { TextDocument } from "vscode-languageserver-textdocument";
+import {
+  Preprocessor,
+  SemanticsKind,
+  Token as ApiToken,
```
Lotes, line 16: `ApiToken` sounds misleading. Did you mean something like `PreprocessorToken`?
The API part classifies the package as an interface to the preprocessor. The preprocessing is the important part in my eyes.
We could have many different API thingies.

packages/language/src/preprocessor/preprocessor-context.ts
```diff
@@ -0,0 +1,561 @@
...
+   * from - see `collectExecMetadata`.
+   */
+  apiTokens?: api.Token[];
+  /**
+   * Which `MappedToken` each re-embedded `apiTokens` identifier became (identifiers whose
+   * image was located in `text` - see `findEmbeddedImage`). `collectExecMetadata` sets
+   * each pair's `mapped.sourceToken` from the PL/I token it builds for the same api token,
+   * so host-variable go-to-definition lands on real source positions.
+   */
+  identifierPairs?: { apiToken: api.Token; mapped: MappedToken }[];
```
Lotes, line 143: I am a bit wondering about this field here... `api.Token` and `MappedToken` are also used in own lists. Would be indices to these lists not enough?

Lotes (commented):
I made it through all files now. Tough reading. Especially files with more than 400 added lines. I will test the extension after some pausing time.

packages/preprocessor-db2/src/engine/parsing.ts
msujew: The issue is that instantiating a new visitor every parse is incredibly expensive. In my benchmarks, half of the total time in `collect` is used to instantiate the visitor. I'll wrap this in a mutex to ensure that we don't call this multiple times simultaneously.

packages/preprocessor-db2/test/context-execution.test.ts
msujew: Right, that's not quite correct. I'll put those replacement IDs into a `PUT` statement 👍

packages/preprocessor-db2/src/engine/parsing.ts
msujew: Actually, no mutex necessary, the code is completely sync. I'm not sure what do here. It should be fine as it is.

scripts/check-license-headers.mts
msujew: Yeah, I couldn't figure out why the linting CI failed, since I couldn't run it locally. After running it locally, I figured out the issue (wrong UTF8-BOM encoding on the file in question).

packages/preprocessor-cics/test/context-execution.test.ts
msujew: Right, but `RETURN` also just generates a `DO; END;`. We don't generate the "real" preprocessor code for CICS and SQL yet (but should do so in the future).

packages/preprocessor-cics/src/engine/preprocessor.ts
msujew: I wasn't aware of this either, but it checks out. Runs on the compiler.

packages/preprocessor-cics/src/engine/collect-semantic-errors.ts
msujew: Roughly 50%. It's a lot!

packages/language/src/preprocessor/preprocessor-context.ts
msujew: Possible, but since these are just pointers, I don't think it really matters?

packages/language/src/preprocessor/exec-phase.ts
msujew: It's the token from the v2 preprocessor API. I think that fits to be honest.

packages/preprocessor-api/src/recording-context.ts
msujew: Yes, exactly. We could probably solve it using the same API? The issue is that the info where tokens come from (i.e. the uri) get lost when just using `replace`.

packages/language/test/test-builder.ts
msujew: `expectLinks` performs the "go to definition" logic, while `expectReferences` does the "find references" logic. They are not the same.

packages/language/src/preprocessor/comment-stripper.ts
msujew: Added a test for this.

packages/language/src/preprocessor/macro-phase.ts
msujew: I would do this as a follow up maybe.

packages/language/src/preprocessor/macro-phase.ts
msujew: `includeAlt` is the value of the `PP(INCLUDE(...))` value that can be used to define a separate include statement. I.e. `PP(INCLUDE('ID(++include)'))` allows you to use `++include lib` to perform an `%INCLUDE lib;`. Don't ask me, why that exists, I don't know either - we have to support it.

packages/language/src/preprocessor/token-annotator.ts
msujew: `internUri` performs string **interning**. It makes sure that we don't generate multiple URI objects for the same string.

packages/language/test/fourslash/preprocessor/include/exec-sql-include.ts
msujew: It's important, because the behavior broke in the PR (i.e. clicking on an `INCLUDE` brings you to the first line of that included file). We had no tests that made sure that this actually works as expected.

packages/language/test/preprocessor/preprocessor-context.test.ts
msujew: Doesn't really happen, but the context shouldn't just throw because the external preprocessor (which we cannot control) provides invalid ranges.

packages/language/src/preprocessor/macro-phase.ts
Lotes: ok, just if it is worth it. If we end up with 3 folders with one file each, we can skip it.

packages/language/src/preprocessor/token-annotator.ts
Lotes: Never heard this verb. Ok.

packages/language/src/language-server/constants.ts
```diff
@@ -60,4 +60,10 @@ export namespace Commands {
    * program configuration is bound to.
    */
   export const GO_TO_PROCESS_GROUP = "pli.goToProcessGroup";
+
+  /**
+   * Client-side command to open the fully preprocessed text of the
+   * compilation unit containing the active .pli file in a read-only editor.
+   */
+  export const SHOW_PREPROCESSED_TEXT = "pli.showPreprocessedText";
```
ssmifi, line 68: Cool. So it could be possible in the future to show the state after each PP step? :)

packages/language/src/language-server/rename-request.ts
```diff
@@ -12,20 +12,50 @@
...
-): Record<DocumentUri, Location[]> {
+): RenameResult {
+  // A symbol whose declaration only exists in preprocessor-generated text (e.g. the CICS
+  // `DFH*` declarations) cannot be renamed: the declaration is regenerated on every run,
+  // so the rename can never take effect - refuse instead of silently renaming only the
+  // real-source usages.
+  const token = getTokenAt(unit, uri, offset);
+  if (token) {
+    const element = findTokenElementReference(token);
+    const nameToken = element && getNameToken(element);
```
ssmifi: _Question_: Is it possible to generate an empty edit, e.g., if `getTokenAt` returns nothing (if that is possible)? This would result in `{ kind: "edits", changes: {} }` which will probably just be reported as successful rename. I was just wondering if that is possible and if we distringuish the failure cases more carefully if this could be another one.

packages/language/src/language-server/rename-request.ts
```diff
@@ -12,20 +12,50 @@
...
+    if (nameToken?.synthetic) {
+      return { kind: "generated", name: nameToken.image };
+    } else if (nameToken?.uri?.scheme === BuiltinsUriSchema) {
+      return { kind: "builtin", name: nameToken.image };
+    }
+  }
+
+  // Locations of `synthetic` usage tokens are already suppressed by
+  // `getReferenceLocations`, so the edit set only touches real source text.
   const references = getReferenceLocations(unit, uri, offset);
```
ssmifi: _Minor_: `getReferenceLocation` does the same `getTokenAt` and `findTokenElementReference`, which is a duplication of work here. Would it make sense to allow an optional `element` parameter here to skip that step?

packages/language/src/parser/tokenizer/shared.ts
```diff
@@ -258,42 +246,47 @@ export function tokenizeSlashWithComment(
 
     if (nextChar === "*") {
       // Block comment
```
ssmifi: Can we remove this comment line? :)

packages/language/src/parser/tokenizer/shared.ts
```diff
@@ -23,16 +23,20 @@ export class TokenizerContext {
...
   private storedIndex: number = 0;
-  private storedLine: number = 0;
-  private storedColumn: number = 0;
+  /**
+   * Whether a line break has been seen since the last token (or comment) was created.
+   * Cheap to maintain (a single flag flip while scanning whitespace/comments, which the
+   * tokenizer already does character-by-character) and replaces per-token line/column
+   * tracking - see `Token.startsNewLine`. Public (like the other tokenizer-internal fields
+   * on this class) so comment-scanning functions in this module can set it directly.
+   */
```
ssmifi, line 35: _Suggestion_: In general, it is ofc nice to have detailed comments. In this file, personally, it is a bit too much, so I would suggest to check all comments for redundancy or too deep information. E.g., the `public` note here, the collision lesson at the hash function, or duplicated information in comments.

packages/language/src/preprocessor/comment-stripper.ts
```diff
@@ -0,0 +1,181 @@
...
+ * Blanks every `/ *  ... * /`-style block comment and `//...` line comment in `text` to
+ * whitespace, preserving every other character's offset - mirrors `PliMarginsProcessor`'s
+ * length-preserving blanking, so the result can still be sliced/indexed with the original
+ * offsets everywhere downstream.
+ *
+ * This exists so external SQL/CICS preprocessors - which scan the *full text* of an `EXEC`
+ * fragment rather than tokens - never see comment characters embedded in `EXEC` code (a
+ * comment could contain unbalanced quotes or a stray `EXEC SQL`-looking substring). It runs
+ * once, on the margin-stripped *original* source text, before any preprocessor phase runs.
+ * It is unrelated to the MACRO preprocessor's "unfinished token" problem (a comment opened by
```
ssmifi: Is specific issue information something that should be included in the functions doc?

packages/language/src/preprocessor/comment-stripper.ts
```diff
@@ -0,0 +1,181 @@
...
+
+/** Replaces `text[start, end)` with spaces, except newline characters, which are kept as-is. */
+function blank(text: string, start: number, end: number): string {
+  const parts = new Array<string>(end - start);
+  for (let k = start; k < end; k++) {
+    const code = text.charCodeAt(k);
+    parts[k - start] =
+      code === LINE_FEED || code === CARRIAGE_RETURN ? text[k] : " ";
+  }
+  return parts.join("");
```
ssmifi: Functionally ok. Just wondering: 
_Question_: Since comments mostly contain non lf/cr characters and also often non of them, could we optimize this per-character approach (especially for large comment blocks which are not too unusual for e.g. SQL files)). Don't know if it is worth it, but maybe something to consider for later optimizations.

packages/language/src/preprocessor/exec-phase.ts
```diff
@@ -10,168 +10,920 @@
...
   };
 }
 
+function getGenerationCacheEntry(
+  cache: Map<number, GenerationCache>,
+  offset: number,
+): GenerationCache {
+  let entry = cache.get(offset);
+  if (!entry) {
+    entry = {
```
ssmifi, line 358: _Minor_: The cache always creates all fields even though a given entry ios really owned by either the CICS or the SQL phase. Probably fine since the cache is never shared, but could perhaps be modelled more concretely.

packages/language/src/preprocessor/macro-phase.ts
```diff
@@ -137,16 +156,68 @@ export class MacroPreprocessorPhase implements PreprocessorPhase {
...
+            ...output.execTokens.filter(
+              (t) => t.uri?.toString() === uri.toString(),
+            ),
+          ],
+          input.sourceMap,
+        ),
+        // Fragments inside %INCLUDEd files already carry file-local offsets; the
+        // entry file's source map must not remap them.
+        ...output.execTokens.filter(
+          (t) => t.uri && t.uri.toString() !== uri.toString(),
```
ssmifi: _Minor_: Both filter results can be collected in one run.

packages/language/src/preprocessor/pli-lexer.ts
```diff
@@ -172,31 +206,132 @@ export class PliLexer {
...
+    // identity-dedupe as above.
+    const foreignTokens = annotated.tokens.filter(
+      (t) => !t.synthetic && t.uri && t.uri.toString() !== uriString,
+    );
+    this.mergeForeignTokens(
+      unit,
+      [...new Set([...directiveTokens, ...foreignTokens])],
+      uriString,
+    );
+  }
```
ssmifi, line 254: I am still a bit conflicted w.r.t. these overly detailed comments. Sure, in general detailed comments are nice, but I also get the feeling that there is a lot of overlap here and I am drowning in comments. :)

packages/language/src/preprocessor/preprocessor-context.ts
```diff
@@ -0,0 +1,587 @@
...
+    return nested;
+  }
+
+  /** Mirrors the diagnostic the MACRO preprocessor's include handling raises on failure. */
+  private pushUnresolvedIncludeDiagnostic(name: string, range?: Range): void {
+    const missingConfiguration =
+      !this.unit.processGroup && !this.unit.programConfig;
+    // Without a range the diagnostic has no position and `DiagnosticsStore` drops it -
+    // anchor it to the include statement whenever the caller told us where that is.
+    const diagnostic = range
```
ssmifi: _Minor_: Four-way ternary testing is a bit dense here. Can we write that more clear?

packages/language/src/utils/messages.ts
```diff
@@ -34,6 +36,42 @@ export function createNotificationType<P>(method: string): NotificationType<P> {
...
+export function onRequest<P, R>(
+  connection: Connection,
+  type: RequestType<P, R>,
+  handler: (params: P) => R | Promise<R>,
+): void {
+  connection.onRequest(type.method, handler);
+}
+
+export function sendRequest<P, R>(
+  connection: Connection,
```
ssmifi, line 52: _Question_: The `plugin-configuration-provider` still uses the `sendRequest` from the `connection-handler`. Could/Should it use this one?

packages/language/src/workspace/compilation-unit.ts
```diff
@@ -383,6 +399,11 @@ export class CompilationUnitHandler {
     textDocuments.listen(connection);
     textDocuments.onDidChangeContent((event) => {
       const uri = UriUtils.toUri(event.document.uri);
+      if (uri.scheme === PreprocessedTextUriSchema) {
+        // Completely reject open notifications for full text output
+        // It should purely be highlighted by textmate, and receive to language support
```
ssmifi: _Questions_: 
"no" language support?
Here on the other hand, I would like to get more infos. Why not? Is this something conceptual or could we provide support in the future?

packages/language/test/parser/pli-tokenizer.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    // (computed by `scanExecFragments`) against the ExecFragment token's offsets, so
+    // the two scans must agree on the statement extent - in particular for bodies
+    // containing quoted semicolons or unterminated strings.
+    const delimiters = { quotes: ["'", '"'], lineComments: ["--"] };
+    const cases = [
+      "EXEC SQL SELECT ';' FROM T;",
+      'EXEC SQL SELECT ";" FROM T;',
+      "EXEC SQL INSERT INTO T VALUES('it''s;fine');",
+      "EXEC SQL INSERT INTO T VALUES(';EXEC SQL X;');",
+      "EXEC SQL SELECT 'unterminated\nFROM T;",
```
ssmifi, line 169: _Suggestion_: While reading here... all tests here cover `\n` line endings. I suggest to add at least one test covering `\r\n` as well.

packages/language/test/preprocessor/source-map.test.ts
```diff
@@ -0,0 +1,555 @@
...
+    const leftAssoc = SourceMap.compose(SourceMap.compose(m1, m2), m3);
+    const rightAssoc = SourceMap.compose(m1, SourceMap.compose(m2, m3));
+
+    for (let i = 0; i < 12; i++) {
+      expect(leftAssoc.mapToOriginal(i)).toEqual(rightAssoc.mapToOriginal(i));
+    }
+  });
+
+  test("a foreign verbatim segment survives composition atomically, not resolved through first", () => {
+    // A real bug: a foreign segment's origStart/origEnd are real offsets in *another*
```
ssmifi: Was this a real bug? Should this be mentioned here?

packages/preprocessor-api/src/context-utils.ts
```diff
@@ -0,0 +1,213 @@
...
+ * Re-embedding the literal name (rather than just carrying it as metadata) is required by
+ * the real parser's `execVariableReference` rule, which recognizes exactly this shape
+ * (a synthetic marker immediately before a real identifier token) to resolve host-variable/
+ * reference clicks back to their declaration - see `token-annotator.ts`'s
+ * `MappedToken.execHostVariable` handling in the language package. The host locates each
+ * identifier inside the replacement text itself (in-order, verbatim - the contract this
+ * function's output upholds), so only the text is returned; the classified token list goes
+ * to `PreprocessorContext.replace` separately, in host coordinates.
+ */
+export function buildExecReplacement(tokens: Token[]): string {
```
ssmifi, line 199: Is the comment still up to date? I cannot match this to the actual function.

packages/preprocessor-cics/test/check.test.ts
```diff
@@ -17,14 +17,14 @@ describe("CICS CHECK", async () => {
   const cicsPreprocessor = new CICSPreprocessor(HostLanguageType.PLI);
 
   test("Positive (ACTIVITY)", async () => {
-    const { diagnostics } = await cicsPreprocessor.execute(
+    const { diagnostics } = await cicsPreprocessor.parse(
```
ssmifi: All these `await`s in the test are superfluous now. (Also in the other files.)

packages/vscode-extension/package.json
```diff
@@ -38,9 +38,22 @@
...
+        "title": "Show Preprocessed Text",
+        "category": "PL/I",
+        "icon": "$(file-code)"
       }
     ],
     "menus": {
+      "editor/title": [
+        {
+          "command": "pli.showPreprocessedText",
+          "when": "resourceLangId == pli",
```
ssmifi: _Suggestion_: Not tested, but could we add a `resourceLangId == pli && resourceScheme != pli-preprocessed"` to prevent the button from appearing in preprocessor documents?
We should maybe also exit early in the handler if it is a preprocessor document.

ssmifi (approved):
Solid, well-tested refactor.  I only found minor, non-blocking issues. Thank you.
The comment density is quite high though and personally is reaching a point where the code actually becomes harder to read. I would suggest to consider trimming it down a bit. Especially overly detailed overlapping explanations.

Show Preprocessed Text is very nice. 👍

packages/vscode-extension/package.json
ssmifi: Clicking on the button in a `(preprocessed)` editor tab actually opens an empty new one. So, if possible, I would try to avoid that.

Lotes (approved):
I was looking quickly through the code. But I could not find something noteworthy. Let's move on.

PR #812, opened 2026-08-05
Eliminate CICS findings

Lotes:
**Following items were handled**

- [ ] `EXEC CICS DUMP TRANSACTION` shall have a better error message since the `DUMPCODE(...)` option is required.
- [x] DELAY without options should work
- [ ] `EXEC CICS ALLOCATE;` throws errors on mainframe
  - `'( SESSION() | SYSID() | PARTNER() )' REQUIRED BUT NOT SPECIFIED FOR COMMAND. COMMAND NOT TRANSLATED.`
  - `'SYSID()' CONFLICTS WITH 'PARTNER()' AND HAS BEEN IGNORED.`
We should adapt the checkers for `ALLOCATE`
- [x] PSEUDOBIN and STATE for RECEIVE command (there is a mismatch between grammar and documentation… but there is also no syntax documented https://www.ibm.com/docs/de/cics-ts/5.5.0?topic=summary-receive-non-zoscommunications-server-options)
- [x] `EXEC CICS CONVERSE ALTERNATE DEFAULT;` results in Mainframe: `'ALTERNATE' CONFLICTS WITH 'DEFAULT' AND HAS BEEN IGNORED.` We should add this to the checker
- [x] `EXEC CICS CONVERSE CONVID(...) SESSION` results in mainframe: `'SESSION()' CONFLICTS WITH 'CONVID()' AND HAS BEEN IGNORED.` We should add a conflict checker here
- [x] `EXEC CICS CONVERSE LDC(...) FMH` results in mainframe: `'LDC()' CONFLICTS WITH 'FMH' AND HAS BEEN IGNORED.` We should add a conflict checker here
- [x] check all commands for ungiven duplicated option check. It should be enabled in most of the command (if not even every). Example: `ISSUE ABORT` with `DESTID` and `DESTIDLENG`

**The remaining items are handled in a separate PR.**


msujew (approved):
Thanks, looks good 👍

PR #813, opened 2026-08-10
Fix more CICS findings

Lotes:
- [x] `EXEC CICS DUMP TRANSACTION` shall have a better error message since the `DUMPCODE(...)` option is required.
- [x] `EXEC CICS ALLOCATE;` throws errors on mainframe
  - `'( SESSION() | SYSID() | PARTNER() )' REQUIRED BUT NOT SPECIFIED FOR COMMAND. COMMAND NOT TRANSLATED.`
  - `'SYSID()' CONFLICTS WITH 'PARTNER()' AND HAS BEEN IGNORED.`
We should adapt the checkers for `ALLOCATE`
- [x] `ISSUE RECEIVE` must result in a conflict if `INTO` and `SET` are given
- [x] `ISSUE RECEIVE` needs `LENGTH` only if `SET` is given
- [x] `ISSUE REPLACE` requires `KEYLENGTH or RNN` if not given, confusing case
- `RECEIVE MAP` should show that `SET` or `INTO` are required
  - was NOT a bug. There is already a test for that
- `ACQUIRE PROCESS`: if you only write `ACQUIRE`, no meaningful error like "PROCESS required" appears
   - THIS ONE cannot be solved with the usual checkers. Must be implemented differently, since "ACQUIRE PROCESS" decides to visit the corresponding checker, not "ACQUIRE" alone.
- `DEFINE TIMER(XXX)`
  - [x] requires only `AT`, but `AFTER` is also a valid continuation. Add `AFTER` to the missing required options
  - [x] When `MONTH` is a missing option `DAYOFYEAR`should also be listed
  - [x] Same for `YEAR`
- [x] `INQUIRE CONTAINER` must use `PROCESSTYPE` if `PROCESS` is given
- [x] `RUN` without options, result in mainframe saying: `RUN FUNCTION NOT COMPLETELY SPECIFIED. ACQACTIVITY OR ACQPROCESS OR ACTIVITY OR TRANSID MUST BE SPECIFIED. COMMAND NOT TRANSLATED.`. We have no meaningful error here

### DIFF
--- a/packages/preprocessor-cics/src/checks/base.ts
+++ b/packages/preprocessor-cics/src/checks/base.ts
@@ -34,6 +34,17 @@ export class CICSCheckUtilityParameters {
   public literalChecks: CICSLiteralCheckOption = CICSLiteralCheckOption.APOST;
 }
 
+export type AggregatableDiagnostic = Diagnostic & {
+  commonMessage: string;
+  enumerablePart: string;
+};
+
+export function isAggregatableDiagnostic(
+  diagnostic: Diagnostic,
+): diagnostic is AggregatableDiagnostic {
+  return "commonMessage" in diagnostic && "enumerablePart" in diagnostic;
+}
+
 export abstract class CICSOptionsCheckerBase {
   //private context: DialectProcessingContext;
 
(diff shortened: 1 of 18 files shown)


msujew (commented):
Thanks, looks good. Just a minor thing about wording, see below.

msujew (approved):

Issue #804, opened 2026-07-22
Adjustment on config. management


asatklichov:
Currently we are merging items on component level between .plugin folder two files, and VSCode user setting two objects. 
To be align with HLASM extension, we need to adjust logic as below  

- no merging in component level
- priorities (.plugin config beats vscode config [ workspace user config] ) 
- setting should be considered from VSCode user and workspace settings 
- search on file level simply, 
1. e.g. if in .plugin folder both files exist, but e.g. some processor name to found even still existing in VSCode user config, it should fail with not found
2. still possible to work combined (if one file not exist in .pluin folder and exists in VSCode user setting then get that info)

msujew:
Has been resolved by https://github.com/zowe/zowe-pli-language-support/pull/814.

PR #814, opened 2026-08-11
Introduce workspace settings and config fallback

Wagner-Laranjeiras:
Related to #804 

## Per-file config source selection + workspace-settings scope

Plugin config (`pli.pgm_conf` / `pli.proc_grps`) now resolves per file from a
single source by precedence — **project `.pliplugin/` > workspace settings >
user settings** — instead of merging sources. Adds a `workspace` settings
scope alongside `user` (mirroring VS Code), so `.vscode/settings.json` /
`.code-workspace` values are honored and outrank user settings.

### Behavior
- Each config file (`pgm_conf`, `proc_grps`) is selected independently and
  taken **whole** — never merged property-by-property.
- A `.pliplugin/` file that *exists* (even empty/invalid) is used as-is and
  blocks fallback; fallback only kicks in when it's missing.
- Fallback workspace (files outside any folder) no longer flags
  workspace-relative libs from user settings as unresolved, since it has no
  root to resolve them against; absolute libs are still validated.

### Tests
- Config precedence/fallback pinned end-to-end via new Fourslash tests under
  `fourslash/preprocessor/include/program-config-*`.
- Regression tests for the fallback-workspace relative-vs-absolute lib handling.

### DIFF
--- a/packages/language/src/config/schema.ts
+++ b/packages/language/src/config/schema.ts
@@ -106,13 +106,6 @@ export interface GroupRecord extends ProcessGroup {
 }
 
 /**
- * Where a {@link ProgramRecord} was loaded from. Project (`.pliplugin/`)
- * entries take precedence over user-scope `settings.json` entries when a file
- * matches program configs from both sources — see `getProgramConfig`.
- */
-export type ProgramOrigin = "project" | "settings";
-
-/**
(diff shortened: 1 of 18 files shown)


msujew (approved):
Thanks, looks good to me 👍

PR #816, opened 2026-08-12
Properly propose language ID change again

### LINKED ISSUE #788: Ask again about language ID when previously accepted
See:

https://github.com/zowe/zowe-pli-language-support/blob/f103a3bae0e7df17f9748fd26e791b8f61d8d3e8/packages/vscode-extension/src/extension/document-identification.ts#L19

Currently, we never ask again, even if the user previously accepted. We should only store the file there if the user rejected the question.

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/788

Previously, when reopening a file that was already accepted to be PL/I, the notification toast no longer appears to set it to PL/I again afterwards. This change correctly only adds the URI to the set of previously proposed files if the user declines to set it to PL/I.

### DIFF
--- a/packages/vscode-extension/src/extension/document-identification.ts
+++ b/packages/vscode-extension/src/extension/document-identification.ts
@@ -37,13 +37,13 @@ async function checkFileType(
   }
   if (proposedFiles.has(document.uri.toString())) {
     // We've already proposed to change the language for this document.
+    // It was previously declined, so don't propose again.
     return;
   }
   if (document.languageId !== "plaintext") {
     // Only attempt to identify PL/I documents that are currently marked as plaintext
     return;
   }
-  proposedFiles.add(document.uri.toString());
   if (
     // Assert that this can even be a PL/I document
     // (and isn't something else, like a .git file or a listing file)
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/extension/document-identification.ts
```diff
@@ -133,11 +136,17 @@ async function proposePliLanguage(document: vscode.TextDocument) {
     "No",
     "Never",
   );
-  if (selection === "Yes") {
-    vscode.languages.setTextDocumentLanguage(document, "pli");
-  } else if (selection === "Never") {
-    // Change the settings to never recommend PL/I for any file.
-    const settings = Settings.getInstance();
-    await settings.setAutoDetect(false);
+  switch (selection) {
```
ssmifi: _Question_: Since the file is not added unconditionally any longer, how is "dismissal" supposed to behave? It would now _nag_ indefinitely, right?

ssmifi (commented):
LGTM, just one question.

packages/vscode-extension/src/extension/document-identification.ts
msujew: It will nag indefinitely, as long as you don't click `No` or `Never`, yes. I think that's the best solution right now.

ssmifi: Ok, as long as it is a deliberate choice. :)

ssmifi (approved):

Issue #799, opened 2026-07-16
Wildcards for program names


asatklichov:
Wildcard implementation works fine with processor groups for 'lib' folder. 
But, for program names does not work as intended, e.g. Below config should allow any program opened under KO folder is acted as a entry. 
Once I try it, %include resolution does not give either error, or no go-to definition. same for PL/I files which has no .pli or pl1 extension 
Please double check second part of [758 ](https://github.com/zowe/zowe-pli-language-support/issues/758)

  ```
 {
      "program": "KO/**/*",
      "pgroup": "iba"
    },
```

<img width="393" height="176" alt="Image" src="https://github.com/user-attachments/assets/5ff5d5f4-da7b-4b7e-aa85-ab9dc44d1cb5" />

PR #817, opened 2026-08-13
Recognize wildcard and extensionless program entries

Wagner-Laranjeiras:
Solves #799

`pgm_conf.json` entries that use a wildcard or omit a file extension now
correctly bind their on-disk files as PL/I program entry points.

## What changed

- **Extensionless entries bind the real file.** An entry like `KO/PROG1` now
  matches `KO/PROG1.pli` / `.pl1` (and any extension the bound process group
  declares via `include-extensions`), so a program listed without an extension
  still resolves.
- **Wildcards cover extensionless members.** A recursive glob such as
  `KO/lib/**/*` picks up mainframe-style members that have no extension.
- **Client language identification.** The extension can now identify a
  configured file as PL/I even when VS Code opens it as `plaintext`. An
  **exact** program entry is trusted and the language is set silently; anything
  else — including broad **glob** matches — defers to the existing content
  heuristic and the one-time *"set language to PL/I?"* prompt (gated by
  `pli.autoDetect`), so a broad glob can never silently mislabel a non-PL/I file.
- **Unified identification request.** Following review feedback, the separate
  `pli/matchesProgramConfig` request was merged into `pli/existingFile`, which now
  returns `{ existing, programMatch }` in a single round-trip. The matcher
  classifies each candidate as `exact` / `glob` / `none`, and an exact match always
  wins over a glob (extensionless stems that resolve via an assumed extension still
  count as exact).

### DIFF
--- a/packages/language/src/workspace/plugin-configuration-provider.ts
+++ b/packages/language/src/workspace/plugin-configuration-provider.ts
@@ -18,7 +18,7 @@ import {
   offsetLengthToRange,
 } from "../language-server/types";
 import { mergeAbstractOptions } from "../config/compiler-options-merge";
-import { expandGroup } from "../config/lib-expander";
+import { DATASET_MEMBER_FILE_REGEX, expandGroup } from "../config/lib-expander";
 import { resolveLibUri } from "../config/path-resolver";
 import {
   ParseEntry,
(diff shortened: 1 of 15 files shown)


packages/language/src/workspace/plugin-configuration-provider.ts
```diff
@@ -1185,13 +1189,108 @@ export class PluginConfigurationProvider {
...
+        `Invalid program pattern "${pattern}" for program "${path}": ${e}`,
+      );
+      return "none";
+    }
+
+    const hasWildcard = decoded.includes("*");
+    const basename = decoded.substring(decoded.lastIndexOf("/") + 1);
+    const assumedExts = programBasenameIsConcrete(basename)
+      ? []
+      : this.assumedProgramExtensions(record);
```
msujew: Question: Is the `assumedExts` part of the requirements for this issue? It's not really clear coming from https://github.com/zowe/zowe-pli-language-support/issues/799.

msujew (commented):
Thanks!

msujew (approved):
Looks good to me, thanks 👍

PR #828, opened 2026-08-18
Fix last remaining parser errors

### LINKED ISSUE #820: Green code - tolaration
Please tolerate the code, when there is no description in spec

e.g BUFFERS(1)  in PDUMVS.pli .

And look errors in: FORM01.pli, IMSDBUT.pli, PDUMVS.pli

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/820

Fixes the issues found in `PDFUMVS.pli` mostly. The other files mentioned in the issue don't compile on the mainframe due to unexpected syntax issues:

1. `BUFFERS` is an undocumented option for the `ENVIRONMENT` attribute.
2. `ENTRY` statement labels should be visible outside of their owning procedure.
3. Some array declarations using asterisks are fine like `DCL SOME(*) CTL CHAR`.


packages/language/src/linking/symbol-table.ts
```diff
@@ -602,6 +603,24 @@ function handleNode(
...
+        ) {
+          const procedureStatement = getContainer(
+            statement,
+            SyntaxKind.ProcedureStatement,
+          );
+          if (procedureStatement?.container) {
+            // Labels of ENTRY statements declare entry points of a new procedure.
+            // Like the parent procedure itself, they are visible in the
+            // scope that contains the procedure, not just inside of it.
+            parentScope =
```
ssmifi: _Question_: Doesn't this move outer forward declarations into the same scope? For example, the following code now clashes but did not so before. Can you please verify if this is intented or a new false positive?
```
 DCL SOME_ENTRY ENTRY(CHAR(11)) RETURNS(CHAR(11) VAR);
 MAIN: PROC OPTIONS(MAIN);
   PUT(SOME_ENTRY('X'));
 END MAIN;
 OTHER: PROC;
   SOME_ENTRY: ENTRY(C) RETURNS(CHAR(11) VAR);
 END OTHER;
```

packages/language/src/validation/type-check-validator.ts
```diff
@@ -140,12 +140,6 @@ function validateBound(
 ) {
   if (bound.expression) {
     if (bound.expression.kind === ast.SyntaxKind.WildcardItem) {
-      if (
-        bound.token &&
-        !ast.getContainer(bound.node, ast.SyntaxKind.ProcedureStatement)
-      ) {
-        acceptor(diagnosticFromCode(PLICodes.Severe.IBM1629I, bound.token));
-      }
       return "unbounded";
```
ssmifi: _Question_: Just making sure I understand, `DCL ANYTHING(*) FIXED BIN;` is valid then, right? And also, this removes `IBM1629I` entirely. But shouldn't there a case (probably with `STATIC`) where this applies?

packages/vscode-extension/syntaxes/pli.manual.json
```diff
@@ -24,7 +24,7 @@
     {
       "name": "string.double.single.pli",
```
ssmifi: _Minor artifact_: Should probably be `string.quoted.double.pli`.

ssmifi (commented):
Thanks. Resolves the issue. I have two related questions.

packages/language/src/linking/symbol-table.ts
msujew: On the compiler I get:
```
IBM1916I S      15.0    The PROCEDURE/ENTRY CH15 has already been defined.
```

Which is exactly what we're producing ([playground](https://zowe.github.io/zowe-pli-language-support/pr/pr-828/?content=AQYQEgjArAXMAKAlA8iA3AKGN%2BBVAKgBQBE%2BAogMr7ECUmwZAcgCL0CCAamYmwOJlwkqMs1yIywZPHwBJZIwrBCAWTYzGdLNmYgAMsFyMAQskPNCAKhrAAYjIAaI4EfWh8u%2BjgKFDJs4QgaTWxgHX1wYH5GbhkQJXAAZghgAHUwJkI7R2ZndUIkoIAaLRDSsrLwaFT0xkyHJxda6CDgphzObj4yekSIQRQQQjZrcXwxBUJwNkQAwOAOabpy5ZXV0pKQsOdQMGnZ6wXEYEKlNkKU6yyGvIK0NfvsDa98YCpEdV5CI2sRGSJhpQ2fZLB6g1ZPSLIfDIHb2O5ghGItZYZjIHqQWAMRj4RAATUmIzIY0QEymMwgc0OIKRYIhWziVxyjQCUGpNPZ63KeBebw%2BXx%2BzD%2BBMBwPhwCwbTFHJWWHA9jgKQAvFx3jZ8UZCgByYCatnS-Xi8pGRUUXBGN5fc56g0GiGjcb8qU252PcptHZJNBAA)).

packages/language/src/validation/type-check-validator.ts
msujew: We could narrow it down, yes. The compiler does show the error, but not in all cases where `(*)` is used as bounds. However, the check is quite a bit more complicated. I would defer it.

ssmifi (approved):
Thanks

PR #831, opened 2026-08-18
Show error on PROCEDURE if end is missing

### LINKED ISSUE #823: Highlight unterminated procedure block
If a procedure is missing its corresponding `END` statement, it's sometimes difficult to find out, what exactly is the problem. Currently, users just see a parser error message, about a missing `END` statement after the last token.

If we find that a procedure has no END statement, we should show a diagnostic on the procedure label itself, to make it clear which procedure is actually at fault here.

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/823

Will show a new error if the procedure is unterminated.

### DIFF
--- a/packages/language/src/validation/language-server/procedure-end.ts
+++ b/packages/language/src/validation/language-server/procedure-end.ts
@@ -1,0 +1,34 @@
+/**
+ * This program and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+import { diagnosticFromCode } from "../../language-server/types";
+import { ProcedureStatement } from "../../syntax-tree/ast";
+import { CompilationUnit } from "../../workspace/compilation-unit";
+import { LspCodes } from "../lsp-codes";
+import { ValidationAcceptor } from "../validator";
+
+export function checkProcedureEnd(
+  node: ProcedureStatement,
+  acceptor: ValidationAcceptor,
+  compilationUnit: CompilationUnit,
+): void {
+  if (
+    !node.end?.endToken &&
+    // TODO: Find a way to also support the multi-close option here
+    !compilationUnit.compilerOptions.rules?.multiClose
+  ) {
+    // If the procedure has no end, and multi-close is not enabled,
+    // we provide an additional diagnostic (in addition to the parser error)
+    // to indicate that the procedure is missing an END statement.
+    // This should help users identify the problem more easily.
+    // Note: This shows the diagnostic on the PROC token, not a procedure label
+    acceptor(diagnosticFromCode(LspCodes.MissingEnd, node.procToken));
+  }
+}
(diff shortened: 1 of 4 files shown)


packages/language/src/validation/language-server/procedure-end.ts
```diff
@@ -0,0 +1,32 @@
...
+): void {
+  if (
+    !node.end?.endToken &&
+    !compilationUnit.compilerOptions.rules?.multiClose
+  ) {
+    // If the procedure has no end, and multi-close is not enabled,
+    // we provide an additional diagnostic (in addition to the parser error)
+    // to indicate that the procedure is missing an END statement.
+    // This should help users identify the problem more easily.
+    acceptor(diagnosticFromCode(LspCodes.MissingEnd, node.procToken));
```
ssmifi: _Remark:_ Just a remark that the diagnostic is actually shown at the keyword and not at the label as stated in the issue, which is still fine, I guess.

packages/language/src/validation/language-server/procedure-end.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ValidationAcceptor } from "../validator";
+
+export function checkProcedureEnd(
+  node: ProcedureStatement,
+  acceptor: ValidationAcceptor,
+  compilationUnit: CompilationUnit,
+): void {
+  if (
+    !node.end?.endToken &&
+    !compilationUnit.compilerOptions.rules?.multiClose
```
ssmifi: _Minor_: Minor because multiclose is probably relatively exotic. This, however, suppresses the diagnostic if the multiclose actually is missing an end statement. If this is out of scope now, please leave a comment.

packages/language/test/fourslash/validate/missing-end-statement.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright Contributors to the Zowe Project.
+ *
+ */
+
+/// <reference path="../framework.ts" />
+
+//// MAIN: <|PROC|> OPTIONS(MAIN);
+////  DCL TEST_VALUE CHAR(5) INIT("HELLO");
+
+verify.expectDiagnosticsAt("PROC", code.LspCodes.MissingEnd);
```
ssmifi: _Future suggestion_: Most likely out of scope, but a missing semicolon
```
 MAIN: PROCEDURE;
    DECLARE DEPT FIXED;
 END
```
will emit the same diagnostic. Would be nice to give an explicit message that the semicolon is missing, but I guess this would require a parser change, right? So, just a suggestion for the future.

ssmifi (approved):
Generally good to go. I just left a few remarks for future improvements.

packages/language/src/validation/language-server/procedure-end.ts
msujew: It's quite a bit more difficult to check for this. I would probably want to introduce an additional flag in the AST for this. I'll add a todo.

packages/language/test/fourslash/validate/missing-end-statement.ts
msujew: The issue with the missing `;`, is that we can't parse it correctly, because `END` by itself is not clearly a full end-statement. It could also be a start to something like `END = 24;`.

PR #832, opened 2026-08-19
Display diagnostics on large macro arrays

### LINKED ISSUE #791: [Parser] Resolve hidden message and display
After architectural change regarding preprocessor. Please resolve and provide error message for below scenarious 

```
%DCL X(999999999) CHAR;
 %X1 = "X(1)";
 %ACT X1;
 X1  dcl xx fixed;   

  HELLO:
    PROCEDURE OPTIONS(MAIN);
    /*A PROGRAM TO OUTPUT HELLO WORLD*/  ;
    PUT LIST ('HELLO, WELCOME TO PLI/1 WORLD ');
  END HELLO;
```

msujew:
Closes https://github.com/zowe/zowe-pli-language-support/issues/791

Right now, if we detect that a variable is too large, we just silently ignore the actual value and set it too an empty array. This change adds a diagnostic warning for this case.

### DIFF
--- a/packages/language/src/validation/lsp-codes.ts
+++ b/packages/language/src/validation/lsp-codes.ts
@@ -69,6 +69,13 @@ export const LspCodes = {
     message: () => `Missing END statement for procedure.`,
   },
 
+  TooLarge: {
+    code: "LSPTL001",
+    severity: Severity.W,
+    message: (max: number) =>
+      `Dimensions exceed the maximum allowed size of ${max}.`,
+  },
+
   BuiltinAttributes: {
     IsForbiddenUsage: {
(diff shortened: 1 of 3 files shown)


ssmifi (approved):
LGTM

packages/language/src/validation/lsp-codes.ts
```diff
@@ -69,6 +69,13 @@ export const LspCodes = {
     message: () => `Missing END statement for procedure.`,
   },
 
+  TooLarge: {
```
ssmifi: _Minor_: Could be names `DimensionsTooLarge`, but is also fine as it is. :)

PR #834, opened 2026-08-20
Remove plugin loader error

msujew:
Removes an error message that we generate in case the pliplugin files are malformed. Quietly rejects them now. We do not lose any user facing diagnostics, because the json schema already validates that the corresponding settings/files are correctly formed JSON.


Wagner-Laranjeiras (approved):
LGTM 👍

